Issue #1179, opened 2024-04-02
🦋 Allow 100% discount on an order on POS seat


Tirodem:
Currently blocked :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/fdc4e61c-8648-4a51-b328-c5f9a23c30d9)

Allow it in both % and currency amount as long as the "justification" field is filled.

PR #1182, opened 2024-04-02
Allow 100% discount on an order on POS seat #1179

ithiame:
🦋 Allow 100% discount on an order on POS seat #1179


coyotte508:
In that case we shouldn't add an order payment, the order should directly be paid, and `payments` should be an empty array. No need for payment method handling.

coyotte508:
We still need to redo the logic that happens when an order is paid - add subscriptions, update challenges, tickets, mark order as paid, update totals etc

It can be factored out from `onOrderPayment` into a separate function

PR #1184, opened 2024-04-02
Notify bebop super admin

ithiame:


coyotte508 (approved):

src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts
```diff
@@ -119,7 +122,19 @@ export const actions = {
...
+		<p>Follow <a href="${ORIGIN}/order/">this link</a> to see the change.</p>
+		<p>The change was made by ${locals.user?.alias}.</p>`;
+		if (runtimeConfig.shopInformation?.contact.email) {
+			await collections.emailNotifications.insertOne({
+				_id: new ObjectId(),
+				createdAt: new Date(),
+				updatedAt: new Date(),
+				subject: 'Update Payment Information',
+				htmlContent: templateKey,
+				dest: runtimeConfig.shopInformation?.contact.email || SMTP_USER
```
Comment (Tirodem) [inline src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:135]: This should not be this :
runtimeConfig.shopInformation?.contact.email
But the super admin account contact (right now let's consider the email as mandatory, and in the future we'll deal with nostr too).
But² I've remembered that superadmin account has no contact information at creation, I did it manually on dev env.
I need to deal with that on a new ticket.

Comment (coyotte508) [reply src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:135]:
```suggestion
				dest: runtimeConfig.sellerIdentity?.contact.email || SMTP_USER
```

To be consistent

(and it's what we use everywhere at the moment @Tirodem, for copying orders, the "reply to" field in emails, discounts, etc)

Issue #1193, opened 2024-04-03
☢️ Free cart cannot be ordered (anymore ?)


Tirodem:
Related to https://github.com/B2Bitcoin/beBOP/issues/1179 / https://github.com/B2Bitcoin/beBOP/pull/1182

Here's the current behavior on a cart at 0 with free product :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/aac8e0e5-ff95-47d8-931d-70481b36263e)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/9573f117-f9d6-4efd-b64f-c4dfd36db2cf)

I'm nearly sure it worked in the past, but that's not the point, so as #1120 is progressing, here's how I see the thing :
- create a new payment method "Nothing to pay (free cart)"
- when cart amount to pay = 0, only display this payment mean
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/8bb7a592-482b-42a2-859f-ad8aa3c37c0b)
- when processing the checkout, create the payment directly at "paid" status

coyotte508:
related to 100% discount cc @ithiame

Tirodem:
Sure, but to be clearer on the "why", I created the issue because selling
cart with free product is for now more important that 100% discount on POS,
that is a nice to have (but allowed to find the issue).

PR #1195, opened 2024-04-08
 Free cart cannot be ordered (anymore ?) #1193

ithiame:


Tirodem:
Functional feedbacks (seen on workshop thread) :
❌ no sent notification (because of no payment created ?)
☢️ "nothing to pay (free cart)" is available if I have a free product + a paying product (so, a non-free cart)
🔬 as we're creating no payment the order has no receipt, that can be a problem in... "some cases"

Note that this ticket allowed me to find this :
https://github.com/B2Bitcoin/beBOP/issues/1196

coyotte508 (commented):
Can you test if it still works if we change the payment methods through the UI in `/admin/config`?

It should still allow to do free payments

ithiame:
> Can you test if it still works if we change the payment methods through the UI in `/admin/config`?
> 
> It should still allow to do free payments
It still works

coyotte508:
you can try to find why the tests are failing, otherwise I'll look into it later

coyotte508 (approved):
Seems good for me

If functional ok, then :+1:

Tirodem:
That point is fixed :
☢️ "nothing to pay (free cart)" is available if I have a free product + a paying product (so, a non-free cart)

For the last points :

🔬 as we're creating no payment the order has no receipt, that can be a problem in... "some cases"
That can wait for a bit (as it's OK in the native, law compliant, invoice behavior of beBOP).

❌ no sent notification (because of no payment created ?)
That one is more annoying to me (that's why I expected a "free payment" payment mean, I supposed that the payment was the trigger to the notifications).

Still, we have fixed the free price UX issue, and if there's an issue (loss of order URL), the customer can buy it for free again, so I'll go merge for now.
And create tickets for the future for these two points.

coyotte508:
It's possible to add a payment "free" in the payments and mark it as paid cc @ithiame

like a regular payment

PR #1200, opened 2024-04-08
Refresh number displayed /checkout amount with discount, free deliver…

ithiame:
…y free and VAT exemption on POS seat #1120

### DIFF
--- a/src/routes/(app)/checkout/+page.svelte
+++ b/src/routes/(app)/checkout/+page.svelte
@@ -31,7 +31,7 @@
 	let isFreeVat = false;
 	let offerDeliveryFees = false;
 	let addDiscount = false;
-	let discountAmount: number;
+	let discountAmount: number = 0;
 	let discountType: DiscountType;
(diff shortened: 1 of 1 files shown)


coyotte508 (approved):

Tirodem:
Here's what I have on dev-bootik2 after deploying this branch, on both user & POS seat :
image
It's products without deposit.
@ithiame can you check if you have that too ? 🤔

Tirodem:
@ithiame :
image
✅  Discount currency / % is OK
image
image
✅  VAT exemption is OK
image
❌ Free delivery is KO
image
❌ Unchecking Discount after filling an amount keep the discounted price
image
image
image

Tirodem:
We're good with the amount for free delivery but it would be nice to hide the delivery fees line (like we do with the VAT line) or put the amount to 0
image
can be in a later ticket by the way so ok ✅ 

It's also with no further warning for the check / fill / uncheck discount ✅

Issue #1202, opened 2024-04-09
💸 Add a "Free" payment mean to unlock receipt & notifications on free orders


Tirodem:
Following https://github.com/B2Bitcoin/beBOP/pull/1195#issuecomment-2043195547 ( #1193 ), improve the free orders with a "Free" payment mean.
That'll allow these kinds of orders to have notifications & receipts.

PR #1203, opened 2024-04-09
Add a 'Free' payment mean to unlock receipt & notifications on free o…

ithiame:
…rders #1202

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -1047,10 +1047,6 @@ export async function addOrderPayment(
 	 */
 	opts?: { expiresAt?: Date | null; session?: ClientSession }
 ) {
-	if (paymentMethod !== 'free' && order.status !== 'pending') {
-		throw error(400, 'Order is not pending');
-	}
-
 	if (paymentMethod !== 'free' && isOrderFullyPaid(order, { includePendingOrders: true })) {
 		throw error(400, 'Order already fully paid with pending payments');
 	}
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -544,6 +544,10 @@ export async function createOrder(
 		throw error(400, 'Missing billing address for deliveryless order');
 	}
 
+	if (paymentMethod === 'free' && totalSatoshis !== 0) {
+		throw error(400, "You can't use free payment method on this order");
+	}
+
 	await withTransaction(async (session) => {
 		const order: Order = {
 			_id: orderId,
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -856,10 +856,11 @@ export async function createOrder(
 		};
 		await collections.orders.insertOne(order, { session });
 
+		let orderPayment;
 		if (paymentMethod) {
 			const expiresAt = paymentMethodExpiration(paymentMethod);
 
-			await addOrderPayment(
+			orderPayment = await addOrderPayment(
 				order,
 				paymentMethod,
 				{ currency: 'SAT', amount: partialSatoshis },
@@ -876,6 +877,9 @@ export async function createOrder(
 			if (product.stock) {
 				await refreshAvailableStockInDb(product._id, session);
 			}
+		}
+		if (paymentMethod === 'free' && orderPayment) {
+			await onOrderPayment(order, orderPayment, orderPayment.price, { providedSession: session });
 		}
 	});
 
@@ -1134,9 +1138,8 @@ export async function addOrderPayment(
 		},
 		{ session: opts?.session }
 	);
-	if (paymentMethod === 'free') {
-		await onOrderPayment(order, payment, payment.price, { providedSession: opts?.session });
-	}
+
+	return payment;
 }
 
 export async function updateAfterOrderPaid(order: Order, session: ClientSession) {
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -82,7 +82,7 @@ export async function onOrderPayment(
 ): Promise<Order> {
 	const invoiceNumber = ((await lastInvoiceNumber()) ?? 0) + 1;
 
-	if (!order.payments.includes(payment) && payment.method !== 'free') {
+	if (!order.payments.includes(payment)) {
 		throw new Error('Sync broken between order and payment');
 	}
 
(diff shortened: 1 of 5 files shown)


coyotte508 (commented):
:warning: 

I didn't pick this up in the previous PR, but there's no backend check to allow the use of the `free` payment method

A user can make a request to the backend with the 'free' payment method and get free stuff even when it's not free

Please fix

src/lib/server/orders.ts
```diff
@@ -1110,6 +1115,12 @@ export async function addOrderPayment(
 				}
 			}
 		},
+		...(paymentMethod === 'free' && {
+			invoice: {
+				number: invoiceNumber,
+				createdAt: new Date()
+			}
+		}),
```
coyotte508, line 1123: Hmm what you can do instead (to avoid too much custom code) is:

- call `addOrderPayment` like normal
- if payment method is free, call `onOrderPayment` immediately after

That way you can keep the order pending initially (and remove the check), you can remove this custom code, you can remove the code that marks the order as paid initially, you can remove the code that marks the payment as already paid, and you can even remove the code that calls `updateAfterOrderPaid` (since it will be called inside `onOrderPayment` already)

ithiame: Tried this

``` 
                if (paymentMethod) {
			    const expiresAt = paymentMethodExpiration(paymentMethod);
			    await addOrderPayment(
				    order,
				    paymentMethod,
				    { currency: 'SAT', amount: partialSatoshis },
				    { session, expiresAt }
			    );
		    }
		    
		    if (paymentMethod === 'free') {
			    await onOrderPayment(order, order.payments[0], {
				    amount: 0,
				    currency: runtimeConfig.mainCurrency
			    });
		    }
```
 but got this error.

```
handleError Error: Sync broken between order and payment
    at onOrderPayment (eval at instantiateModule (file:///D:/PVH%20labs/B2BitcoinBootik/node_modules/.pnpm/vite@4.5.0_@types+node@20.9.0/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:56049:28), <anonymous>:78:11)
```

coyotte508: ah yes you would need to add an optional `session` parameter to `onOrderPayment`

If provided, it would use `session`, otherwise, it would use the session in `withTransaction`

src/lib/server/orders.ts
```diff
@@ -91,6 +92,7 @@ export async function onOrderPayment(
 	payment.paidAt = paidAt;
 
 	return await withTransaction(async (session) => {
+		const usedSession = params?.providedSession || session;
```
coyotte508, line 95:
```ts
const fn = async (session: ClientSession) => {
  ...
}

return params?.providedSession ? await fn(params.providedSession) : await withTransaction(fn);
```

src/lib/server/orders.ts
```diff
@@ -1048,11 +1047,11 @@ export async function addOrderPayment(
 	 */
 	opts?: { expiresAt?: Date | null; session?: ClientSession }
 ) {
-	if (order.status !== 'pending') {
```
coyotte508, line 1051: I think you can remove this line and set the order to `pending` initially like with non-free apyment method?

And remove the call to `updateAfterOrderPaid` as well in `createOrder` in the case of free payment method

src/lib/server/orders.ts
```diff
@@ -1048,11 +1047,7 @@ export async function addOrderPayment(
...
-		throw error(400, 'Order is not pending');
-	}
```
coyotte508, line 1053: Sorry i meant "reverse this change"

src/lib/server/orders.ts
coyotte508: Marking unresolved, this is not done:

- set the order to pending initially like with non-free apyment method?
- remove the call to updateAfterOrderPaid as well in createOrder in the case of free payment method

ithiame: > remove the call to updateAfterOrderPaid as well in createOrder in the case of free payment method

already removed ?

coyotte508 (approved):
with last comment taken care of

src/lib/server/orders.ts
```diff
@@ -1135,6 +1134,9 @@ export async function addOrderPayment(
 		},
 		{ session: opts?.session }
 	);
+	if (paymentMethod === 'free') {
+		await onOrderPayment(order, payment, payment.price, { providedSession: opts?.session });
+	}
```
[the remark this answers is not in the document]
ithiame: payment && payment.price can't be accessed in` createOrder` if we move it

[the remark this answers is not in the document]
coyotte508: `return payment` ?

src/lib/server/orders.ts
```diff
@@ -854,13 +855,12 @@ export async function createOrder(
 			...(params.engagements && { engagements: params.engagements })
 		};
 		await collections.orders.insertOne(order, { session });
-		if (order.status === 'paid' && paymentMethod && paymentMethod === 'free') {
-			await updateAfterOrderPaid(order, session);
-		}
-		if (paymentMethod && paymentMethod !== 'free') {
+
+		let orderPayment;
```
coyotte508, line 859: Please type it correctly. Currently it's type `any`.

```suggestion
		let orderPayment: ... | undefined = undefined;
```

src/lib/server/orders.ts
```diff
@@ -878,6 +878,9 @@ export async function createOrder(
 				await refreshAvailableStockInDb(product._id, session);
 			}
 		}
+		if (paymentMethod === 'free' && orderPayment) {
```
coyotte508, line 881:
```suggestion
		if (orderPayment?.method === "free") {
```

src/lib/server/orders.ts
```diff
@@ -77,20 +77,20 @@ export async function onOrderPayment(
 	params?: {
 		bankTransferNumber?: string;
 		detail?: string;
+		providedSession?: ClientSession;
 	}
 ): Promise<Order> {
 	const invoiceNumber = ((await lastInvoiceNumber()) ?? 0) + 1;
 
-	if (!order.payments.includes(payment)) {
+	if (!order.payments.includes(payment) && payment.method !== 'free') {
```
coyotte508, line 85: We can revert this no?

coyotte508 (approved):

Issue #1209, opened 2024-04-14
🦋 Add user contact address on /order/{id} for employee account with read access on /admin/order


Tirodem:
Currently, buyer's email or npub information isn't given outside database.
For logistical and support needs (and to contact buyer), this information is usefull.
So : display user contact information (orders.user.email or oders.user.npub & any future contact information) on /order/{id} for employees having at least read access on /admin/order (so also POS seats)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/02b92cfb-5d63-4226-ad0a-084553461368)

PR #1214, opened 2024-04-15
Add user contact address on /order/{id} for employee account with rea…

ithiame:
…d access on /admin/order #1209


src/routes/(app)/order/[id]/fetchOrderForUser.ts
```diff
@@ -135,6 +136,7 @@ export async function fetchOrderForUser(orderId: string) {
 				isEmployee: note.role !== CUSTOMER_ROLE_ID,
 				alias: note.userAlias
 			})) || [],
-		receiptNote: order.receiptNote
+		receiptNote: order.receiptNote,
+		user: pojo(order.user)
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/fetchOrderForUser.ts:140]: Just send the user.email & npub if that's all that's needed, we don't want to send `sessionId`

coyotte508 (approved):

Issue #1210, opened 2024-04-14
🦋 Add optional "Phone" contact information on delivery form


Tirodem:
Is useful for logistical needs (sometimes required by carriers).
Desired implementation :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/43a69058-55d4-4487-83ee-d7d86f7fd463)
(the green frame is a highlight, it's not meant to be displayed)
Save it on orders.shippingAddress.phoneNumber.
Make 3 translations keys :
- one for phone information need explanation : "Cette information est optionnelle mais parfois requise par les transporteurs. Nous ne vous démarcherons pas sur ce numéro."
- one for the international number disclaimer "Si votre numéro n'est pas du même pays que le pays de livraison, merci de renseigner l'indicatif international."
- one for the international number disclaimer URL target :
  - fr : https://fr.wikipedia.org/wiki/Liste_des_indicatifs_t%C3%A9l%C3%A9phoniques_internationaux_par_indicatif
  - en : https://en.wikipedia.org/wiki/List_of_country_calling_codes
Add a hyperlink on 2e sentence with target="_blank".

Note that we don't use the phone number as a contact information yet because we need to integrate Traccar API Gateway link (related to https://github.com/B2Bitcoin/beBOP/issues/294 ).
That's why we keep the phone number as a delivery information on the said form for now.

PR #1215, opened 2024-04-15
Add optional 'Phone' contact information on delivery form #1210

ithiame:
🦋 Add optional "Phone" contact information on delivery form #1210

### DIFF
--- a/src/lib/translations/en.json
+++ b/src/lib/translations/en.json
@@ -6,6 +6,7 @@
 		"country": "Country",
 		"firstName": "First name",
 		"lastName": "Last name",
+		"phone": "Phone",
 		"state": "State",
 		"vatNumber": "VAT number",
 		"zipCode": "Zip code"
--- a/src/lib/translations/fr.json
+++ b/src/lib/translations/fr.json
@@ -6,6 +6,7 @@
 		"country": "Pays",
 		"firstName": "Prénom",
 		"lastName": "Nom",
+		"phone": "Téléphone",
 		"state": "Canton / Région",
 		"vatNumber": "Numéro TVA",
 		"zipCode": "Code postal"
--- a/src/routes/(app)/checkout/+page.server.ts
+++ b/src/routes/(app)/checkout/+page.server.ts
@@ -171,7 +171,11 @@ export const actions = {
 										city: z.string().min(1),
 										state: z.string().optional(),
 										zip: z.string().min(1),
-										country: z.enum([...COUNTRY_ALPHA2S] as [CountryAlpha2, ...CountryAlpha2[]])
+										country: z.enum([...COUNTRY_ALPHA2S] as [CountryAlpha2, ...CountryAlpha2[]]),
+										phone: z
+											.string()
+											.regex(/^\d+(\.\d+)?$/)
+											.default('0')
 								  }
 						)
 					})
--- a/src/routes/(app)/checkout/+page.svelte
+++ b/src/routes/(app)/checkout/+page.svelte
@@ -245,6 +245,18 @@
 							autocomplete="postal-code"
 						/>
 					</label>
+					<label class="form-label col-span-6">
+						{t('address.phone')}
+						<input type="number" name="shipping.phone" class="form-input" />
+					</label>
+					<p class="col-span-6">
+						This information is optional but sometimes required by carriers. We will not approach
+						you on this number.
+					</p>
+					<p class="col-span-6">
+						If your number is not from the same country as the country of delivery, please enter the
+						international code.
+					</p>
 
 					<label class="col-span-6 checkbox-label">
 						<input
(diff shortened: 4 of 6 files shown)


src/routes/(app)/checkout/+page.svelte
```diff
@@ -245,6 +245,18 @@
 							autocomplete="postal-code"
 						/>
 					</label>
+					<label class="form-label col-span-6">
+						{t('address.phone')}
+						<input type="number" name="shipping.phone" class="form-input" />
```
coyotte508, line 250:
```suggestion
						<input type="tel" name="shipping.phone" class="form-input" />
```

src/routes/(app)/checkout/+page.server.ts
```diff
@@ -171,7 +171,11 @@ export const actions = {
 										city: z.string().min(1),
 										state: z.string().optional(),
 										zip: z.string().min(1),
-										country: z.enum([...COUNTRY_ALPHA2S] as [CountryAlpha2, ...CountryAlpha2[]])
+										country: z.enum([...COUNTRY_ALPHA2S] as [CountryAlpha2, ...CountryAlpha2[]]),
+										phone: z
+											.string()
+											.regex(/^\d+(\.\d+)?$/)
+											.default('0')
```
coyotte508, line 178: no, why would be store `0` in a phone number field? Let's not add a regex either. `maxLength` if you want.

(a regex could be `/^([+]?[\s0-9]+)?(\d{3}|[(]?[0-9]+[)])?([-]?[\s]?[0-9])+$/` but not needed IMO)

Tirodem: Same thought, especially as we'll allow 0033123456789 or +33123456789 or 0123456789 or either "call 0123456789 or 9876543210 depending of the time of the day" or anything else on input for now, that's why I didn't say a thing about phone number format & input restriction in the ticket.

src/lib/translations/en.json
```diff
@@ -6,6 +6,9 @@
 		"country": "Country",
 		"firstName": "First name",
 		"lastName": "Last name",
+		"phone": "Phone",
+		"phoneCode": "If your number is not from the same country as the country of delivery, please enter the international code.",
```
coyotte508, line 10:
```suggestion
		"phoneCode": "If your phone number is not from the same country as the country of delivery, please enter the international code.",
```

src/lib/translations/en.json
```diff
@@ -6,6 +6,9 @@
...
+		"phoneDisclaimer": "This information is optional but sometimes required by carriers. We will not approach you on this number.",
```
coyotte508, line 11:
```suggestion
		"phoneDisclaimer": "The phone number is sometimes required by delivery carriers. We will not use it for marketing purposes.",
```

src/lib/translations/fr.json
```diff
@@ -6,6 +6,9 @@
 		"country": "Pays",
 		"firstName": "Prénom",
 		"lastName": "Nom",
+		"phone": "Téléphone",
+		"phoneCode": "Si votre numéro n'est pas du même pays que le pays de livraison, merci de renseigner l'indicatif international.",
```
coyotte508, line 10:
```suggestion
		"phoneCode": "Si votre numéro de téléphone n'appartient pas au pays de livraison, merci de renseigner l'indicatif international.",
```

src/lib/translations/fr.json
```diff
@@ -6,6 +6,9 @@
...
+		"phoneDisclaimer": "Cette information est optionnelle mais parfois requise par les transporteurs. Nous ne vous démarcherons pas sur ce numéro.",
```
coyotte508, line 11:
```suggestion
		"phoneDisclaimer": "Le numéro de téléphone est parfois demandé par les livreurs. Nous ne vous démarcherons pas sur ce numéro.",
```

src/routes/(app)/checkout/+page.server.ts
```diff
@@ -414,8 +416,9 @@ export const actions = {
 					}
 				},
 				cart,
-				shippingAddress: shippingInfo?.shipping,
-				billingAddress: billingInfo?.billing || shippingInfo?.shipping,
+				shippingAddress: shippingInfo?.shipping as OrderAddress,
+				billingAddress:
+					(billingInfo?.billing as OrderAddress) || (shippingInfo?.shipping as OrderAddress),
```
coyotte508, line 421: Why this change?

ithiame: It's show an error in the lint and in the code

coyotte508: It's because the zod types are messed up for `shippingInfo`.

For example you can add the `phone` optional field for POS users too, it should fix it. Another solution is to refactor: `firstName: isPos ? z.string().default('') : z.string().min(1), ...`.

In your editor you can hover over variables to see the types. You can see that the type for `shippingInfo` is wrong.

Please fix problems at the source instead of casting to hide the issues. Here the editor warns you that you didn't handle the POS case. (eg POS users's phone numbers are ignored)

coyotte508 (approved):

Issue #1222, opened 2024-04-19
🦋 Change preorder product date information on /admin/product/{id}


Tirodem:
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/532e77eb-cf0e-4486-aac3-3be7489f16ec)
Once a product is saved with a future availability date, emptying it doesn't not enough to make it available, you need to fill today or yesterday date.
But the text is kinda confusing, we need to change it (OR save today / yesterday's date - to test - if date is empty on form when updating product).

PR #1223, opened 2024-04-22
Change preorder product date information on /admin/product/{id} #1222

ithiame:
🦋 Change preorder product date information on /admin/product/{id} #1222

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/+page.server.ts
@@ -85,10 +85,6 @@ export const actions: Actions = {
 			parsed.preorder = false;
 		}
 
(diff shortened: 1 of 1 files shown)


coyotte508 (approved):
Make sure to test this thoroughly in order to ensure it doesn't remove the available date of preorders in some scenario

Tirodem:
> Make sure to test this thoroughly in order to ensure it doesn't remove the available date of preorders in some scenario

That point is OK but the initial point of the issue is KO.
When I replace my preorder date
image
by nothing :
image
and I save, here's what I have :
image
Initial goal of the issue is to fix that, I don't understand the diff of the PR as the empty value is not saved.

ithiame:
It should save an empty date and make the product available. Are you sure  there's no change ?

Tirodem:
My bad, deploy error.
Is OK after recheck ✅

Issue #890, opened 2023-12-20
🦋 Manage a 404 behavior (& other errors)


Tirodem:
Add a behavior selector in /admin/config (similar to what was done in lightning administration for invoice info selector)

Choices :
- display error (current behavior)
- display CMS page /error
- redirect to page /[to-define] + custom error message in a page headband

For the last :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/c0c3616d-a213-4981-8fc1-c2e7dd24f4fd)

You can take inspiration for redirection from error  from this commit : https://github.com/B2Bitcoin/beBOP/commit/1becd1091790056e1baae087f28a50441b6888d3

In the future it would be good to be able to define a global behavior on errors, but also to have a specific behavior on a specific error code, so keep that in mind while implementing this ticket to make the later full implementation easier.

coyotte508:
I think we should never redirect to another page in case of error.

Instead find a way to display the contents of the CMS error page on the current page if we want.

With a way to display the error message & error code if we want with a template.

Current way (redirecting to /error) is because it was tricky to do otherwise, but if we want to go further we should do it righit

PR #1225, opened 2024-04-22
Manage a 404 behavior (& other errors) #890

ithiame:


Tirodem:
Testing, pre-review :

- We need a default selected behavior, as the choice is blank on 1st time access
image
As the default behavior seems to be "Error 404 CMS Page not found" while selecting nothing, check this one if setting value does not exist.

- While choosing "Display error" & updating on /admin/config, I have a crash :
image

- While choosing "CMS page / error" without filling the custom error message in page headband :
image
This has to be optional (and **IF** not, error management should be pre-done in UI, raw error while submitting form should not happen)

- Selected behavior doesn't apply, and we still on a display error behavior

- Custom error message in page headband should only be displayed on "Display CMS page error" (here I have it on "redirect to page" and it disappears when clicking on "display cms page error")
image

Tirodem:
✅ Front errors in /admin/config are now OK

❓ Question
- Does it apply only to 404 errors ? (it should)

⚠️ Warning / Thoughts
- I've just thought about /error CMS page not existing yet. In that case, it'll go 404, and if errorBehavior.data = redirectPageError, it'll loop by trying to redirect on itself while not existing. We should exclude 404 processing.

❌ Errors (tested on dev-bootik2)
- redirectPageError does not apply
- displayCMSPageError does not apply
- only "Error 404 CMS Page not found" is displayed

ithiame:
It apply now on 404 error.

When error page doesn't exist is handled

Tirodem:
I confirm it applies on 404 error ✅ 

2 things :
🙇 my bad, I did a c&p mistake : headband custom message is meant to be with "Redirect to page" and not "Display CMS page error"
💡 to avoid a 404 loop we should assure that the target URL for "Redirect to page" is an existing CMS page, otherwise, redirect to / with the headband (this field is only meant for internal target URL, not external one)

src/lib/server/renderErrorPage.ts
coyotte508, line 1: This file can be removed

src/routes/(app)/[slug]/+page.server.ts
```diff
@@ -1,7 +1,7 @@
 import { collections } from '$lib/server/database';
-import { error } from '@sveltejs/kit';
 import { omit } from 'lodash-es';
 import { cmsFromContent } from '$lib/server/cms.js';
+import { error } from '@sveltejs/kit';
 
 export async function load({ params, locals }) {
 	const cmsPage = await collections.cmsPages.findOne(
```
Comment (coyotte508) [inline src/routes/(app)/[slug]/+page.server.ts:7]:
```suggestion
	let cmsPage = await collections.cmsPages.findOne(
```

src/routes/(app)/[slug]/+page.server.ts
```diff
@@ -23,7 +23,33 @@ export async function load({ params, locals }) {
 	);
 
 	if (!cmsPage) {
-		throw error(404, 'CMS Page not found');
+		const errorPages = await collections.cmsPages.findOne(
```
Comment (coyotte508) [inline src/routes/(app)/[slug]/+page.server.ts:26]:
```suggestion
		cmsPage = await collections.cmsPages.findOne(
```

src/routes/(app)/[slug]/+page.server.ts
```diff
@@ -23,7 +23,33 @@ export async function load({ params, locals }) {
...
+			}
+		);
+
+		if (errorPages) {
+			return {
+				cmsPage: omit(errorPages, ['content']),
+				cmsData: cmsFromContent(errorPages.content, locals),
+				layoutReset: errorPages.fullScreen
+			};
+		} else {
```
Comment (coyotte508) [inline src/routes/(app)/[slug]/+page.server.ts:50]:
```suggestion
		if (!cmsPage) {
```

src/routes/(app)/[slug]/[sub]/[...catchall]/+page.server.ts
```diff
@@ -1,5 +1,34 @@
+import { cmsFromContent } from '$lib/server/cms';
+import { collections } from '$lib/server/database';
 import { error } from '@sveltejs/kit';
+import { omit } from 'lodash-es';
 
-export async function load() {
-	throw error(404, 'Page not found');
+export async function load({ locals }) {
+	const errorPages = await collections.cmsPages.findOne(
```
Comment (coyotte508) [inline src/routes/(app)/[slug]/[sub]/[...catchall]/+page.server.ts:7]:
```suggestion
	const errorPage = await collections.cmsPages.findOne(
```

coyotte508 (approved):

coyotte508 (changes requested):
you need to add +page.svelte for the catchall case

coyotte508 (approved):

PR #1235, opened 2024-04-29
redirect to cms page product

ithiame:


src/routes/(app)/product/[id]/+layout.server.ts
```diff
@@ -0,0 +1,98 @@
+import { collections } from '$lib/server/database';
+import { error } from '@sveltejs/kit';
+import { cmsFromContent } from '$lib/server/cms';
+import type { Product } from '$lib/types/Product';
+import { omit } from 'lodash-es';
+
+export const load = async ({ params, locals }) => {
+	const product = await collections.products.findOne<
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+layout.server.ts:8]: instead of `.findOne` you can just `.count({_id: ...})` to see if the product exists

You certainly don't need the whole projection.

src/routes/(app)/product/[id]/+layout.server.ts
```diff
@@ -0,0 +1,98 @@
...
+
+		if (errorPage) {
+			return {
+				cmsPage: omit(errorPage, ['content']),
+				cmsData: cmsFromContent(errorPage.content, locals),
+				layoutReset: errorPage.fullScreen
+			};
+		} else {
+			throw error(404, 'Page not found');
+		}
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+layout.server.ts:96]: No need to throw? We still throw in `+page.server.ts`

src/routes/(app)/product/[id]/+layout.server.ts
```diff
@@ -0,0 +1,98 @@
...
+					hideFromSEO: 1
+				}
+			}
+		);
+
+		if (errorPage) {
+			return {
+				cmsPage: omit(errorPage, ['content']),
+				cmsData: cmsFromContent(errorPage.content, locals),
+				layoutReset: errorPage.fullScreen
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+layout.server.ts:92]: I don't think this is used

coyotte508 (commented):
Looks good with the change! (if it works!)

No need to fetch the product from DB, especially with the complicated projection. You should have noticed that! :)

Issue #1237, opened 2024-04-30
📱 Add a "Mobile behavior" selector on /admin/layout


Tirodem:
In the past, I did this commit to replace `<meta name="viewport" content="width=device-width" />` with `<meta name="viewport" content="width=1000" />` with the greenlight of our current (at the time) pilot to have a decent display and spare resources on mobile display workload :
https://github.com/B2Bitcoin/beBOP/commit/ed018822b236a3882771430f8764f045452eafa8

Today, for various purposes, we need to bench mobile display and allow some pilots to try it.

So, I'd like to have this kind of selector in /admin/layout :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/2eaaa0e2-ff5a-42d9-8c33-f5ae2b5cc4ec)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/f2857bc0-0386-4cce-9fac-9ed3dcfc9fb4)
- If Viewport width is updated : apply the new value on `<meta name="viewport" content="width=newvalue" />` (raise front error if value > 1000 ; "Invalid value")
  - on save, display warning `confirm("Mobile-friendly display is still under construction. Viewport width is made to display desktop-like experience on mobile and small devices with a low resolution. Favored value is 1000 for now. Are you really sure you want to change this ?")` and save setting only on confirm.
- If 1st option is selected : apply `<meta name="viewport" content="width=XXX" />` for everyone
- if 2st option only  is selected : apply  `<meta name="viewport" content="width=device-width" />` for employees (POS included) and `<meta name="viewport" content="width=XXX" />` for visitors
  - on save, display warning `confirm("Mobile-friendly display is still under construction. Are you sure ? This option is meant to be use to check your be-BOP with mobile specific display as an employee. For now, it's mean to be a temporary, and can reduce useability of the back-office to your team and POS session. Are you really sure ?")` and save setting only on confirm.
- if 3rd option only is selected : apply  `<meta name="viewport" content="width=device-width" />` for visitors and `<meta name="viewport" content="width=XXX" />` for employees (POS included)
  - on save, display warning `confirm("Mobile-friendly display is still under construction. Are you sure ? This option is meant to be use to give your visitors a mobile-friendly experience. For now, it's mean to be a temporary, and can degrade display of CMS and other pages of user journey. Are you really sure ?")` and save setting only on confirm.
- if 4th option is selected : apply  `<meta name="viewport" content="width=device-width" />` for everyone
[...]

PR #1238, opened 2024-04-30
📱 Add a 'Mobile behavior' selector on /admin/layout #1237

ithiame:
Add a "Mobile behavior" selector on /admin/layout #1237

### DIFF
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -14,6 +14,9 @@ export async function load(event) {
 			runtimeConfig.websiteTitle,
 		websiteShortDescription:
 			runtimeConfig[`translations.${event.locals.language}.config`]?.websiteShortDescription ||
-			runtimeConfig.websiteShortDescription
+			runtimeConfig.websiteShortDescription,
+		viewportContentWidth: runtimeConfig.viewportContentWidth,
+		viewportFor: runtimeConfig.viewportFor,
+		roleId: event.locals.user?.roleId
 	};
 }
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -7,16 +7,17 @@ export async function load(event) {
 			case 'everyone':
 				return 'width=device-width';
 			case 'employee':
-				return event.locals.user?.roleId !== CUSTOMER_ROLE_ID
+				return event.locals.user?.roleId !== CUSTOMER_ROLE_ID ||
+					event.locals.user?.roleId !== undefined
 					? 'width=device-width'
-					: runtimeConfig.viewportContentWidth;
+					: runtimeConfig.viewportContentWidth.toString();
 			case 'visitors':
-				return event.locals.user?.roleId !== POS_ROLE_ID &&
-					event.locals.user?.roleId !== SUPER_ADMIN_ROLE_ID
+				return event.locals.user?.roleId === undefined ||
+					event.locals.user?.roleId === CUSTOMER_ROLE_ID
 					? 'width=device-width'
-					: runtimeConfig.viewportContentWidth;
+					: runtimeConfig.viewportContentWidth.toString();
 			default:
-				return runtimeConfig.viewportContentWidth;
+				return runtimeConfig.viewportContentWidth.toString();
 		}
 	})();
 
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -1,23 +1,24 @@
 import { runtimeConfig, runtimeConfigUpdatedAt } from '$lib/server/runtime-config';
-import { CUSTOMER_ROLE_ID } from '$lib/types/User';
+import { CUSTOMER_ROLE_ID, POS_ROLE_ID, SUPER_ADMIN_ROLE_ID } from '$lib/types/User';
 
 export async function load(event) {
-	let viewportWidth;
-	if (runtimeConfig.viewportFor === 'everyone') {
-		viewportWidth = 'width=device-width';
-	} else if (
-		event.locals.user?.roleId !== CUSTOMER_ROLE_ID &&
-		runtimeConfig.viewportFor === 'employee'
-	) {
-		viewportWidth = 'width=device-width';
-	} else if (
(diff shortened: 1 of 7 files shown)


src/routes/+layout.svelte
```diff
@@ -18,6 +19,15 @@
...
 	<title>{data.websiteTitle}</title>
+	{#if data.roleId !== CUSTOMER_ROLE_ID && data.viewportFor === 'employee'}
+		<meta name="viewport" content="width=device-width" />
+	{:else if data.roleId === CUSTOMER_ROLE_ID && data.viewportFor === 'visitors'}
+		<meta name="viewport" content="width=device-width" />
+	{:else if data.viewportFor === 'everyone'}
+		<meta name="viewport" content="width=device-width" />
+	{:else}
+		<meta name="viewport" content="width={data.viewportContentWidth}" />
+	{/if}
```
coyotte508, line 30: Let's move this to the backend, the backend should  send `viewportWidth` depending on the role.

Here we should only have : 
```suggestion
	<meta name="viewport" content="width={data.viewportWidth}" />
```

This means that you will need to send both `viewportFor` and `viewportContentWidth` in the admin's `+page.server.ts`

src/routes/+layout.server.ts
```diff
@@ -1,6 +1,24 @@
...
+	) {
+		viewportWidth = 'width=device-width';
+	} else if (
+		event.locals.user?.roleId === CUSTOMER_ROLE_ID &&
+		runtimeConfig.viewportFor === 'visitors'
+	) {
+		viewportWidth = 'width=device-width';
+	} else {
+		viewportWidth = runtimeConfig.viewportContentWidth;
+	}
```
coyotte508, line 20: use a `switch`:

```ts
const viewPortWidth = () => ({
  switch (runtimeConfig.viewportFor) {
    case ...: return ... ? ... : ...;
    ...
  }
})()
```

`switch` has better type-checking.

Also makes the logic clearer, I think some cases are wrong with the current logic (for example `'visitors` should include when `roleId` is `undefined`)

Tirodem:
@ithiame :
❌  one too many "width=" :
image
❌ I cannot save when selector value is "everyone" (no error on console, "OK" on confirm nothing happens) ; it only saves when I click "cancel" on confirm, in opposition of the other values 🤔

src/routes/+layout.server.ts
```diff
@@ -1,23 +1,24 @@
...
+		switch (runtimeConfig.viewportFor) {
+			case 'everyone':
+				return 'width=device-width';
+			case 'employee':
+				return event.locals.user?.roleId !== CUSTOMER_ROLE_ID
+					? 'width=device-width'
+					: runtimeConfig.viewportContentWidth;
+			case 'visitors':
+				return event.locals.user?.roleId !== POS_ROLE_ID &&
+					event.locals.user?.roleId !== SUPER_ADMIN_ROLE_ID
```
coyotte508, line 15: Some employees can have another role id.

You should check instead:

- roleId === undefined || roleId = CUSTOMER_ROLE_ID

src/routes/+layout.server.ts
```diff
@@ -1,23 +1,24 @@
...
-		viewportWidth = 'width=device-width';
-	} else {
-		viewportWidth = runtimeConfig.viewportContentWidth;
-	}
+	const viewportWidth = (() => {
+		switch (runtimeConfig.viewportFor) {
+			case 'everyone':
+				return 'width=device-width';
+			case 'employee':
+				return event.locals.user?.roleId !== CUSTOMER_ROLE_ID
```
coyotte508, line 10: Guests (not logged in) have `roleId = undefined`, with this code, they have the same view as employees

src/routes/+layout.server.ts
```diff
@@ -1,6 +1,26 @@
...
+import { CUSTOMER_ROLE_ID } from '$lib/types/User';
 
 export async function load(event) {
+	const viewportWidth = (() => {
+		switch (runtimeConfig.viewportFor) {
+			case 'everyone':
+				return 'width=device-width';
+			case 'employee':
+				return event.locals.user?.roleId !== CUSTOMER_ROLE_ID ||
+					event.locals.user?.roleId !== undefined
```
coyotte508, line 11: This condition will always be `true`

src/routes/+layout.server.ts
```diff
@@ -1,6 +1,26 @@
...
+					? 'width=device-width'
+					: runtimeConfig.viewportContentWidth.toString();
+			case 'visitors':
+				return event.locals.user?.roleId === undefined ||
+					event.locals.user?.roleId === CUSTOMER_ROLE_ID
+					? 'width=device-width'
+					: runtimeConfig.viewportContentWidth.toString();
```
coyotte508, line 18: It's the opposite no?

Tirodem:
- "Everyone" selection is now ok ✅ 
- Selecting "Employee" gives `width=device-width` to employees but also to non-employees ❌ (cf https://github.com/B2Bitcoin/beBOP/pull/1238#discussion_r1585374423 )
- About https://github.com/B2Bitcoin/beBOP/pull/1238#discussion_r1585374869 , functional behavior seems good 🤔 (when selecting "Visitors", visitors have `width=device-width` and employees have the standard view

For personal notes & future documentation, about viewport width :
- header / footer go mobile when width < 640px
- product page goes mobile when width < 768px
- /checkout page goes mobile when width < 768px
- /order page goes mobile when width < 768px
- /cart doesn't seem to have significant change

Tirodem:
+ @ithiame as you removed the 2nd "width=" for width=device-width, we don't have "width" anymore for viewport size, so it doesn't work :
image
I'm pushing a quick modification to allow @CypherJack to do some tests today and tomorrow.

ithiame:
My Bad 
let me change it quickly

src/routes/+layout.server.ts
ithiame: it's correct i think 

??

coyotte508 (approved):

Issue #1244, opened 2024-05-02
📱 Create a mobile-dedicated countdown widget variation


Tirodem:
Following #1238 context,

TBD

Current behavior with `width=device-width` :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/b8023380-e58a-4855-90a1-9820c2f00937)
OK tier but not very elegant 🤷‍♂️ 

Countdown mobile-dedicated template suggestion :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/4af8917c-ca47-4e7e-a30b-72928b193141)

PR #1259, opened 2024-05-10
Create a mobile-dedicated countdown widget variation #1244

ithiame:


ithiame:
Befor:
image

After:
image

Tirodem:
Functional tests OK for me ✅

src/lib/components/CountdownWidget.svelte
```diff
@@ -18,7 +18,7 @@
 	});
 </script>
 
-<div class="flex flex-row gap-4 {className}">
+<div class="hidden sm:contents gap-4 {className}">
 	<div class="flex flex-row w-full tagWidget tagWidget-main grow pr-5">
```
coyotte508, line 22: You can edit class names in two places rather than copy/paste everything

For example here:

```suggestion
	<div class="flex flex-col sm:flex-row w-full tagWidget tagWidget-main grow pr-5">
```

Same principle with the `justify-end` (`justify-end sm:justify-normal`)

This PR can be a 2-lines change

src/lib/components/CountdownWidget.svelte
```diff
@@ -18,33 +18,31 @@
 	});
 </script>
 
-<div class="flex flex-row gap-4 {className}">
```
coyotte508, line 21: The surrounding div has been removed, is it intended?

ithiame: reverted

coyotte508 (approved):

Issue #1257, opened 2024-05-09
📱 Force viewport content width on /admin


Tirodem:
Following https://github.com/B2Bitcoin/beBOP/pull/1238 & #1248 .
Whatever the setting in /admin/layout for mobile display, fix this on the /admin/* and /pos routes for now : `<meta name="viewport" content="width=1000">` when device-width =< 1000px

PS : actually, I say 1000px per habit because it's the default value, but stick to the mobile display parameter from /admin/layout (aka runtimeConfig.viewportContentWidth) in order to keep it dynamic.

PR #1260, opened 2024-05-10
Force viewport content width on /admin #1257

ithiame:


coyotte508 (approved):

Tirodem:
Here's what I have on /admin/* & /pos pages :
image
Functional behavior is OK and I have content of width=1000, but is there a way that having 2 viewport meta balises leads to some issues ?

ithiame:
Don't think it leads to issue, it takes the last one. @coyotte508 ??

coyotte508:
We already have the same thing on /pos/session so it's fine at the moment

Issue #1185, opened 2024-04-02
🦋 Allow Superadmin (and only Superadmin) account to edit its contact information on /admin/arm


Tirodem:
Currently, super-admin account has no contact information (unless you insert them in database).
And the nostr / email fields are read-only in /admin/arm :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/982811ab-075e-4f63-a799-4548c294ce09)
Following https://github.com/B2Bitcoin/beBOP/pull/1184/files#r1548236780 , we need for the superadmin account to have contact informations.
So, I'd like to remove "disabled" attribute on fields recoveryEmail and recoveryNpub **only for the superadmin account** on /admin/arm .
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/a5668773-002e-46cf-b432-15cb298aa8e5)

PR #1279, opened 2024-05-29
Allow Superadmin (and only Superadmin) account to edit its contact in…

ithiame:
…formation on /admin/arm #1185

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts
@@ -48,25 +48,40 @@ export const actions = {
 		if (user.roleId !== SUPER_ADMIN_ROLE_ID && !parsed.recoveryEmail && !parsed.recoveryNpub) {
 			throw error(400, 'You must provide a recovery email or npub');
 		}
-
-		await collections.users.updateOne(
-			{ _id: user._id },
-			{
-				$set: {
-					login: parsed.login ?? user.login,
-					...(parsed.alias && { alias: parsed.alias }),
-					recovery: {
-						...(parsed.recoveryEmail && { email: parsed.recoveryEmail }),
-						...(parsed.recoveryNpub && { npub: parsed.recoveryNpub })
-					},
-					...(parsedAdmin.alias && { alias: parsedAdmin.alias }),
-					disabled: parsed.status === 'disabled',
-					roleId: parsed.roleId ?? user.roleId
+		if (user.roleId === SUPER_ADMIN_ROLE_ID) {
+			await collections.users.updateOne(
+				{ _id: user._id },
+				{
+					$set: {
+						...(parsedAdmin.alias && { alias: parsedAdmin.alias }),
+						recovery: {
+							...(parsed.recoveryEmail && { email: parsed.recoveryEmail }),
+							...(parsed.recoveryNpub && { npub: parsed.recoveryNpub })
+						}
+					}
 				}
-			}
-		);
+			);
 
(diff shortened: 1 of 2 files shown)


coyotte508 (changes requested):
This PR: only super admin can edit the information of anyone

Expected behavior: admins can edit the information of other users, and super admin can edit the information of anyone

coyotte508 (commented):
Did you test the PR work?

Is there no backend check to prevent super admin role being edited?

ithiame:
Yes Tested

The backend check doesn't prevent superadmin to edit email && npub

super admin role change is disabled in front

coyotte508:
> Yes Tested
> 
> The backend check doesn't prevent superadmin to edit email && npub
> 
> super admin role change is disabled in front

Are you sure?

https://github.com/B2Bitcoin/beBOP/blob/24a387dc05f1050938259794f522a6663f721888/src/routes/(app)/admin%5B%5Bhash%3Dadmin_hash%5D%5D/arm/user/%5Bid%5D/%2Bpage.server.ts#L51-L62

This code should make only the alias editable

src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts
```diff
@@ -48,36 +48,25 @@ export const actions = {
...
-							...(parsed.recoveryNpub && { npub: parsed.recoveryNpub })
-						},
-						disabled: parsed.status === 'disabled',
-						roleId: parsed.roleId
-					}
+		await collections.users.updateOne(
+			{ _id: user._id },
+			{
+				$set: {
+					login: parsed.login ?? user.login,
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:56]: This allows anyone to update the super admin's login & recovery info, including non-super admins

coyotte508 (changes requested):
Please test locally without frontend checks that

- a super admin can edit its login & recovery info
- a non-super admin **cannot** edit the super admin's login & recovery info

src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts
```diff
@@ -48,36 +42,27 @@ export const actions = {
...
-			await collections.users.updateOne(
-				{ _id: user._id },
-				{
-					$set: {
-						...(parsedAdmin.alias && { alias: parsedAdmin.alias })
-					}
-				}
-			);
+		if (user.roleId === SUPER_ADMIN_ROLE_ID && locals.user?.roleId !== SUPER_ADMIN_ROLE_ID) {
+			throw error(400, 'You are not allowed to edit admin information');
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:46]: Small regression - before other admins could edit the alias of the super admin. But it's ok imo cc @Tirodem , up to you

coyotte508 (approved):

src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts
Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:46]: If the regression forbid for non SU admin to edit SU admin login, it's not a regression, it's a good feature 👌

Comment (coyotte508) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:46]: just the alias that showed up in the UI

Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:46]: oh, damn, didn't understand that
/arm is a sensitive page so it's not shared to every profile (usually super-admin and another admin+) but it would be great to hide superadmin login for non-super-admin users with access to /arm
i'll keep this to create an improvement ticket

PR #1297, opened 2024-06-03
icon menu to be same color as header tab

ithiame:


ithiame:
Before : image
image


After : image
image

coyotte508 (approved):

Issue #1298, opened 2024-06-03
🏪Create pos-touch interface


Tirodem:
🏪From #1291

- [ ] create a route to /pos-touch

- [ ] On /pos , add a link to /pos-touch (target = new tab)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/68738889-8e4d-4001-932e-0b15a2ea15d0)

- [ ] Restrict access to /pos-touch to POS accounts

- [ ] Create /pos-touch page layer
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/e3ded964-4829-426b-af1e-e4e60cc68cce)
  - 

- [ ] On /pos-touch access, sw

Tirodem:
Finally done entierely.

cc @gntw 

<img width="1200" height="675" alt="Image" src="https://github.com/user-attachments/assets/5eb07d73-d049-4397-bd49-d793fa90e6ff" />

PR #1310, opened 2024-06-04
Create pos-touch interface #1298

ithiame:


src/routes/(app)/pos/touch/+page.server.ts
```diff
@@ -0,0 +1 @@
+export const load = async ({}) => {};
```
coyotte508, line 1: Should add `layoutReset: true` in the data, to avoid rendering all the html for the regular layout (even if it's hidden)

coyotte508 (commented):
Let me know when the PR is ready to review, mark it as draft otherwise

Issue #1311, opened 2024-06-04
🏪Include /pos/touch into theme configurator


Tirodem:
🏪From https://github.com/B2Bitcoin/beBOP/issues/1291 & https://github.com/B2Bitcoin/beBOP/issues/1298

We need to integrate this interface to theme configuration.

So, we need in **POS Touchscreen interface** :
- Category main CTA background color
- Category main CTA font
- Category main CTA font color
- Product list main CTA background color
- Product list main CTA font
- Product list main CTA font color
- Ticket menu background color
- Ticket menu font
- Ticket menu font color
- Action main CTA background color
- Action main CTA font
- Action  main CTA font color
- Action secondary CTA background color
- Action secondary CTA font
- Action secondary CTA font color
- Action cancel CTA background color
- Action delete CTA background color

And add these default values on new theme creation :
- Category main CTA background color : 310d40
- Category main CTA font : Outfit
- Category main CTA font color : white
- Product list main CTA background color : ab47a5
- Product list main CTA font : Outfit
- Product list main CTA font color : black
- Ticket menu background color : b6b6b6
- Ticket menu font : Outfit
- Ticket menu font color : black
- Action main CTA background color : 310d40
- Action main CTA font : Outfit
- Action  main CTA font color : white
- Action secondary CTA background color : ab47a5
- Action secondary CTA font : Outfit
- Action secondary CTA font color : white
- Action cancel CTA background color : eee100
- Action delete CTA background color : ee0000

Tirodem:
Default values to push on existing theme with no values or new themes :
```
{
  "touchScreen": {
    "category": {
      "cta": {
        "backgroundColor": {
          "light": "#310d40",
          "dark": "#310d40"
        },
        "color": {
          "light": "#ffffff",
          "dark": "#ffffff"
        }
      },
      "fontFamily": "Outfit"
    },
    "product": {
      "cta": {
        "backgroundColor": {
          "light": "#ab47a5",
          "dark": "#ab47a5"
        },
        "color": {
          "light": "#ffffff",
          "dark": "#ffffff"
        },
        "fontFamily": "Outfit"
      },
      "secondaryCTA": {
        "backgroundColor": {
          "light": "#ffffff",
          "dark": "#ffffff"
        },
        "color": {
          "light": "#b6b6b6",
          "dark": "#b6b6b6"
        },
        "fontFamily": "Outfit"
      }
    },
    "ticket": {
      "menu": {
        "backgroundColor": {
[...]

PR #1312, opened 2024-06-04
Include /pos/touch into theme configurator #1311

ithiame:
Include /pos/touch into theme configurator #1311


src/app.css
coyotte508, line 1: Do you think it's possible to do a separate CSS file and import it in `pos/touch/+layout.svelte`?

Same for the touch screen variables

ithiame: committed here 
https://github.com/B2Bitcoin/beBOP/pull/1312/commits/8ce5c8f0c85c7b101222f3dd1add2494a70e021a

src/routes/(app)/pos/touch/+page.svelte
```diff
@@ -0,0 +1,40 @@
+<script lang="ts">
+	import './touch.css';
+</script>
+
+<svelte:head>
+	<meta name="viewport" content="width=1024" />
+</svelte:head>
+<main class="fixed top-0 bottom-0 right-0 left-0 bg-white p-4">
```
coyotte508, line 8: This should be in a `+layout.svelte` - and maybe the layout reset can be in a `+layout.server.ts`

But we can handle this in a follow up PR, just keep it in mind when you start adding other pages - to create a `+layout.svelte`/`+layout.server.ts`

coyotte508 (approved):

Issue #1304, opened 2024-06-03
🏪 Set /pos/touch display tags on admin


Tirodem:
🏪From https://github.com/B2Bitcoin/beBOP/issues/1291 & https://github.com/B2Bitcoin/beBOP/issues/1298

On touchscreen specific interface (/pos/touch), a seller should be able to set some product tags for quick access.

![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/4155d11d-e7e6-42d4-a930-29f24955b012)

- [ ] Create a /admin/pos page
- [ ] Add a link to it in admin header under Config > POS
- [ ] Allow to select (add / remove) existing product tags on configuration

Selected tags will be displayed here on /pos/touch, in addition to "Favorites" and "All product" :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/dd0e7683-5587-4d35-90c9-a20f7cd0d91e)

Note that pos-favorite should be excluded from the selection, as there's a default CTA for that ( #1301 )

PR #1319, opened 2024-06-07
Set /pos/touch display tags on admin #1304

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts
```diff
@@ -0,0 +1,48 @@
...
+	default: async function ({ request }) {
+		const formData = await request.formData();
+		const result = z
+			.object({
+				posTouchTag: z.string().array()
+			})
+			.parse({
+				posTouchTag: JSON.parse(String(formData.get('posTouchTag'))).map(
+					(x: { value: string }) => x.value
+				)
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts:29]: Find a better way of sending the data from the frontend to avoid this

coyotte508 (approved):

Issue #1322, opened 2024-06-10
🦋 Create default visual for products with no picture


Tirodem:
Picture to use when a product doesn't not have image :
- [ ] on product page
- [ ] on product widget
- [ ] on cart, checkout
- [ ] on order
- [ ] on /pos/touch pages

![unknown](https://github.com/B2Bitcoin/beBOP/assets/50206014/df32ba95-824a-4f14-b87c-71a83c4c8726)

Normally a product cannot not have any picture but it'll help on dev environment and it's good to have some default resource to display, and it'll help in the future for migration scripts or bulk load of saved products.

PR #1328, opened 2024-06-10
Create default visual for products with no picture #1322

ithiame:


src/lib/components/ProductWidget/ProductWidgetPOS.svelte
```diff
@@ -13,7 +14,11 @@
 <div class="touchScreen-product-cta flex flex-row {className}">
 	<div>
 		<a href="/product/{product._id}">
-			<PictureComponent picture={pictures[0]} class="object-contain h-24 w-24" />
+			{#if pictures.length > 0 && pictures[0]}
+				<PictureComponent picture={pictures[0]} class="object-contain h-24 w-24" />
+			{:else}
+				<img class="object-contain h-24 w-48" src={DEFAULT_PICTURE} alt="default-img" />
```
coyotte508, line 20:
```suggestion
				<img class="object-contain h-24 w-48" src={DEFAULT_PICTURE} alt="" />
```

coyotte508 (approved):

src/lib/components/ProductWidget/ProductWidgetPOS.svelte
ithiame: @coyotte508 sometime the image doesn't exist in the S3
image

Is there a function or something you already implemented that check if the image exist int the S3 ?

coyotte508: well it should always exist. You can add event handlers in `Picture` (eg `on:error`) or a fallback picture. But it should always exist...

Tirodem: Functionally, it should, but as said earlier, I want the /pos/touch interface to be used even with a broken/down S3.
In this particular case, it's a remainder of import / export feature testing, but that's what gave me the idea of a fallback picture.
And, in the future, maybe we'll allow product creation with no picture for restaurants particular usage, so...
(Here's for the context)

src/lib/assets/default-product.png
coyotte508, line 1: Can we use a SVG image instead?

It should be much more lightweight for such a simple image

coyotte508 (approved):

PR #1330, opened 2024-06-11
Pos product addition error

ithiame:


coyotte508 (commented):
Just do `alert(errorMessage);`

Let's not add components and what not just to handle an error message in one specific case.

ithiame:
I tried to do as specified on the issue.

But if alert is ok it's easier

coyotte508:
ah my bad - we can go back to the popup but then we need a more generic approach - up to you @Tirodem

Tirodem:
@coyotte508 It's good as it is, I'll merge like this if it's OK with you and I'll do a ticket later with a more clean long-term perspective for the new required component for popin.

coyotte508 (approved):

PR #1351, opened 2024-07-01
Remove mandatory currency control

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/challenge/new/+page.server.ts
```diff
@@ -42,7 +42,7 @@ export const actions: Actions = {
 				),
 				goalAmount: data.get('goalAmount'),
 				mode: data.get('mode'),
-				currency: data.get('currency'),
+				currency: data.get('currency') ?? data.append('currency', 'BTC'),
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/challenge/new/+page.server.ts:45]:
```suggestion
				currency: data.get('currency') ?? undefined,
```

https://developer.mozilla.org/fr/docs/Web/API/FormData/append

It returns `undefined`...

coyotte508 (approved):

Issue #1186, opened 2024-04-02
📡 Allow a POS seat to overwrite its IP country on POS Session with a manually selected country


Tirodem:
# Context
With a VAT behavior where VAT rate = VAT rate from customer country, we had to make sales on Belgium, with specific articles and belgian VAT.
Onsite, we had no wiFi, and through 4G, we had a french IP (it was a french SIM card). We tried to go through a belgian VPN, but it was very flaky and not a viable option.
So, in order to sell during this belgian week-end, we had to change the VAT behavior and put the main website in maintenance as beBOP didn't locate us in Belgium with our only connexion options.

# Behavior
On /pos, add an option "Overwrite IP country for this POS Session with selected country" + display the current country deduced from IP.
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/2b471f51-c74e-4948-8bd2-c54845da23c7)
Once checked a country selected, display a warning :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/a5d6d473-82b6-4161-a4f7-26c6cd360686)
English default text :
"Are you sure ? This can have impact on law compliance with your sales and invoices. Do it only if you're sure about it and if the owner requested it."
French default text :
"Êtes vous sur ? Ceci peut avoir un impact sur la légalité de vos ventes et leur facturation. Ne le faites que si vous êtes sur de vous, et que si le propriétaire l'a demandé."

**Then**, for the rest of the POS session, consider the POS seat to be located in the chosen country instead of the one from the IP (I think it's easier to say that done, despite being just one line of description).

The IP country overwrite should be as displayed previously :
"This update will be active only during the current session. If you change browser, computer or have to login again, it'd be resetted."

coyotte508:
Yes it's possible, we can have it for the session or associated to the user (eg in personal info)

Tirodem:
If both are possible, I feel it's better in session (the documentation will
recommand 1 day of session expiration for POS Seat, that'll avoir POS
seater to forget to change the account country once back in it).

PR #1352, opened 2024-07-01
Allow a POS seat to overwrite its IP country on POS Session with a ma…

ithiame:
…nually selected country #1186


src/lib/types/Session.ts
```diff
@@ -16,4 +16,5 @@ export interface Session extends Timestamps {
 		name: string;
 		id: string;
 	}>;
+	clientIp?: string;
```
coyotte508, line 19: Let's store `pos?: { countryCodeOverwrite?: ...}` instead

Let's not store a client ip in the session or mess with ips.

The `ipFromCountry` is a nice bit of code, and was fun to write probably, but boring code/simplest solution is better for long term :pray:

ithiame: So what do you mean ? There's no need to store the IP ?
If so how are we gonna the clientIP ?


(I took Inspiration on `countryFromIP`)

src/hooks.server.ts
```diff
@@ -216,6 +216,10 @@ const handleGlobal: Handle = async ({ event, resolve }) => {
 		event.locals.email = session.email;
 		event.locals.npub = session.npub;
 		event.locals.sso = session.sso;
+		event.locals.clientIp = session.clientIp;
+		event.locals.countryCode = event.locals.clientIp
+			? countryFromIp(event.locals.clientIp)
+			: undefined;
```
coyotte508, line 222:
```suggestion
		if (session.pos?.countryCodeOverride) {
			event.locals.countryCode = session.pos.countryCodeOverride;
		};
```

I think that's what we want @ithiame . No need to store a fake IP, only the fake countrycode

src/routes/(app)/pos/+page.svelte
```diff
@@ -16,6 +26,23 @@
...
 		>{t('pos.adminInterface')}</a
 	>
+	<form action="?/overwrite" method="POST" on:submit={handleSubmit}>
+		<label class="checkbox-label">
+			<input type="checkbox" class="form-checkbox" name="overwriteIP" bind:checked={overwriteIP} />
+			Overwrite IP country for this POS session with selected country
+		</label>
+		<label class="form-label col-span-3">
+			Overwrite session IP country (current value {data.countryCode})
+			<select name="countryCode" class="form-input" required bind:value={data.countryCode}>
```
coyotte508, line 36:
```suggestion
			<select name="countryCode" class="form-input" required value={data.countryCode}>
```

no?

coyotte508 (approved):

Tirodem:
Nice, it works.
But, and that might be because of me : how do we remove the overwrite ? It is possible or shall we overwrite with a new country ?
I suggest to add an option "(Remove ipcountry overwrite)" at the beginning of the country list, or add a button "Remove overwrite".
@coyotte508 , WDYT ?

coyotte508:
Yes agree. we could show the "remove overwrite" button when it's already overwritten

src/routes/(app)/pos/+page.server.ts
```diff
@@ -32,7 +36,8 @@ export const load = async (event) => {
 				})) || [],
 			status: order.status
 		})),
-		countryCode: event.locals.countryCode
+		countryCode: event.locals.countryCode,
+		session: pojo(session)
```
coyotte508, line 40: Please send only the needed data, for security reasons. Not the whole session

coyotte508 (approved):

Issue #1354, opened 2024-07-08
🦋 Add an option on POS account checkout : "


Tirodem:
Following production use case on pilot, we need a new optional option.

1/ in admin/config :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/f2fe368f-b007-42e6-a215-50ff1d8e66be)

2/ in /checkout
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/d39f40ff-a5bb-4441-b7b0-0f5b0b4543b7)
when enabled :
- don't display delivery address on receipt
- don't display "billing address" on billing address on receipt
- pre-check POS option "free delivery fees as a POS sale"
- synchronise this checkbox with "free delivery fees as a POS sale" checkbox (if manuallly unchecked, uncheck to, if manuallly rechecked, recheck too)

coyotte508:
Change "full-paid" to "fully-paid" and "shop" to "on-location" in my opinion

And add `onLocation` flag in DB cc @ithiame

PR #1356, opened 2024-07-09
Add an option on POS account checkout :  #1354

ithiame:


coyotte508:
Hmm currently the checkbox is exactly the same as "free delivery fees as POS"

Should there be a difference in behavior @Tirodem ? (Why do we need two checkboxes with the same behavior on the checkout page?)

Tirodem:
This option should trigger the free POS delivery fees but also have an impact on billing receipt (with no delivery address being displayed)

coyotte508:
> This option should trigger the free POS delivery fees but also have an impact on billing receipt (with no delivery address being displayed)

Yes but if both checkboxes are always synced with each other on the checkout page there's zero difference

Tirodem:
Oh, yeah, that.
The sync must be unilateral : clicking on the new box should check the
legacy one, but seller should be able to uncheck delivery fees gratuity, to
cover some exotic functional cases

coyotte508:
So to be clear:

- ticking new checkbox => free delivery is ticked
- unticking new checkgox => free delivery is unticked
- ticking free delivery => new checkbox doesn't change
- unticking free delivery => new checkbox doesn't change

src/routes/(app)/admin[[hash=admin_hash]]/config/+page.svelte
```diff
@@ -165,6 +165,16 @@
 		/>
 		Limit product line per cart (for law purpose and small business )
 	</label>
+	<label class="checkbox-label">
+		<input
+			type="checkbox"
+			name="defaultOnLocation"
+			class="form-checkbox"
+			checked={data.defaultOnLocation}
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/config/+page.svelte:173]: the "free POS delivery fees" should automatically be checked initially if this checkbox is checked

also when this checkbox changes, the other checkbox should change in consequence

Tirodem:
> So to be clear:
> 
> * ticking new checkbox => free delivery is ticked
> * unticking new checkgox => free delivery is unticked
> * ticking free delivery => new checkbox doesn't change
> * unticking free delivery => new checkbox doesn't change

exactly ✅

coyotte508 (dismissed):

src/routes/(app)/checkout/+page.server.ts
```diff
@@ -335,7 +336,13 @@ export const actions = {
...
 			);
 		}
-
+		const physicalFullyPaid = z
+			.object({
+				onLocation: z.boolean({ coerce: true }).default(false)
+			})
+			.parse({
+				onLocation: formData.get('onLocation')
+			});
```
coyotte508, line 345: This should only work for POS accounts

coyotte508:
There is also the issue that `runtimeConfig.deliveryFees.allowFreeForPOS` needs to be enabled before a POS user can offer delivery fees.

How does this interact with the new `runtimeConfig.defaultOnLocation` setting @Tirodem ?

Tirodem:
> There is also the issue that `runtimeConfig.deliveryFees.allowFreeForPOS` needs to be enabled before a POS user can offer delivery fees.
> 
> How does this interact with the new `runtimeConfig.defaultOnLocation` setting @Tirodem ?

When `runtimeConfig.deliveryFees.allowFreeForPOS` is disabled, display the option but don't seek for the free pos delivery fees option auto-tick.

Can you give me some memory refresh about `runtimeConfig.defaultOnLocation` ?

src/routes/(app)/checkout/+page.server.ts
```diff
@@ -447,7 +457,9 @@ export const actions = {
 					...(agreements.isVATNullForeigner && {
 						acceptedExportationAndVATObligation: agreements.isVATNullForeigner
 					})
-				}
+				},
+				...(physicalFullyPaid &&
+					physicalFullyPaid.onLocation && { onLocation: physicalFullyPaid.onLocation })
```
coyotte508, line 462:
```suggestion
				...(physicalFullyPaid?.onLocation && { onLocation: physicalFullyPaid.onLocation })
```

coyotte508 (approved):

Tirodem:
Gosh I forgot this :
image
We need to prefill the input (I'm not a big fan) OR we need to make this field optional when the option "_Default enable "this is a fully-paid on-location physical order_"is checked, otherwise, we lose part of this feature value.

PR #1357, opened 2024-07-09
fix slide block disappearing

ithiame:
Discord Issue

https://discord.com/channels/1022592041709998080/1088417396232028170/1260138572703469628

### DIFF
--- a/src/lib/components/CarouselWidget.svelte
+++ b/src/lib/components/CarouselWidget.svelte
@@ -13,16 +13,16 @@
 		currentIndex = (currentIndex + 1) % pictures.length;
 	}
 
-	let interval: number;
+	let interval: ReturnType<typeof setInterval>;
 	onMount(() => {
-		interval = Number(setInterval(nextSlide, autoplay));
+		interval = setInterval(nextSlide, autoplay);
 		return () => clearInterval(interval);
 	});
 
 	function setIndex(index: number) {
 		currentIndex = index;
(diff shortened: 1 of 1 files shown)


src/lib/components/CarouselWidget.svelte
```diff
@@ -9,31 +9,30 @@
 	let className = '';
 	export { className as class };
 	let currentIndex = 0;
-	// Function to move to the next slide
 	function nextSlide() {
 		currentIndex = (currentIndex + 1) % pictures.length;
 	}
+
 	let interval: number;
```
coyotte508, line 16:
```suggestion
	let interval: ReturnType<setInterval>;
```

And remove the `Number()` casts, it's not guaranteed to work

coyotte508 (approved):

Tirodem:
It's better, but, compared to before :
- there's no sliding effect anomyre (it's picture 1 then picture 2 then picture 3 without any transition)
- switching slides manually doesn't work anymore
At least there's no more variation in height.
What happened ? 🤔

Tirodem:
It's better but I still have issues for manual sliding.
Not working on desktop except with the "oooo" selector, not working on mobile (I'm stuck between the 1st and 4th, to see the 2nd & 3rd I have to wait).
I'll merge this because it's improving the actual but the current rendering for me is not satisfying and worse than few days / weeks ago, we still need to dig on that.

PR #1365, opened 2024-07-16
When scanning a nostr qr code, remove nostr: on notification address …

ithiame:
…input

### DIFF
--- a/src/routes/(app)/checkout/+page.svelte
+++ b/src/routes/(app)/checkout/+page.svelte
@@ -70,7 +70,9 @@
 			}
 
 			input.value = input.value.trim();
-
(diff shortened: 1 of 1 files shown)


src/routes/(app)/checkout/+page.svelte
```diff
@@ -70,7 +70,9 @@
 			}
 
 			input.value = input.value.trim();
```
coyotte508, line 72:
```suggestion
			input.value = trimPrefix(input.value.trim(), "nostr:");
```

`import { trimPrefix } from '$lib/utils/trimPrefix';`

coyotte508 (approved):

PR #1379, opened 2024-07-18
Reply for beBOP that required IP on deliveryless order #1368

ithiame:
Reply "Sorry, this beBOP required IP for orders for legal motives, orders through Nostr are disabled ; you can check why on {bebop-url}/why-collect-ip" for beBOP that required IP on deliveryless order #1368

### DIFF
--- a/src/lib/server/locks/handle-messages.ts
+++ b/src/lib/server/locks/handle-messages.ts
@@ -528,10 +528,7 @@ const commands: Record<
 				return;
 			}
 
-			if (
-				runtimeConfig.collectIPOnDeliverylessOrders &&
-				products.every((product) => !product.shipping)
-			) {
+			if (runtimeConfig.collectIPOnDeliverylessOrders) {
 				await send(
 					`Sorry, this beBOP requires an IP address or shipping address for each order, which is not possible via NostR at the moment`
 				);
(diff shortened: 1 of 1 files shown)


src/lib/server/locks/handle-messages.ts
```diff
@@ -528,6 +528,16 @@ const commands: Record<
 				return;
 			}
 
+			if (
+				runtimeConfig.collectIPOnDeliverylessOrders &&
+				products.every((product) => !product.shipping)
+			) {
+				await send(
+					`Sorry, this beBOP required IP for orders for legal motives, orders through Nostr are disabled ; you can check why on ${ORIGIN}/why-collect-ip`
```
coyotte508, line 536:
```suggestion
					`Sorry, this beBOP requires an IP address or shipping address for each order, which is not possible via NostR at the moment`
```

src/lib/server/locks/handle-messages.ts
```diff
@@ -528,6 +528,16 @@ const commands: Record<
 				return;
 			}
 
+			if (
+				runtimeConfig.collectIPOnDeliverylessOrders &&
+				products.every((product) => !product.shipping)
```
coyotte508, line 533:
```suggestion
				runtimeConfig.collectIPOnDeliverylessOrders
```

coyotte508 (approved):

PR #1381, opened 2024-07-18
Reply when trying to add a product with no stock

ithiame:
Reply "Sorry, this product is out of stock" when trying to add a product with no stock #1371


coyotte508 (approved):

coyotte508 (changes requested):
Is this PR needed?

https://github.com/B2Bitcoin/beBOP/blob/9c8fbcc2d2dbc2ef5c46673f7551e159aab04d24/src/lib/server/cart.ts#L129-L131

Should already take care of it

PR #1382, opened 2024-07-18
Reply when trying to add a product with future release date …

ithiame:
Reply "Sorry, this product cannot be ordered yet" when trying to add a product with future release date and no allowed pre-sale #1370


src/lib/server/locks/handle-messages.ts
```diff
@@ -372,6 +372,11 @@ const commands: Record<
 				return;
 			}
 
+			const isPreorder = isPreorderFn(product.availableDate, product.preorder);
+			if (!isPreorder && product.availableDate && product.availableDate > new Date()) {
+				await send('Sorry, this product cannot be ordered yet');
```
coyotte508, line 377:
```suggestion
				await send('Sorry, this product is not available yet to order');
```

coyotte508 (approved):

coyotte508 (changes requested):
Is this PR needed? 

https://github.com/B2Bitcoin/beBOP/blob/9c8fbcc2d2dbc2ef5c46673f7551e159aab04d24/src/lib/server/cart.ts#L99-L101

This should already take care of it

PR #1383, opened 2024-07-18
✨ Stripe integration

coyotte508:


Tirodem:
Passing case if OK.
I need more testing no other payment means provided through Stripe.
But it's 🚀 so far.

Tirodem:
As said in MP, we can merge if that can help you with Paypal integration.
But, note for watchers :
⚠️ don't use it for know (or at your own risks), I still have some QA to do in regard of every payment mean made available by Stripe (+ documentation heavily needed).

PR #1384, opened 2024-07-18
Reply when trying to add a standalone product with a quantity > 1

ithiame:
Reply "Sorry, you cannot order more than 1 product at a time" when trying to add a standalone product with a quantity > 1


src/lib/server/locks/handle-messages.ts
```diff
@@ -371,7 +371,10 @@ const commands: Record<
 				);
 				return;
 			}
-
+			if (product.standalone && quantity > 1) {
+				await send(`Sorry, you cannot order more than 1 product at a time`);
```
coyotte508, line 375:
```suggestion
				await send(`Sorry, you cannot order more than one of this product at a time`);
```

coyotte508 (approved):

coyotte508 (changes requested):
Is this PR needed?

https://github.com/B2Bitcoin/beBOP/blob/9c8fbcc2d2dbc2ef5c46673f7551e159aab04d24/src/lib/server/cart.ts#L163-L165

This should already take care of it (the error message is forwarded)

PR #1385, opened 2024-07-19
Reply when trying to add a product and the 'Max quantity per order' a…

ithiame:
…mount is reached #1375


coyotte508 (changes requested):
Is this PR needed?

Because `addToCartInDb` should already throw an error with the same message

PR #1387, opened 2024-07-19
Create a sales canal 'Nostr' on the sales canal selector #1386

ithiame:


src/lib/server/migrations.ts
```diff
@@ -26,7 +26,7 @@ const migrations = [
 		}
 	},
 	{
-		_id: new ObjectId('39811201e92e590e858af8ba'),
+		_id: new ObjectId('39811201e92e590e858af8bb'),
```
coyotte508, line 29: Please don't reuse an existing migration, create an additional migration at the end

src/lib/server/migrations.ts
```diff
@@ -436,6 +440,37 @@ const migrations = [
 		run: async () => {
 			await collections.pictures.dropIndex('productId_1').catch(console.error);
 		}
+	},
+	{
+		_id: new ObjectId('39811201e92e590e858af8bb'),
```
coyotte508, line 445: This is an id from the year `2000`. Use `new ObjectId()` locally to generate a new one

src/lib/server/migrations.ts
```diff
@@ -436,6 +440,37 @@ const migrations = [
...
+								visible: true,
+								canBeAddedToBasket: true
+							},
+							googleShopping: {
+								visible: true
+							},
+							nostr: {
+								visible: true,
+								canBeAddedToBasket: true
+							}
```
coyotte508, line 467:
```suggestion
						"actionSettings.nostr": {
								visible: true,
								canBeAddedToBasket: true
							}
```

src/lib/server/migrations.ts
```diff
@@ -436,6 +440,24 @@ const migrations = [
 		run: async () => {
 			await collections.pictures.dropIndex('productId_1').catch(console.error);
 		}
+	},
+	{
+		_id: new ObjectId('669a90d18bc5aaf40c863b63'),
+		name: 'Adding actionSettings nostr to products',
```
coyotte508, line 446: We should also add it to `runtimeConfig`

Otherwise there will be an error when creating a new product with default action settings - since `ProductForm` since it calls `bind:checked={product.actionSettings.nostr.visible}`

src/lib/server/runtime-config.ts
ithiame, line 1: Already done here 

??

ithiame: And I don't have error when creating product 
@coyotte508

coyotte508: - Go to main branch
- Delete the last migration from `db.migrations` 
- Edit the sales canal
- Go back to your PR's branch
- Go to http://localhost:5174/admin-test/product?/update
- 500 error

coyotte508: What you did here is change the default value, but it doesn't mean anything if there's already a value in DB

src/lib/server/migrations.ts
```diff
@@ -436,6 +440,42 @@ const migrations = [
...
+						'data.nostr': {
+							visible: true,
+							canBeAddedToBasket: true
+						}
+					}
+				},
+				{ session }
+			);
+		}
 	}
```
coyotte508, line 479: No need to create two migrations in one PR, you can merge it with the previous one

coyotte508 (approved):

Issue #1377, opened 2024-07-18
Reply (english version of) "Votre panier a atteint la taille maximale. Veuillez supprimer des lignes de votre panier pour ajouter plus d'articles." when runtineConfig._id "cartMaxSeparateItems" is already reached


ithiame:

ithiame:
@coyotte508 This one also is not needed i think ?

coyotte508:
probably not (you can test locally)

PR #1388, opened 2024-07-19
Reply  when runtineConfig._id 'cartMaxSeparateItems' is already reach…

ithiame:
…ed #1377


coyotte508 (commented):
Same question as other Nostr PRs - isn't this already handled in `addToCart`?

PR #1389, opened 2024-07-20
✨ Paypal integration

coyotte508:
When paypal is configured:

- The "card" payment method becomes available (SumUp and Stripe have priority if they're also configured). **Guest checkout** should be enabled in the paypal settings so users do not have to create a paypal account
- The "paypal" payment method becomes available

On payment, the client is redirected to paypal's page. Paypal doesn't offer a widget for paypal payments, at least not that I could easily find.


Tirodem:
"Secret" input should be secret
(same for Phoenixd http password, now I think about it)

Tirodem:
Would it be possible, on /admin/paypal, to test connexion with filled credentials through a "Test connexion" CTA, and display a "succesfully connected" / "there was an error" message in return ?

coyotte508:
> "Secret" input should be secret (same for Phoenixd http password, now I think about it)

do you mean it should be like a password input, with "***" ?

Tirodem:
Yeah.
In fact I'm not sure we should display the password / secret key / whatever from be-BOP admin UI but it's a deeper thought we'll have later (as usually people knowing credential infos will take them from paypal / sum up / etc and put them in be-BOP, no one really need to take the information from be-BOP as they can fetch it again on payment processor backoffices).

coyotte508:
can we open a separate issue maybe, as it touches sumup/stripe/phoenix/paypal?

Tirodem:
BTW, Paypal goes from 0.01$ to 4000$, 5000$ or more for each transaction, depending of the seller account status.
I think, in order to display dynamically Paypal on a checkout, we should do something like this :
image
It can be done in a round 2 but, in that case, we must implement a hard maximum treshold (like the minimum treshold for Bitcoin that is currently implemented) and set it to the minimum of the "maximum amounts for a transaction" on paypal's side.

Tirodem:
BTW, it's a like detail, but before implementing a SVG, is it possible to change icon from 💸 to 🅿 for payment mean ? (on /admin/order for example)

coyotte508:
> BTW, it's a like detail, but before implementing a SVG, is it possible to change icon from 💸 to 🅿 for payment mean ? (on /admin/order for example)

yes

coyotte508:
> BTW, Paypal goes from 0.01$ to 4000$, 5000$ or more for each transaction, depending of the seller account status. I think, in order to display dynamically Paypal on a checkout, we should do something like this : (...)

> It can be done in a round 2 but, in that case, we must implement a hard maximum treshold (like the minimum treshold for Bitcoin that is currently implemented) and set it to the minimum of the "maximum amounts for a transaction" on paypal's side.

`MIN_SATOSHIS_FOR_BITCOIN_PAYMENT` is only used on the frontend, no checks in the backend :see_no_evil: 

Can we merge as is? I'll do a quick PR afterwards to add the payment methods minimum/maximum amounts in `admin/config` in the "Payment methods" section

Tirodem:
We can merge, I review it, and we'll see later for improvements (but let's not them waste on an iceberg).

Tirodem (approved):

PR #1390, opened 2024-07-22
Reply when trying to add a product with 'Allow partial deposit' AND '…

ithiame:
…Enforce deposit' #1376

### DIFF
--- a/src/lib/server/locks/handle-messages.ts
+++ b/src/lib/server/locks/handle-messages.ts
@@ -371,7 +371,10 @@ const commands: Record<
 				);
 				return;
 			}
-
+			if (product.deposit?.enforce) {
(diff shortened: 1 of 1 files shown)


src/lib/server/locks/handle-messages.ts
```diff
@@ -371,7 +371,10 @@ const commands: Record<
 				);
 				return;
 			}
-
+			if (product.deposit?.enforce) {
+				await send(`Sorry, this product cannot be ordered through Nostr`);
```
coyotte508, line 375:
```suggestion
				await send(`Sorry, this product cannot be ordered through Nostr due to the deposit mechanism`);
```

coyotte508 (approved):

PR #1403, opened 2024-07-23
Improve nostr bot error management

ithiame:


coyotte508 (commented):
Can you:

- merge https://github.com/B2Bitcoin/beBOP/pull/1387 first
- filter out the errors that should already be handled (and if they're not being handled, find out why)

?

coyotte508 (approved):

Issue #1411, opened 2024-07-28
🔎 SEO : 🐛 products & CMS pages have 2 meta descriptions


Tirodem:
Exemple :
https://pvh-editions.com/product/stachtchenko-2
```
<!-- HEAD_svelte-1e2lcw4_START -->
<meta name="viewport" content="width=device-width">
<meta name="description" content="Pensez la SFFF autrement, en physique ou en numérique, avec un esprit libriste et militant.">
<link rel="stylesheet" href="/style/variables.css?v=92">
<link rel="icon" href="/favicon/logo-bookinerie-png-i2pvZc">
<script lang="javascript" src="/script/language/en.js?v=17120a128e1f7381212d56511e7e33adfca20daa-1718010897756"></script><script lang="javascript" src="/script/language/fr.js?v=17120a128e1f7381212d56511e7e33adfca20daa-1718010868180"></script><script defer data-domain="pvh-editions.com" src="https://plausible.dev-bootik.pvh-labs.ch/js/scriptexclusion.js" data-exclude="/admin/*, /admin-*"></script><!-- HEAD_svelte-1e2lcw4_END --><!-- HEAD_svelte-joaawn_START -->
<meta name="description" content="Dans ce petit essai, le spécialiste de Bitcoin Alexandre Stachtchenko examine les dangers de la surveillance financière et son impact sur la vie privée.">
<meta property="og:description" content="Dans ce petit essai, le spécialiste de Bitcoin Alexandre Stachtchenko examine les dangers de la surveillance financière et son impact sur la vie privée.">
<meta property="og:url" content="https://pvh-editions.com/product/stachtchenko-2">
<meta property="og:type" content="og:product">
<meta property="og:title" content="Les dérives de la surveillance financière menacent nos démocraties">
<meta property="og:image" content="https://pvh-editions.com/picture/raw/les-derives-de-la-surveillance-financiere-menacent-nos-democraties-bfpaCC/format/1365">
<meta property="product:price:amount" content="4.73">
<meta property="product:price:currency" content="EUR">
<meta property="og:type" content="og:product">
<!-- HEAD_svelte-joaawn_END -->
```

As you can see, we have 2 meta name="description" :
- `<meta name="description" content="Pensez la SFFF autrement, en physique ou en numérique, avec un esprit libriste et militant.">`, which is the one set on the main be-BOP config
[...]

PR #1418, opened 2024-08-20
SEO : products & CMS pages have 2 meta descriptions #1411

ithiame:


src/routes/+layout.svelte
```diff
@@ -19,7 +19,9 @@
 <svelte:head>
 	<title>{data.websiteTitle}</title>
 	<meta name="viewport" content={data.viewportWidth} />
-	<meta name="description" content={data.websiteShortDescription} />
```
coyotte508, line 22: I think you can use `$page.data.websiteShortDescription` (maybe rename it to something else)

Then, it will default to the layout's `websiteShortDescription`, but a page could also return it in its data, and it will then take priority.

coyotte508 (approved):

Issue #1251, opened 2024-05-07
🦋 Allow to upload and link (for download) digital files on S3 through be-BOP


Tirodem:
Use-case : downloadable files from product or from CMS page (like product notice)
Contrary to digital products files, it should not be protected by token and should be accessibles without completing an order.

PR #1420, opened 2024-08-21
Allow to upload and link (for download) digital files on S3 through b…

ithiame:
…e-BOP #1251


coyotte508:
As discussed, instead we should add a "files" tab similar to pictures.

So that digital files are handled the same way as pictures in the admin (optional productId associated)

src/routes/(app)/admin[[hash=admin_hash]]/file/+page.server.ts
```diff
@@ -0,0 +1,7 @@
+import { collections } from '$lib/server/database';
+
+export const load = async () => {
+	return {
+		pictures: await collections.digitalFiles.find({}).toArray()
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/file/+page.server.ts:5]:
```suggestion
		pictures: await collections.digitalFiles.find({productId: {$exists: false}}).toArray()
```

Similar to pictures, only display the digital files with no product associated.

`productId` should also be made optional on a digital file

src/routes/(app)/admin[[hash=admin_hash]]/file/+page.svelte
```diff
@@ -0,0 +1,18 @@
...
+		<div class="flex flex-col text-center">
+			<a
+				href="{data.adminPrefix}/file/{picture._id}"
+				class="flex flex-col items-center body-hyperlink hover:underline"
+			>
+				<span>{picture.name}</span>
+			</a>
+		</div>
+	{/each}
+</div>
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/file/+page.svelte:18]: picture => file

There should also be a "upload file" button

src/routes/(app)/admin[[hash=admin_hash]]/file/[id]/+page.svelte
```diff
@@ -0,0 +1,22 @@
...
+
+	export let data;
+</script>
+
+<form method="post" action="?/update" use:enhance class="flex flex-col gap-4">
+	{#if data.digitalFile.productId}
+		<a href="{data.adminPrefix}/file" class="underline body-hyperlink text-center">
+			Back to file
+		</a>
+	{/if}
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/file/[id]/+page.svelte:12]: uh it should be a link to the product, and "back to product"

coyotte508 (commented):
@ithiame there is already `/digital-file` routes, you should reuse/modify them instead of adding new `/file` routes.

src/routes/(app)/admin[[hash=admin_hash]]/picture/[id]/+page.svelte
```diff
@@ -28,6 +34,28 @@
 	</label>
 
 	<Picture picture={data.picture} class="object-contain max-h-[500px] max-w-full" />
+	{#each data.picture.storage.formats as format, i}
+		<code
+			>{`${data.websiteLink}/picture/raw/${data.picture?._id}/format/${format.width}`}
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/picture/[id]/+page.svelte:39]: Actually @ithiame we should not need `data.websiteLink` at all in the frontend, it should be removed from the codebase.

We can use `$page.url.origin`, let's at least use it here.

coyotte508 (commented):
I think that there should still be a way for users to download digital files associated to no products?

Maybe we can add a secret key to each digital file uploaded, and give a link with that secret key in the url?

Then the admin can share that link, and users can use it to download the file

src/routes/(app)/admin[[hash=admin_hash]]/picture/[id]/+page.svelte
Comment (coyotte508) [reply src/routes/(app)/admin[[hash=admin_hash]]/picture/[id]/+page.svelte]: ok - see my PR review comment for work remaining:


> I think that there should still be a way for users to download digital files associated to no products?
> 
> Maybe we can add a secret key to each digital file uploaded, and give a link with that secret key in the url?
> 
> Then the admin can share that link, and users can use it to download the file

ithiame:
> I think that there should still be a way for users to download digital files associated to no products?
> 
> Maybe we can add a secret key to each digital file uploaded, and give a link with that secret key in the url?
> 
> Then the admin can share that link, and users can use it to download the file

Isn't this already done here 

image

When we click the donwload link button we can copy the url and share it no ?

coyotte508:
we don't want to change the expiration to 1 week or whatever, the link must work permanently

So we need to create a new URL that will work permanently. But it should be secure.

Which is why I suggested adding a "secret" to the digital files. And we can create a new endpoint with `key=<secret>` in the query to allow downloading digital files forever. That endpoint will redirect to S3's download link if the key matches, otherwise it will error.

src/routes/(app)/digital-file/[key]/+server.ts
Comment (coyotte508) [inline src/routes/(app)/digital-file/[key]/+server.ts:1]: I prefer `/[id]/+server.ts` and to put `key` in the query. It's more consistent (esp if we allow changing the key later)

Also to be consistent with pictures, it should be `digital-file/raw/[id]/+server.ts` (add the `raw)

src/lib/types/DigitalFile.ts
```diff
@@ -2,11 +2,12 @@ import type { Timestamps } from './Timestamps';
...
 	_id: string;
-	productId: string;
+	productId?: string;
 	name: string;
 	storage: {
 		key: string;
 		size: number;
 		url?: string;
 	};
+	secret?: string;
```
coyotte508, line 12: add a sparse unique index on it

src/lib/server/database.ts
```diff
@@ -140,6 +140,7 @@ const indexes: Array<[Collection<any>, IndexSpecification, CreateIndexesOptions?
 	[collections.orders, { 'payments.status': 1 }],
 	[collections.orders, { status: 1, 'payments.status': 1 }],
 	[collections.digitalFiles, { productId: 1 }],
+	[collections.digitalFiles, { secret: 1 }, { sparse: true }],
```
coyotte508, line 143: ... can you make it unique too?

src/routes/(app)/admin[[hash=admin_hash]]/digital-file/finalize/+server.ts
```diff
@@ -38,7 +38,8 @@ export async function POST({ request }) {
 				createdAt: new Date(),
 				updatedAt: new Date(),
 				storage: pendingFile.storage,
-				productId: pendingFile.productId
+				...(pendingFile.productId && { productId: pendingFile.productId }),
+				...(!pendingFile.productId && { secret: crypto.randomUUID() })
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/digital-file/finalize/+server.ts:42]: I think we can always set it for new files

src/routes/(app)/admin[[hash=admin_hash]]/digital-file/[id]/+page.svelte
```diff
@@ -23,3 +34,25 @@
 		<input type="submit" value="Delete" formaction="?/delete" class="btn btn-red" />
 	</div>
 </form>
+{#if data.digitalFile.secret}
+	<h2 class="text-md font-light">This permenantly valid link can be used to share file</h2>
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/digital-file/[id]/+page.svelte:38]:
```suggestion
	<h2 class="text-md font-light">This permanently valid link can be used to share the file to anyone</h2>
```

coyotte508 (approved):

Tirodem:
For obvious reasons I wasn't able to test the non-expiracy link but it seems to work finely.

Issue #1424, opened 2024-09-02
🦩 Add an BO admin flag to disable Nostr-bot intro message


Tirodem:
Nostr-bot is a nice feature but some owners want to use Nostr be-BOP account as a social media account and not just as a Nostr-bot.

So, in /admin/nostr, we want a toggle to enable or not this autoresponse :
![image](https://github.com/user-attachments/assets/d3640176-917a-4e3c-a35e-6b84efbe14ce)
Code location : https://github.com/B2Bitcoin/beBOP/blob/179cebfa730157febb9585f9af6660874a771d64/src/lib/server/locks/handle-messages.ts#L183

PR #1437, opened 2024-09-04
Add an BO admin flag to disable Nostr-bot intro message #1424

ithiame:
🦩 Add an BO admin flag to disable Nostr-bot intro message #1424

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
@@ -86,7 +86,19 @@
 		<li>{relay}</li>
 	{/each}
 </ul>
-
+<h2 class="text-2xl">Intro Message</h2>
+<form action="?/disableIntro" method="post" class="flex flex-col gap-4">
+	<label class="checkbox-label">
+		<input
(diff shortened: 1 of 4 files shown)


src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
```diff
@@ -86,7 +86,19 @@
 		<li>{relay}</li>
 	{/each}
 </ul>
-
+<h2 class="text-2xl">Intro Message</h2>
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte:89]: Add text to specificy this is the message sent when receiving a message that doesn't match a command.

coyotte508 (approved):
@Tirodem note that if we add `!` to all commands, we may want to keep a message when the user typos a command name (eg sends `!hlep` instead of `!help`)

src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte:89]: Yes, nice idea.
If we have !unknowncommand, we can trigger the !help answer.

Issue #1414, opened 2024-07-28
🐛 Configuration done on Merch / Layout for mobile version is not airtight


Tirodem:
![image](https://github.com/user-attachments/assets/a1a45bbb-3740-4b69-bb9f-709656cddf30)
We have some pilot with configuration = "No one" for mobile version, but :
- they have mobile header
- they have widget substitutions for mobile
It's not supposed to happen and they have to keep the desktop version with this setting.

PR #1442, opened 2024-09-09
Configuration done on Merch / Layout for mobile version is not airtig…

ithiame:
…ht #1414

### DIFF
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -20,7 +20,7 @@ export async function load(event) {
 				return `width=${runtimeConfig.viewportContentWidth}`;
 		}
 	})();
-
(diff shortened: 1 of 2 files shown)


src/routes/+layout.server.ts
```diff
@@ -35,6 +35,7 @@ export async function load(event) {
 		websiteShortDescription:
 			runtimeConfig[`translations.${event.locals.language}.config`]?.websiteShortDescription ||
 			runtimeConfig.websiteShortDescription,
-		viewportWidth
+		viewportWidth,
+		notResponsive
```
coyotte508, line 39: move this to `(app)/+layout.server.ts` since it's only needed for UI-rendered pages

coyotte508 (approved):

Issue #1425, opened 2024-09-02
🦩 Detect Zap on nostr-bot npub through be-BOP


Tirodem:
TBD

Tirodem:
From what we see :
- having phoenixd (LN nodeless) doesn't generate /.well-known/lnurlp/ln and doesn't give address when certifying nostr-bot
- nostr doesn't accept BOLT12 LN address for zap
- phoenixd doesn't allow to have non-BOLT12 static address

And, issues to review, for nostr-bots with a node :
- on amethyst (android nostr client), zaps are working for be-bop.io
- on amethyst (android nostr client), zaps are not working for dev-bootik2 :
![image](https://github.com/user-attachments/assets/847651fb-1544-4847-ae4d-4c139faedd58)
- on desktop web client (like primal.net or snort.social), no zap are working of live pilots with node :
![image](https://github.com/user-attachments/assets/a34e9b08-d35e-41ee-a565-c57f585c3436)
(but THERE IS a lightning address, seen onscreen)
![image](https://github.com/user-attachments/assets/a5f759e9-679c-4f38-abdc-4fff0e389137)
![image](https://github.com/user-attachments/assets/94f28252-5758-4a09-b34a-55263ac68653)
I know messages are coming from client, but it doesn't happen with a non-nostr-bot account
Primat.net seems to need a web_ln_extension, as it triggers this on zap on real persons :
![image](https://github.com/user-attachments/assets/fc7f8233-7341-418a-9fb8-2ad60737da15)

Tirodem:
my bad, phoenixd provides a static, non-BOLT12 address : https://phoenix.acinq.co/server/api#get-lightning-address
we have to dig a bit

Tirodem:
I was able to do it on PipeliNostr, I'll share the information on howto later.

```
    // Parse zap if kind 9735
    let zapContext: ZapContext | undefined;
    let parsedZap: ParsedZap | null = null;
    if (event.kind === 9735) {
      parsedZap = parseZapReceipt({
        id: event.id,
        pubkey: event.pubkey,
        kind: event.kind,
        created_at: event.created_at,
        tags: event.tags,
        content: event.rawContent,
      });
      if (parsedZap) {
        zapContext = {
          amount: parsedZap.amount,
          sender: parsedZap.sender.npub,
          sender_pubkey: parsedZap.sender.pubkey,
          recipient: parsedZap.recipient.npub,
[...]

PR #1443, opened 2024-09-10
Detect Zap on nostr-bot npub through be-BOP #1425

ithiame:


Tirodem:
Please create a notification.nostrzap collection, feed it when you detect a zap, and display zap on /admin/nostr
image

coyotte508:
Also make sure the whole protocol is handled, can the nostr client get to the ln url and send a tip?

and then we need to log how much we received in DB

Issue #1449, opened 2024-09-19
✉️ On /admin/config, add option to hide either email or nostr options


Tirodem:
## Summary
Some people are saying "fuck smtp, let's go full Nostr".
As discussed with @coyotte508 , the whole stuff can be a bit complicated (especially on admin & employee side, some being smtp-only for now).
In order to split the need, we'll first display a new option in /admin/config :
![image](https://github.com/user-attachments/assets/23a89867-52ef-481f-8b70-45f9b5dd602f)
If enabled, hide email field on forms and just leave nostr field.

## /identity
Before :
![image](https://github.com/user-attachments/assets/a5057b52-d815-4659-8238-97302e328981)
After when checked :
![image](https://github.com/user-attachments/assets/1daac640-c92d-4216-a50b-86ba35408304)

## /checkout
Before :
![image](https://github.com/user-attachments/assets/d8978dc8-2ff9-4910-bd4c-aad5cfd1de7b)
After when checked :
![image](https://github.com/user-attachments/assets/d72ec717-5416-4f2d-8e4e-b59e80f8ca8a)

## tag widget
![image](https://github.com/user-attachments/assets/787f5377-ba88-485a-9396-631dd6a3fecd)
- [ ]⚠️ if session has a filled email (from before the email disabling, or from SSO/SAML session), don't prefill the "from" field with it when "Prefill with session information" is checked for this form
- [ ]⚠️ on /admin/form (new and existing), refuse email on "Target" field
- [ ]⚠️ on integrated form widget, refuse email on "From" field

## notifications (can be on a 2nd time)
- [ ]⚠️ If I order with an account owning an email (from information filled previously before the option check in /admin/config, or from SSO/SAML) in addition to my filled npub, only send message through nostr, don't try to send email
- [ ]⚠️ For any employee notification through email, if the option is checked, don't trigger the smtp send :
- [ ] for an employee notification, if a npub is filled for this employee on /admin/ARM, use nostr notification instead, otherwise don't send message
- [ ] on /admin/config, option "Copy order emails to contact@pvh-labs.com (set in
 identity section)" should be ignored and notifications should be send through nostr to superadmin npub if filled

PR #1452, opened 2024-09-20
On /admin/config, add option to hide either email or nostr options #1449

ithiame:


src/lib/server/runtime-config.ts
```diff
@@ -97,7 +97,7 @@ const baseConfig = {
 	defaultOnLocation: false,
 	languages: locales,
 	defaultLanguage: 'en' as LanguageKey,
-
+	hideEmailOptions: false,
```
coyotte508, line 100: Let's rename to `contactModes` with default value of `["email", "nostr"]`

coyotte508 (commented):
This is way too many files modified.

A better way is to provide it in `(app)/+layout.server.ts`n and in the components you can check for `$page.data.contactModes` directly (importing `page` from `'$app/stores'`)

There are other ways (like we do for `exchangeRate` but it should be enough for now

src/routes/(app)/admin[[hash=admin_hash]]/config/+page.server.ts
```diff
@@ -87,11 +91,13 @@ export const actions = {
...
 				cartMaxSeparateItems: z.number({ coerce: true }).int().default(0),
-				disableLanguageSelector: z.boolean({ coerce: true })
+				disableLanguageSelector: z.boolean({ coerce: true }),
+				contactModes: z.string().array()
 			})
 			.parse({
 				...Object.fromEntries(formData),
-				paymentMethods: formData.getAll('paymentMethods')
+				paymentMethods: formData.getAll('paymentMethods'),
+				contactModes
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/config/+page.server.ts:100]:
```suggestion
				contactModes: formData.getAll("contactModes")
```

Do it like for paymentMethods just above, to avoid calling `JSON.parse` / avoid doing validation outside of `zod`

Tirodem:
With both methods :
image
With only Nostr and empty field :
image
Please add a front control on input to avoid that.

src/routes/(app)/admin[[hash=admin_hash]]/config/+page.server.ts
Comment (ithiame) [reply src/routes/(app)/admin[[hash=admin_hash]]/config/+page.server.ts:100]: Doesn't work as excepted if i remove `JSON.parse`

Tirodem:
It's OK for nostr only ✅ 
BTW, if I set to email only, nostr is now always displayed. As we hesitated, we decided to allow nostr disabling. So, removing nostr should remove the nostr field.
And while testing it, I saw we're able to do this :
image
Currently, it'll display nostr input. But, for now, it's better to oblige be-BOP owner to have at least either email, or nostr, or both, but not none.
In the future we'll try do to no-contact options for anon be-BOP, but as it's now, we need one or another. And nostr should be masked if the owner wants, like email.

src/routes/(app)/admin[[hash=admin_hash]]/config/+page.server.ts
Comment (coyotte508) [reply src/routes/(app)/admin[[hash=admin_hash]]/config/+page.server.ts:102]: if you change the frontend it should.

In the frontend you can create an `<input name="contactModes">` for each contact mode.

src/lib/components/ContactForm.svelte
```diff
@@ -38,7 +38,14 @@
...
 				placeholder="From"
-				value={contactForm.prefillWithSession ? sessionEmail ?? '' : ''}
+				value={contactForm.prefillWithSession && $page.data.contactModes.includes('email')
+					? sessionEmail ?? ''
+					: ''}
+				pattern={$page.data.contactModes.includes('email')
+					? ''
+					: $page.data.contactModes.includes('nostr')
+					? '^(?!.*@).*'
+					: '^(?!npub).*'}
```
coyotte508, line 48: For emails, instead of a pattern, you can add `type=email` instead of `type=text`

coyotte508 (approved):

coyotte508 (approved):

Issue #537, opened 2023-10-24
🚧 Add a feature flag flag to make the cart preview read only


Tirodem:
Cart preview = 
![image](https://github.com/B2Bitcoin/B2BitcoinBootik/assets/50206014/07497f4d-ee81-4612-9a51-534609cc6f2f)

Currently, product lines can be edited (qty) or deleted (delete button).

For various purposes I'd like a new flag in admin/config :
- [ ] Make cart preview interactive

default : false
- false : the cart preview is read only (qty button are disabled, deletion button is hidden)
- true : the cart preview is interactive (qty button are enabled, deletion button is displayed & enabled)

PR #1453, opened 2024-09-20
Add a feature flag flag to make the cart preview read only #537

ithiame:


src/lib/server/runtime-config.ts
```diff
@@ -85,7 +85,7 @@ const baseConfig = {
 	viewportFor: 'no-one' as 'employee' | 'no-one' | 'visitors' | 'everyone',
 	viewportContentWidth: 1000,
 	socialNetworkIcons: [] as Array<{ name: string; svg: string; href: string }>,
-
+	cartPreviewReadOnly: false,
```
coyotte508, line 88: Rename the parameter, it should be 
```suggestion
	cartPreviewInteractive: false,
```

src/routes/(app)/+layout.svelte
```diff
@@ -325,7 +325,11 @@
 												{#if item.product.type !== 'subscription' && !item.product.standalone}
 													<div class="flex items-center gap-2">
 														<span class="text-xs">{t('cart.quantity')}: </span>
-														<CartQuantity {item} sm />
+														{#if data.cartPreviewReadOnly}
+															<CartQuantity {item} sm />
+														{:else}
+															<CartQuantity {item} sm disabled />
+														{/if}
```
coyotte508, line 332:
```suggestion
														<CartQuantity {item} sm disabled={!cartPreviewInteractive} />
```

coyotte508 (approved):

Tirodem:
I've had some feedback but after reading the initial ticket again it's what I asked for, so... Let's go 👌

PR #1459, opened 2024-09-24
Add magnifying glass option ('loupe') on product visual #871

ithiame:


coyotte508:
you should be able to do it in pure CSS.

For example:

```
zoom ? 'scale-150' : ''
```

This can be replaced by `hover:scale-150`

If the result is not satisfactory, and you only want to zoom when hovering over the parent, you can add the `group` class to the parent and `group-hover:scale-150` to the child: https://hf.co/chat/r/1OC1ugt?leafId=688d5d12-917b-4aab-9954-854f2c60afac

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -174,10 +174,12 @@
 				<!-- add product name -->
 				<h2 class="text-4xl body-title">{data.product.name}</h2>
 				<!-- Getting this right with rounded borders on both chrome & FF is painful, chrome NEEDs overflow-hidden -->
-				<div class="aspect-video w-full overflow-hidden px-4">
+				<!-- svelte-ignore a11y-no-static-element-interactions -->
+				<!-- svelte-ignore a11y-mouse-events-have-key-events -->
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:178]:
```suggestion
```

coyotte508 (approved):

Tirodem:
I need to test it again once the `was loaded over HTTPS, but requested an insecure image` issue on dev sandbox are fixed.
I see the philosophy, it seems good, some things are bothering me (I'll take videos once it's fixed with true pictures), but it goes in a good way.

Tirodem:
BTW, from what I see, on mobile display, we need an unzoom action (right now, by clicking, it zooms, but it stays zoomed)

Issue #333, opened 2023-07-15
Implement Google Shopping feature


Tirodem:
prerequisite: #8 

https://support.google.com/merchants/answer/7439058?hl=fr

https://www.lengow.com/fr/aller-plus-loin/comment-creer-un-flux-google-shopping/

export can be enabled / disabled globally through a toggle in /admin/config
but, on product level, it can be enabled / disabled for export with canal selector (products with "google shopping" disabled should not be appearing in google export)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/ee535363-ea4c-4e64-9a12-557ff544441a)

PR #1460, opened 2024-09-24
Implement Google Shopping feature #333

ithiame:


src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -107,6 +108,24 @@ export const load = async ({ params, locals }) => {
 			sort: { percentage: -1 }
 		}
 	);
+	const jsonLd = {
+		'@context': 'https://schema.org/',
+		'@type': 'Product',
+		name: product.name,
+		image: `${ORIGIN}/picture/raw/${pictures[0]._id}/format/${pictures[0].storage.formats[0].width}`,
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:115]: Is there a recommended picture size?

`.formats[0]` is always the highest resolution image, maybe we can give something more appropriate

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -107,6 +108,24 @@ export const load = async ({ params, locals }) => {
...
+		description: product.description,
+
+		brand: {
+			'@type': 'Brand',
+			name: product.name
+		},
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:121]: No, we should only specfiy `brand` if we have the brand of the product (and it's not information we store at the moment)

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -107,6 +108,24 @@ export const load = async ({ params, locals }) => {
...
+		offers: {
+			'@type': 'AggregateOffer',
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:123]:
```suggestion
			'@type': 'Offer',
```

`AggregateOffer` is only used when there are multiple offers, and then we need to specify a `highPrice` and `lowPrice`

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -118,7 +137,8 @@ export const load = async ({ params, locals }) => {
 			productCMSAfter: cmsFromContent({ content: product.contentAfter }, locals)
 		}),
 		showCheckoutButton: runtimeConfig.checkoutButtonOnProductPage,
-		websiteShortDescription: product.shortDescription
+		websiteShortDescription: product.shortDescription,
+		jsonLd
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:141]: Instead of passing it to the frontend like this, is there a reason for not defining it directly in the frontend?

Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.server.ts:141]: There is not. I just put it to make code readable

Comment (coyotte508) [reply src/routes/(app)/product/[id]/+page.server.ts:141]: If it's readability we could maybe create the function to convert product to JSON-LD in `src/lib/types/Product.ts` 

But let's avoid doing the frontend's work on the backend.

Up to you wheter to put it directly in `+page.svelte` or in `src/lib/types/Product.ts`

src/routes/(app)/product/[id]/+page.server.ts
Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.server.ts]: **Minimum requirements:**

Image size: at least 100 x 100 pixels
Aspect ratio: 1:1 (square)
File format: JPEG, GIF, PNG, or BMP
File size: less than 2MB

**Recommended image sizes:**

For product images:
500 x 500 pixels (or larger)
Aspect ratio: 1:1 (square) or 4:3 (landscape)
For product variants (e.g., different colors or sizes):
1000 x 1000 pixels (or larger)
Aspect ratio: 1:1 (square) or 4:3 (landscape)

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -103,6 +104,27 @@
 		return true;
 	}
 	const { t, locale } = useI18n();
+	onMount(() => {
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:107]: Why is this in `onMount`?

Google shopping needs the info to be avaialble at first render of the page

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -103,6 +104,25 @@
...
+			priceCurrency: data.product.price.currency
+		}
+	};
+
+	if (data.product.actionSettings.googleShopping.visible) {
+		const script = document.createElement('script');
+		script.type = 'application/ld+json';
+		script.textContent = JSON.stringify(jsonLd);
+		document.head.appendChild(script);
+	}
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:124]: I don't think you can call this server-side

Can you not create the `script` tag directly in the page content? (or in `svelte:head`)

See also this article: https://navillus.dev/blog/json-ld-in-sveltekit

Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.svelte:124]: If i create new script i will not be able to access `data`.

Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.svelte:124]: Maybe i can use microdonnée in page content

 ```
<div itemtype="https://schema.org/Product" itemscope>
      <meta itemprop="mpn" content="925872" />
      <meta itemprop="name" content="Executive Anvil" />
      <link itemprop="image" href="https://example.com/photos/16x9/photo.jpg" />
      <link itemprop="image" href="https://example.com/photos/4x3/photo.jpg" />
      <link itemprop="image" href="https://example.com/photos/1x1/photo.jpg" />
```

Comment (coyotte508) [reply src/routes/(app)/product/[id]/+page.svelte:124]: look at the article, at the end

Comment (coyotte508) [reply src/routes/(app)/product/[id]/+page.svelte]: to clarify:

- can you script to add the script tag in `svelte:head` ?
- **if** it doesn't work, you can use `@html` in `svelte:head` (like in the article) to render the LD+JSON script tag

File: src/routes/(app)/product/[id]/+page.svelte [hunk 7]
```diff
@@ -123,6 +137,10 @@
 	<meta property="product:price:amount" content={String(data.product.price.amount)} />
 	<meta property="product:price:currency" content={data.product.price.currency} />
 	<meta property="og:type" content="og:product" />
+	{#if data.product.actionSettings.googleShopping.visible}
+		<!-- eslint-disable svelte/no-at-html-tags -->
+		{@html serializeSchema(schema)}
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:142]:
```suggestion
		<script type="application/ld+json">
			{
				JSON.stringify(schema);
			}
		</script>
```

Did you try this? I suggested to try `@html` only if it didn't work (I even bolded the **if** :) )

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -103,6 +105,18 @@
 		return true;
 	}
 	const { t, locale } = useI18n();
+	const schema: WithContext<SchemaOrgProduct> = {
+		'@context': `https://schema.org`,
+		'@type': 'Product',
+		name: data.product.name,
+		image: `${$page.url.origin}/picture/raw/${data.pictures[0]._id}/format/${data.pictures[0].storage.formats[0].width}`,
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:112]: can you get the first image with >500px width & height? (or the first image if not available)

src/routes/(app)/product/[id]/+page.svelte
Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.svelte:142]: as i already said, if i have two <script> tag i will not be able to access the variable in the first tag. I tried it but it doesn't work

Comment (coyotte508) [reply src/routes/(app)/product/[id]/+page.svelte:142]: ok :)

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -49,6 +51,10 @@
 	$: currentPicture =
 		data.pictures.find((picture) => picture._id === $page.url.searchParams.get('picture')) ??
 		data.pictures[0];
+	const googlePicture =
+		data.pictures.find((picture) => {
+			return picture.storage.formats.find((image) => image.width > 500 && image.height > 500);
+		}) ?? data.pictures[0];
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:57]: ok no actually `currentPicture` should be used always

BUT the correct format should be displayed - the smallest one that has both >= 500 width & height

Tirodem:
Ah, damn biatches.
I wanted to test, then :
- google merchant account seems required
- image
- image
I cannot go further for now on sandbox, unless I send an email to anything@dev-bootik2.pvh-labs.ch.
WIP...

Tirodem:
(btw even if I manage to do it with email, it might be interesting to allow the addition of the said html meta name "google-site-verification" through admin, for smtp-less be-BOP)

Tirodem:
Mail issue resolved, but I'm stuck here :
image
I expected to find a file on be-BOP to link its url here : 
image
but I don't find the information about it in the code
did I miss something ?

Tirodem:
We need around 24 hours before testing again on dev-bootik2
dev-bootik2 is now connected to Google Search Console, we'll see if Google is able to crawl the website and propose a "assisted inscription", in that case, no file will be needed, but it'll depend of the Google Search Console x Google Merchant Center integration.

Tirodem:
🧊 
From what I see we really need a true google merchant center integration with export file, structured datas aren't sufficient for google advance search console to create products.

Tirodem:
Google Search Console is now able to check products, but Google Merchant Center refuse to import them as long as it's not through individual form, export file or direct API connexion.
I'll do the file generation ticket later.
image

Tirodem:
@coyotte508 as seen it's not enough to push products to Google Merchant Center, still, Google Search Console is able to identify products, so we can already merge what is done as it's already improving search engine recognition.
Can you do one final review before merging ? I'll check for create another ticket with the google merchant file generation to export later.
Thanks.

coyotte508 (approved):

PR #1462, opened 2024-09-25
✨ Fix receipts with discounts

coyotte508:
- Reorder to show shipping price before total
- Add "discount" section
- Show correct VAT/HT prices by applying discount proportionally

Much simpler PR than I expected. Not sure if it handles shipping prices correctly (we can open an issue if not) but it should be a nice improvement and fix a lot of things. It works with Multi-VAT orders:

image

## Todo:

In another PR, also fix discount + shipping price if it's an issue

### DIFF
--- a/src/routes/(app)/order/[id]/payment/[paymentId]/receipt/+page.svelte
+++ b/src/routes/(app)/order/[id]/payment/[paymentId]/receipt/+page.svelte
@@ -71,8 +71,6 @@
 		currency
 	};
 
-	const vatPercentage = totalVat.amount ? totalVat.amount / totalNoTax.amount : 0;
-
 	const discountNoTax = {
 		currency,
 		// Alternative, but not as precise with decimal point:
--- a/src/routes/(app)/order/[id]/payment/[paymentId]/receipt/+page.svelte
+++ b/src/routes/(app)/order/[id]/payment/[paymentId]/receipt/+page.svelte
@@ -6,6 +6,7 @@
 	import { invoiceNumberVariables } from '$lib/types/Order.js';
 	import { fixCurrencyRounding } from '$lib/utils/fixCurrencyRounding';
 	import { sum } from '$lib/utils/sum.js';
+	import { sumCurrency } from '$lib/utils/sumCurrency.js';
 	import { marked } from 'marked';
 
 	export let data;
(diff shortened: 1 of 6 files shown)


src/routes/(app)/order/[id]/payment/[paymentId]/receipt/+page.svelte
```diff
@@ -4,7 +4,10 @@
 	import Trans from '$lib/components/Trans.svelte';
 	import { useI18n } from '$lib/i18n.js';
 	import { invoiceNumberVariables } from '$lib/types/Order.js';
+	import { fixCurrencyRounding } from '$lib/utils/fixCurrencyRounding';
 	import { sum } from '$lib/utils/sum.js';
+	import { sumCurrency } from '$lib/utils/sumCurrency';
+	import { toCurrency } from '$lib/utils/toCurrency';
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/payment/[paymentId]/receipt/+page.svelte:10]:
```suggestion
```

Tirodem:
I've tried with mono-VAT cart :
image
The discount I made was 500€ and my cart went to 5000€ TTC to 4500€ TTC, so it was 500€ discount on TTC price.
Here's what I have in the proforma :
image
Discount is display before the cart HT amount and VAT application.
Thing is I'm nearly sure that laws oblige us to display "Discount : -€462.53" (aka full HT price minus discounted HT price) instead of "Discount : -€500" (the discount on TTC price filled on pos /checkout)

coyotte508:
We can change "Rabais" to "Rabais TTC" wording if you want ?

Tirodem:
I'm clearly not sure it's enough to do that in term of law & accountability compliance.
Is it bothersome because of the source data ?
Can't we display cart total ttc minus cart total ht instead of discount amount value ?

coyotte508:
> I'm clearly not sure it's enough to do that in term of law & accountability compliance.

To be honest I think the current PR is enough: 

- There is HT
- There is TTC
- There is amount of VAT paid

From an accounting point of view, what's annoying is we don't display the updated amount of VAT (or HT) for each VAT rate, the accountant will have to recompute it

No problem to display the individual amount of VAT for each rate instead of just "TVA totale", I can display "TVA 5%" and "TVA 20%" if you want?

> Can't we display cart total ttc minus cart total ht instead of discount amount value ?

Not sure what that means, cart total ttc minus cart total ht is VAT, and we already display it

Tirodem:
> Not sure what that means, cart total ttc minus cart total ht is VAT, and we already display it
Yeah, I was to quick on that. Ignore that point.

BTW, our accounting referents confirmed to me : discount should be display before TVA application, either as a discount amount on HT total price, or as a discount %.
It might seem dumb but it seems that it's the expected behavior.

coyotte508:
image

Tirodem:
Is good, thanks for the change 👌 
If you can fix the lint error, I'll merge it immediatly after & deploy it.

Tirodem (approved):

PR #1463, opened 2024-09-30
Offer full order checkbox

ithiame:


src/routes/(app)/checkout/+page.svelte
```diff
@@ -33,8 +33,9 @@
 	let onLocation = data.defaultOnLocation;
 	$: offerDeliveryFees = onLocation;
 	let addDiscount = false;
-	let discountAmount = 0;
-	let discountType: DiscountType;
+	let offerOrder = false;
+	$: discountAmount = offerOrder ? 100 : 0;
+	$: discountType = (offerOrder ? 'percentage' : 'fiat') as DiscountType;
```
coyotte508, line 38: I think it may have weird behavior when interacting with the controls. You can do something like this:

```suggestion
	$: if (offerOrder) {
		discountType = 'percentage';
		discountAmount = 100;
	}
	$: if (discountType !== 'percentage' || discountAmount !== 100) {
		offerOrder = false;;
	}
```

src/routes/(app)/checkout/+page.svelte
```diff
@@ -33,8 +33,14 @@
 	let onLocation = data.defaultOnLocation;
 	$: offerDeliveryFees = onLocation;
 	let addDiscount = false;
+	let offerOrder = false;
 	let discountAmount = 0;
 	let discountType: DiscountType;
+	$: if (offerOrder) {
+		discountType = 'percentage';
+		discountAmount = 100;
+	}
```
coyotte508, line 42: The problem is if the user changes the discount amount, the `offerOrder` checkbox stays ticked, that's why I added the second `if` in the comment

src/routes/(app)/checkout/+page.svelte
```diff
@@ -959,13 +965,23 @@
 							required
 						>
 							<option value="fiat">{data.currencies.main}</option>
-							<option value="percentage">%</option>
+							<option value="percentage" selected={offerOrder}>%</option>
```
coyotte508, line 968: not needed if we control `discountType` directly

src/routes/(app)/checkout/+page.svelte
ithiame: The second if was giving this error `Cyclical dependency detected: discountType → offerOrder → discountType`

coyotte508: you can try reordering the elements

or find another solution

ithiame: When offer order is checked you can't update discountAmount and discountType

coyotte508 (approved):

Issue #1335, opened 2024-06-12
🏪 Harmonize page display on /pos/touch


Tirodem:
🏪From https://github.com/B2Bitcoin/beBOP/issues/1291 & https://github.com/B2Bitcoin/beBOP/issues/1298

Display should always be the sames : CTA on same position, no vertical variations, only product list changes.

What we have now :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/1379b91d-a019-4042-b527-7b2d59a0b271)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/4d91dd92-0186-487f-b654-123ecf331257)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/346cde79-1bfd-42df-aa63-6a4e02a54535)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/a0e0237c-ae96-4a95-8f17-810ffccffb3f)
On "Tous les articles", display is not good.
On pagination, bottom CTAs height should be the same, whatever the number of products displayed on the list.
We can display more products on each page, have a longer ticket, and bottom CTA lower.

What do I expect :
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/5e9e6274-b727-422e-89c3-22205aa4d83a)
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/98e4adc8-7e89-4812-a4b5-7773f5a7098c)

PR #1464, opened 2024-09-30
Harmonize page display on /pos/touch #1335

ithiame:


Tirodem:
When page does have more than one product page :
image
it's ok

But if there's the page selector :
image
there's indesirable margin

After analysis, it's not related to having X page or not, but to a long product name :
image
image

=> We must introduce a chain breaker to avoid bad display like this

We can see if we can fix https://github.com/B2Bitcoin/beBOP/issues/1323 at the same time.
Beside that, the fix seems good.

Also, what I forgot (I'm creating a new ticket for later) : https://github.com/B2Bitcoin/beBOP/issues/1476
And : https://github.com/B2Bitcoin/beBOP/issues/1477

Tirodem:
Code already onboarded through other PR. It's OK but avoid cascading PR, it's confusing and might be dangerous @ithiame .

PR #1468, opened 2024-10-01
Fix secondary currency pos

ithiame:


coyotte508 (approved):

Tirodem:
image
image
Why the 5 sats difference @ithiame ?

ithiame:
I don't know
I will check

Tirodem:
image
image
image
Pending cart is not good (but i'll open another ticket if total is good)
Pending order is good :
image
image
But it seems, only, check here :
image
"Reste à payer" display the fiat sum (1st currency) in bitcoin (2nd currency) 😱

Tirodem:
As said earlier, this PR is not finish, as intermediate amount are still wrong, and we need to align every thruth source from /checkout and /order pages.

Tirodem:
Variations before validation /checkout are normal and acceptables.
While order is creating and pending, it's OK, except this particular case :
image
https://dev-bootik2.pvh-labs.ch/order/3bb25afa-f5e4-40b4-846d-295475507091
It'll create a non-urgent ticket for this one.

Issue #1466, opened 2024-10-01
🦋 Add selection item to product (with no specific price nor specific stock)


Tirodem:
Some pilots want to sell products with variations.
Products with variation can be long & expensive to develop & test.
We did some kind of bypass with dev-bootik1, but it's not scalable when products have 2 variations (like color & size for a tee).

To bypass that again, **as long as the price is the same for every variation**, and **as long as the seller doesn't need stock management per variation**, we can introduce the notion of "customisable product with variation".

It'll look something like that on admin side when editing or creating a product :
![image](https://github.com/user-attachments/assets/d6556241-ffb3-43fb-8899-5d40df9ee8f5)
![image](https://github.com/user-attachments/assets/f2524f9b-7858-4564-91df-276d378a65d8)
(clicking on "Add variation" add a new line)

Product page size, each distinct value of "Variation name" will be treated as a selection menu.
When there's only one occurrence, the selection menu appears but is disabled.
Default value is the first (as filled).
![image](https://github.com/user-attachments/assets/26e2d4cb-5ecc-4cdb-b6cf-d44c297844b6)

On cart, checkout, order & invoice :
- display "Product name - Variation value 1 - Variation value 2 - ... - Variation value n"
![image](https://github.com/user-attachments/assets/d7a90e52-5501-4e62-986c-9b83d2db3664)

Tirodem:
Note that it'll only edit the product label, but it's the same price, it's the same stock. It's some kind of "metadata" for small shops, as long as we don't have a true variation engine manager.

Tirodem:
As alerted this morning : for funcitonal reason, it's better to link variations to a standalone product (that'll allow to as 2 green XL tee, 1 red XXL tee and 3 black XS tee, one by one).
So, in admin, if standalone is disabled, disable variation & variations application too.

PR #1469, opened 2024-10-02
Add selection item to product (with no specific price nor specific st…

ithiame:
…ock) #1466
🦋 Add selection item to product (with no specific price nor specific stock) #1466

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -71,7 +71,7 @@
 			hideContentAfter: false
 		},
 
-		hasLightVariations: false
+		hasVariations: false
 	};
 
 	let paymentMethods = product.paymentMethods || [...availablePaymentMethods];
@@ -89,7 +89,7 @@
 		percentage: 50,
 		enforce: false
 	};
-	$: variationLines = product.variations?.length || 2;
+	$: variationLines = product.variationLabels?.length || 2;
 	if (product._id && isNew) {
 		product.name = product.name + ' (duplicate)';
 		product._id = generateId(product.name, false);
@@ -445,30 +445,30 @@
 			<input
 				class="form-checkbox"
 				type="checkbox"
-				bind:checked={product.hasLightVariations}
-				name="hasLightVariations"
+				bind:checked={product.hasVariations}
+				name="hasVariations"
 				disabled={!product.standalone}
 			/>
 			Product has light variations (no stock nor price difference)
 		</label>
-		{#if product.hasLightVariations}
-			{#each [...(product.variations || []), ...Array(variationLines).fill( { name: '', value: '' } )].slice(0, variationLines) as variation, i}
+		{#if product.hasVariations}
+			{#each [...(product.variationLabels || []), ...Array(variationLines).fill( { name: '', value: '' } )].slice(0, variationLines) as variationLabel, i}
 				<div class="flex gap-4">
 					<label class="form-label">
 						Name
 						<input
 							type="text"
-							name="variations[{i}].name"
+							name="variationLabels[{i}].name"
 							class="form-input"
-							value={variation.name}
+							value={variationLabel.name}
 						/>
 					</label>
 					<label class="form-label">
 						Value <input
 							type="text"
-							name="variations[{i}].value"
+							name="variationLabels[{i}].value"
 							class="form-input"
-							value={variation.value}
+							value={variationLabel.value}
 						/>
 					</label>
 				</div>
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -27,6 +27,7 @@
 	import type { PojoObject } from '$lib/server/pojo';
 	import type { PaymentMethod } from '$lib/server/payment-methods';
 	import { useI18n } from '$lib/i18n';
+	import { toLower } from 'lodash-es';
 
 	const { t } = useI18n();
 
@@ -89,7 +90,7 @@
 		percentage: 50,
 		enforce: false
 	};
-	$: variationLines = product.variations?.length || 2;
+	$: variationLines = product.variationLabels?.names ? 0 : 2;
 	if (product._id && isNew) {
 		product.name = product.name + ' (duplicate)';
 		product._id = generateId(product.name, false);
@@ -192,6 +193,9 @@
 			event.preventDefault();
 		}
 	}
+	let variationLabelsNames: string[] = [];
+
+	let variationLabelsValues: string[] = [];
 </script>
 
 <form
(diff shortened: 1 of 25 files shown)


Tirodem:
Is it possible to change the menu display ?
Currently :
image
Wished :
image
or
image

Tirodem:
Alors, add to cart popin :
image
and cart preview :
image
Should have the name with variation.

Tirodem:
Name with variation should include a space around the "-"
Here :
image
Wished :
image

Tirodem:
On order, item name should be changed :
image
in order to appear on the invoice :
image

Tirodem:
One last thing : I made a product standalone, then I gave it variations, then I removed the standalone property.
Here's the result :
image
image
I'm not against keeping the filled properties, but :
- as long as standalone checkbox is unchecked, variation checkbox should be unchecked and stay unchecked
- as long as standalone checkbox is unchecked, variations should not be displayed on product page

coyotte508:
will review after @Tirodem "ok" given all the comments :)

Tirodem:
It's ok functionally, @coyotte508 it's open for review 👌

src/lib/types/Product.ts
```diff
@@ -18,6 +18,10 @@ export interface ProductTranslatableFields {
 		href: string;
 		fallback?: boolean;
 	}[];
+	variations?: {
+		name: string;
+		value: string;
+	}[];
```
coyotte508, line 24: This can't be stored in `ProductTranslatableFields` in my opinion. This determines how the backend works (at checkout, ...) and also whether or not the product needs to be standalone.

This structure is more for frontend fields.

Maybe something like this instead:

```suggestion
	variationLabels?: {
		names: Record<string,string>; // name by category id
		values: Record<string, Record<string, string>>; // value by value id inside category id  
	}[];
```

With a `variations` field instead inside `Product.ts` - that will contain ids - and later prices. The ids can be internal/hidden (eg set initially when creating the variation with `generateId` function, and then reused smartly by the frontend)

src/lib/types/Product.ts
```diff
@@ -73,6 +77,7 @@ export interface Product extends Timestamps, ProductTranslatableFields {
 		hideContentBefore: boolean;
 		hideContentAfter: boolean;
 	};
+	hasLightVariations?: boolean;
```
coyotte508, line 80: Does not need to be stored in my opinion - unless we want to add a way to disable variations while keeping them in DB (but this isn't the behavior at the moment, unticking the checkbox and saving the product will erase the variations)

We can rely on `variations.length` instead

coyotte508 (commented):
Good job!

I'm sorry about the data model change request :sweat:

src/lib/types/Product.ts
Tirodem: For now we don't need that but I think it can be useful in the future (that's a feeling, no precise plan yet).

coyotte508: Rename to more generic `hasVariations` then

Tirodem: Yeah, in the end, what won't be covered by "light" variations will be probably called "Product component" or "Product addon" or something like that. So let's go for hasVariations.

src/lib/types/Product.ts
ithiame: Can you explain the content of each Record ? An example ?

coyotte508: For example, two variations: "Color" "Blue" and "Color" "Red"

It's stored as : 

```ts
  variations: [{name: "color", value: "blue"}, {name: "color", value: "red"}],
  variationLabels: {values: {color: {red: "Red", blue: "Blue"}}, names: {color: "Color"}},
  translations: {
    fr: { variationLabels: {values: {color: {red: "Rouge", blue: "Bleu"}}, names: {color: "Couleur"}}}
  }
```

src/lib/types/Product.ts
```diff
@@ -73,6 +77,7 @@ export interface Product extends Timestamps, ProductTranslatableFields {
 		hideContentBefore: boolean;
 		hideContentAfter: boolean;
 	};
+	hasVariations?: boolean;
```
coyotte508, line 80: There should be a `variations` field - same a ebfore, but just containing ids

src/lib/types/Product.ts
```diff
@@ -18,6 +18,10 @@ export interface ProductTranslatableFields {
 		href: string;
 		fallback?: boolean;
 	}[];
+	variationLabels?: {
+		names: Record<string, string>;
+		values: Record<string, Record<string, string>>;
+	}[];
```
coyotte508, line 24:
```suggestion
	};
```

src/lib/types/Cart.ts
```diff
@@ -26,6 +26,7 @@ export interface Cart extends Timestamps {
 		reservedUntil?: Date;
 		depositPercentage?: number;
 		internalNote?: { value: string; updatedAt: Date; updatedById?: User['_id'] };
+		customProductName?: string;
```
coyotte508, line 29: Rather than storing just the name, the chosen variations should be stored (only their ids)

eg `chosenVariations?: {name: string, value: string}[]` or `chosenVariations: Record<string, string>`

src/lib/types/Order.ts
```diff
@@ -147,6 +147,7 @@ export interface Order extends Timestamps {
 		product: Product;
 		quantity: number;
 		customPrice?: { amount: number; currency: Currency };
+		customProductName?: string;
```
coyotte508, line 150: Same, variation ids should be stored.

When creating an order, if the variations ids from the cart are no longer existing, it should throw an error or remove items from the cart

src/lib/components/ProductForm.svelte
```diff
@@ -439,6 +441,42 @@
...
+							type="text"
+							name="variations[{i}].name"
+							class="form-input"
+							value={variation.name}
+						/>
+					</label>
+					<label class="form-label">
+						Value <input
+							type="text"
+							name="variations[{i}].value"
```
coyotte508, line 469: Those inputs should be on `variationLabels`.

There should be also `variations` but as hidden inputs - which are "slugified" for each name/value

src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/+page.server.ts
```diff
@@ -177,7 +176,15 @@ export const actions: Actions = {
 					...(parsed.vatProfileId && { vatProfileId: new ObjectId(parsed.vatProfileId) }),
 					...(parsed.restrictPaymentMethods && {
 						paymentMethods: parsed.paymentMethods ?? []
-					})
+					}),
+					...(parsed.standalone && { hasVariations: parsed.hasVariations }),
+					...(parsed.hasVariations && {
+						variations: parsed.variations?.filter((variation) => variation.name && variation.value)
+					}),
+					...(parsed.hasVariations &&
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/+page.server.ts:184]:
```suggestion
					...(parsed.standalone && parsed.hasVariations &&
```

src/routes/(app)/admin[[hash=admin_hash]]/product/new/+page.server.ts
```diff
@@ -176,6 +176,16 @@ export const actions: Actions = {
 						},
 						tagIds: parsed.tagIds,
 						cta: parsed.cta?.filter((ctaLink) => ctaLink.label && ctaLink.href),
+						...(parsed.standalone && { hasVariations: parsed.hasVariations }),
+						...(parsed.hasVariations && {
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/new/+page.server.ts:180]:
```suggestion
						...(parsed.standalone && parsed.hasVariations && {
```

src/routes/(app)/admin[[hash=admin_hash]]/product/new/+page.server.ts
```diff
@@ -176,6 +176,16 @@ export const actions: Actions = {
...
+							variations: parsed.variations?.filter(
+								(variation) => variation.name && variation.value
+							)
+						}),
+						...(parsed.hasVariations &&
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/new/+page.server.ts:185]:
```suggestion
						...(parsed.standalone && parsed.hasVariations &&
```

coyotte508 (approved):

Tirodem:
Whole feedbacks here, there are many ones : https://discord.com/channels/1022592041709998080/1088417396232028170/1292991794900111440

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -145,13 +153,17 @@ async function addToCart({ params, request, locals }: RequestEvent) {
...
 		})
 		.parse({
 			quantity: formData.get('quantity') || undefined,
 			customPriceAmount: formData.get('customPriceAmount') || undefined,
 			customPriceCurrency: formData.get('customPriceCurrency') || undefined,
-			deposit: formData.get('deposit') || undefined
+			deposit: formData.get('deposit') || undefined,
+			chosenVariations: formData.get('chosenVariations')
+				? JSON.parse(formData.get('chosenVariations')?.toString() || '[]')
+				: undefined
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:166]: Use this if you must:

```ts
		const json: JsonObject = {};
		for (const [key, value] of await request.formData()) {
			if (value) {
				set(json, key, value);
			}
		}
```

but no `JSON.parse` here please, it can cause 500 errors

src/lib/components/ProductForm.svelte
```diff
@@ -27,6 +27,7 @@
 	import type { PojoObject } from '$lib/server/pojo';
 	import type { PaymentMethod } from '$lib/server/payment-methods';
 	import { useI18n } from '$lib/i18n';
+	import { toLower } from 'lodash-es';
```
coyotte508, line 30:
```suggestion
	import { toLower } from 'lodash-es';
```

Let's not import anything from lodash on the frontend

you can use ?.toLowerCase() on a string

coyotte508 (approved):

Tirodem:
Some feedback but will be done on a new ticket as improvement

PR #1479, opened 2024-10-03
Harmonize pos cart content

ithiame:


coyotte508 (commented):
frontend-only, POS-only, will let @Tirodem review

Tirodem (approved):

Issue #1472, opened 2024-10-02
🦋 Add price delta on product variation


Tirodem:
Following https://github.com/B2Bitcoin/beBOP/issues/1466
Add a new input to fill a price delta on variation :
![image](https://github.com/user-attachments/assets/37debfdd-b8c8-4c66-badc-ad775b3e209d)
Currency applied is the same than here :
![image](https://github.com/user-attachments/assets/0dff2c98-786d-4598-8683-87786d8114d5)

On product page, price will be adapted depending of the variations, and will be product price + variation1 price + variation2 price + ... + variationN price.

Variation price delta, as shown in screenshot, can be either positive or negative.
Variation price delta are tax excluded.

coyotte508:
why not

PR #1491, opened 2024-10-09
Add price delta on product variation #1472

ithiame:


Tirodem:
# 1
image
Replace
`(no stock nor price difference)`
with
`(no stock difference)`

 # 2
If price variation is empty, make it 0 (just to be clearer in the product edit / creation UI)

# 3
image
I'd like "Variation de prix" to disappear, and prive variation to apply directly on the main currency price, and so second currency price, as it's usually done on eshop websites.
Is it complicated ?
I want a change estimation before you going for code change, to see if it's worth. Otherwise, I'll find some quickwin / wording backdoor solution.

Tirodem:
Negative price variations are OK on product page but not elsewhere.
 
Example with this configuration :
image
image
image
Let's empty the cart and add the same variation check order directly :
image

It's good.
But let's now introdude a negative prive variation combination :
image
image
image
image

Tirodem:
tl;dr from my opinion : the minimum amount is applied to cart and checkout.
We can work like this, especially since overkill negative price variations can lead to this :
image
But, in we do that, we have to limit price variation to 0 or positive amount only.

Tirodem:
Tests OK, @coyotte508 ready to PR

src/lib/types/Product.ts
```diff
@@ -78,10 +78,7 @@ export interface Product extends Timestamps, ProductTranslatableFields {
 		hideContentAfter: boolean;
 	};
 	hasVariations?: boolean;
-	variations?: {
-		name: string;
-		value: string;
-	}[];
+	variationPrices?: Record<string, Record<string, number>>;
```
coyotte508, line 81: please switch it back, just add `price` optional in addition to `name` and `value`

Later we may want to add more things than just price, and we'll be able to add more optional props to `variations`

ithiame: Was too complicate to use ` variations?: {name: string; value: string; }[];` 
but let's try again

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -177,7 +177,9 @@ async function addToCart({ params, request, locals }: RequestEvent) {
 
 	await addToCartInDb(product, quantity, {
 		user: userIdentifier(locals),
-		...(product.payWhatYouWant && { customPrice }),
+		...((product.payWhatYouWant || (product.hasVariations && product.standalone)) && {
+			customPrice
+		}),
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:182]: When variations are enabled, the user can decide to pay 0 € for the product

Instead, the price should be calculated from `chosenVariations`

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -166,15 +166,30 @@ async function addToCart({ params, request, locals }: RequestEvent) {
...
 
 	const customPrice =
 		customPriceAmount && customPriceCurrency
 			? {
 					amount: parsePriceAmount(customPriceAmount, customPriceCurrency),
 					currency: customPriceCurrency
 			  }
+			: variationPrice > 0
+			? {
+					amount: parsePriceAmount(variationPrice.toString(), runtimeConfig.mainCurrency),
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:189]: `parsePriceAmount` is only to parse data passed from the frontend. It's not needed here

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -166,15 +166,30 @@ async function addToCart({ params, request, locals }: RequestEvent) {
...
+		product.hasVariations && chosenVariations
+			? Object.entries(chosenVariations).reduce((total, variation) => {
+					return (
+						total +
+						(product.variations?.find(
+							(vari) => vari.name === variation[0] && variation[1] === vari.value
+						)?.price || 0)
+					);
+			  }, 0)
+			: 0;
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:179]: You should use `sumCurrencies` here

src/lib/server/cart.ts
```diff
@@ -69,7 +69,7 @@ export async function addToCartInDb(
 		throw error(400, "Product can't be added to basket");
 	}
 
-	if (params.customPrice && !product.payWhatYouWant) {
+	if (params.customPrice && !product.payWhatYouWant && !product.hasVariations) {
```
coyotte508, line 72:
```suggestion
	if (params.customPrice && !product.payWhatYouWant) {
```

src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/+page.server.ts
```diff
@@ -222,7 +220,7 @@ export const actions: Actions = {
 					...(!parsed.depositPercentage && { deposit: '' }),
 					...(!parsed.vatProfileId && { vatProfileId: '' }),
 					...(!parsed.restrictPaymentMethods && { paymentMethods: '' }),
-					...(!hasVariations && { variations: '', variationLabels: '' })
+					...(!hasVariations && { variationPrices: '', variationLabels: '' })
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/+page.server.ts:223]:
```suggestion
					...(!hasVariations && { variations: '', variationLabels: '' })
```

src/routes/(app)/admin[[hash=admin_hash]]/product/product-schema.ts
```diff
@@ -64,7 +64,11 @@ export const productBaseSchema = () => ({
 		.array(
 			z.object({
 				name: z.string().trim(),
-				value: z.string().trim()
+				value: z.string().trim(),
+				price: z
+					.string()
+					.default('0')
+					.transform((val) => Number(val))
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/product-schema.ts:71]: It's here that you should use `parsePriceAmount`, like for `priceAmount` above.

Here you can do the same regex as `priceAmount`, and call `parsePriceAmount` after the validation (once you know the currency used)

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -166,18 +167,37 @@ async function addToCart({ params, request, locals }: RequestEvent) {
...
+						currency: runtimeConfig.mainCurrency
+					}))
+			  )
+			: 0;
 
 	const customPrice =
 		customPriceAmount && customPriceCurrency
 			? {
-					amount: parsePriceAmount(customPriceAmount, customPriceCurrency),
+					amount: Number(customPriceAmount),
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:187]: Revert? We do want to call `parsePriceAmount` when the user adds a custom price for a product that allows it.

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -166,18 +167,37 @@ async function addToCart({ params, request, locals }: RequestEvent) {
...
+					currency: runtimeConfig.mainCurrency
+			  }
 			: undefined;
-
 	await addToCartInDb(product, quantity, {
 		user: userIdentifier(locals),
-		...(product.payWhatYouWant && { customPrice }),
+		...((product.payWhatYouWant || (product.hasVariations && product.standalone)) && {
+			customPrice
+		}),
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:200]:
```suggestion
		...(customPrice && { customPrice }),
```

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -166,18 +167,37 @@ async function addToCart({ params, request, locals }: RequestEvent) {
...
+					runtimeConfig.mainCurrency,
+					Object.entries(chosenVariations).map((variation) => ({
+						amount:
+							product.variations?.find(
+								(vari) => variation[0] === vari.name && variation[1] === vari.value
+							)?.price ?? 0,
+						currency: runtimeConfig.mainCurrency
+					}))
+			  )
+			: 0;
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:182]: This should be moved to the `addToCartInDb` function, so the logic is present for each caller of the function and does not need to be duplicated

Also, in that function, we should make sure that:

- Only one variation per category is chosen
- At least One variation for each caetgory is chosen

If a variation has a negative price offset, it will prevent a user from adding that variation many times to reduce the price

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -166,18 +167,37 @@ async function addToCart({ params, request, locals }: RequestEvent) {
 			chosenVariations: z.record(z.string(), z.string()).optional()
 		})
 		.parse(json);
+	const variationPrice =
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:170]:
```suggestion
	const variationPriceDelta =
```

src/routes/(app)/product/[id]/+page.server.ts
Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.server.ts]: Minimum variation price = 0

Comment (coyotte508) [reply src/routes/(app)/product/[id]/+page.server.ts]: > Minimum variation price = 0

I'm not sure what you mean

From https://github.com/B2Bitcoin/beBOP/issues/1472 we can see in the screenshot a price delta of -10 for the first variation.

If a user adds the same variation many times they could bring down the total price of the product to 0 no?

Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.server.ts]: We can no more have a price delta of -10

Comment (ithiame) [reply src/routes/(app)/product/[id]/+page.server.ts]: price delta minimum = 0 when we create product

Comment (coyotte508) [reply src/routes/(app)/product/[id]/+page.server.ts]: By the way, here you set `customPrice = sum(variations)`, it should be `price = sum(originalPrice, variations)` no?

Even if we cannot have variations with negative price deltas (@Tirodem ?) we should still enforce integrity of variation choices.

And the rest of the comment (moving the logic to the `addToCartInDb`) still holds

src/lib/server/cart.ts
```diff
@@ -142,6 +156,8 @@ export async function addToCartInDb(
 			params.customPrice.amount,
 			toCurrency(params.customPrice.currency, product.price.amount, product.price.currency)
 		);
+	} else if (variationPriceDelta > 0) {
+		params.customPrice = { amount: variationPriceDelta, currency: runtimeConfig.mainCurrency };
```
coyotte508, line 160: `customPrice` is not a delta, so it should be in addition to the original product price

src/lib/server/cart.ts
```diff
@@ -114,7 +115,20 @@ export async function addToCartInDb(
...
-
+	const variationPriceArray = params.chosenVariations
+		? Object.entries(params.chosenVariations).map((variation) => ({
+				amount:
+					product.variations?.find(
+						(vari) => variation[0] === vari.name && variation[1] === vari.value
+					)?.price ?? 0,
+				currency: runtimeConfig.mainCurrency
+		  }))
+		: [];
```
coyotte508, line 126: Please add a check that if a product has variations in DB, a variation for each kind should be chosen by the customer (and not more than one)

also the same check in `createOrder` (in case the variations of the product changed between adding to cart & making order)

src/lib/server/cart.ts
```diff
@@ -114,7 +116,35 @@ export async function addToCartInDb(
...
+	let variationPriceArray: Price[] = [];
+	if (params.chosenVariations) {
+		const variationNamesInDB =
+			Object.keys(product.variationLabels?.names || []).map((key) => key) || [];
+
+		const chosenVariationNames = Object.keys(params.chosenVariations);
+
+		const allVariationsChosen =
+			variationNamesInDB.length === chosenVariationNames.length &&
+			variationNamesInDB.every((name) => chosenVariationNames.includes(name));
```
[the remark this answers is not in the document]
ithiame: That's why i use the key ?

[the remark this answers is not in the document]
coyotte508: but `variationLabels` is only for UI, you cannot rely on it, it could become inconsistent. 

The source of truth is `variations`, we should use it, and not the derived `variationLabels`

src/lib/server/cart.ts
```diff
@@ -114,7 +116,34 @@ export async function addToCartInDb(
 	) {
 		throw error(400, 'Cart has too many items');
 	}
-
+	let variationPriceArray: Price[] = [];
+	if (product.variations?.length) {
+		const variationNamesInDB = [...new Set(product.variations.map((vari) => vari.name))];
+
+		const chosenVariationNames = Object.keys(params.chosenVariations || []);
```
coyotte508, line 123:
```suggestion
		const chosenVariationNames = Object.keys(params.chosenVariations ?? {});
```

src/lib/server/cart.ts
```diff
@@ -142,6 +171,14 @@ export async function addToCartInDb(
 			params.customPrice.amount,
 			toCurrency(params.customPrice.currency, product.price.amount, product.price.currency)
 		);
+	} else if (variationPriceDelta > 0) {
```
coyotte508, line 174: @Tirodem can you confirm that we don't allow a variation with negative pricing to decrease the price below the product's base price?

For example if there's a product at 100€ and a variation with -10€, if I chose that variation, how much do I pay?

This seems suspect to me

src/lib/server/orders.ts
```diff
@@ -659,6 +659,19 @@ export async function createOrder(
 		throw error(400, "You can't use free payment method on this order");
 	}
 
+	items.forEach((item) => {
```
coyotte508, line 662: use `for (const ... of)` instead of `xxx.forEach`

src/lib/server/orders.ts
```diff
@@ -659,6 +659,19 @@ export async function createOrder(
...
+			const chosenVariationNames = Object.keys(item.chosenVariations);
+			const allVariationsChosen =
+				variationNamesInDB.length === chosenVariationNames.length &&
+				variationNamesInDB.every((name) => chosenVariationNames.includes(name));
+
+			if (!allVariationsChosen) {
+				throw error(400, 'error matching on variations choice');
+			}
+		}
+	});
```
coyotte508, line 674: We should re-compute `customPrice` when there are variations.

In case the variation price changes between adding to cart and creating the order

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -129,6 +129,22 @@
...
+		);
+		for (const [key, value] of Object.entries(selectedVariations)) {
+			customAmount += toCurrency(
+				data.currencies.main,
+				data.product.variations?.find(
+					(variation) => variation.name === key && variation.value === value
+				)?.price ?? 0,
+				data.product.price.currency
+			);
+		}
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:146]: use `sumCurrencies` here - it takes care of some details with floating point precision

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -129,6 +129,22 @@
 	}
 
 	let selectedVariations: Record<string, string> = {};
+	$: if (data.product.hasVariations && data.product.variationLabels) {
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:132]:
```suggestion
	$: if (data.product.hasVariations && data.product.variations?.length) {
```

src/lib/server/orders.ts
```diff
@@ -659,6 +659,19 @@ export async function createOrder(
 		throw error(400, "You can't use free payment method on this order");
 	}
 
+	items.forEach((item) => {
+		if (item.chosenVariations && item.product.variations) {
```
coyotte508, line 663: We don't care if `item.chosenVariations`, we only care if `item.product.variations?.length`.

If the backend added variations between add to cart & order creation, we should still prevent the order from being created if there are no chosen variations.

src/lib/server/cart.ts
Tirodem: We removed negative variation because, for a 1st version, it introduced complex needs.
So I confirm we aren't supposed to make negative prive variation, only null or positive.

src/lib/server/orders.ts
Tirodem: On a functional behavior PoV I'm more into "price displayed when adding to cart should stay to the end" (except for currency variation, that are fixed once a cart becomes an order through checkout).
I let you decide @coyotte508 but for me it's not so much of an issue if we add a product 1 with variation A to cart, the price change, and we still have the same price on /checkout.
It makes me think more about product price change while a cart exist and is not an order yet, but, for variation price ,I don't really care for now that we refresh it.

src/lib/server/cart.ts
ithiame: I have already changed the front input to take min 0

coyotte508: it should be the backend input on the admin too

and this `variationDelta > 0` condition can be removed

ithiame: We have this in `productForm.svelte `
```
	Price difference<input
		                 type="number"
				 name="variations[{i}].price"
				class="form-input"
				value={variation.price}
				min="0"/>
```

src/routes/(app)/admin[[hash=admin_hash]]/product/product-schema.ts
Comment (coyotte508) [reply src/routes/(app)/admin[[hash=admin_hash]]/product/product-schema.ts:71]: Also do a min(0) check

src/lib/server/cart.ts
coyotte508: See comment here too: https://github.com/B2Bitcoin/beBOP/pull/1491#discussion_r1806157934

src/lib/server/orders.ts
```diff
@@ -659,6 +659,46 @@ export async function createOrder(
 		throw error(400, "You can't use free payment method on this order");
 	}
 
+	let variationPriceArray: Price[] = [];
+
+	for (const item of items) {
+		if (item.product.variations) {
```
coyotte508, line 665:
```suggestion
		if (item.product.variations?.length && product.payWhatYouWant) {
```

src/lib/server/orders.ts
```diff
@@ -659,6 +659,46 @@ export async function createOrder(
...
+					: [];
+			} else {
+				throw error(400, 'error matching on variations choice');
+			}
+		}
+		const variationPriceDelta =
+			item.product.hasVariations && item.chosenVariations
+				? sumCurrency(item.product.price.currency, variationPriceArray)
+				: 0;
+		if (variationPriceDelta > 0) {
```
coyotte508, line 691: remove this condition, like with the cart?

src/lib/server/orders.ts
```diff
@@ -659,6 +659,46 @@ export async function createOrder(
...
+					  }))
+					: [];
+			} else {
+				throw error(400, 'error matching on variations choice');
+			}
+		}
+		const variationPriceDelta =
+			item.product.hasVariations && item.chosenVariations
+				? sumCurrency(item.product.price.currency, variationPriceArray)
+				: 0;
```
coyotte508, line 690: Can you maybe put this in  separate function - used both here and in `addToCart`?

src/lib/server/product.ts
```diff
@@ -155,3 +159,34 @@ export async function amountOfProductSold(productId: string): Promise<number> {
...
+) {
+	let variationPriceArray: Price[] = [];
+	const variationNamesInDB = [...new Set(product.variations?.map((vari) => vari.name))];
+	const chosenVariationNames = Object.keys(chosenVariations ?? {});
+
+	const allVariationsChosen =
+		variationNamesInDB.length === chosenVariationNames.length &&
+		variationNamesInDB.every((name) => chosenVariationNames.includes(name));
+
+	if (allVariationsChosen) {
```
coyotte508, line 175: Can you move this to a separate function

So we have two functions:

- `checkProductVariationsIntegrity` here in `src/lib/server/product.ts`
- `productPriceWithVariations` in `src/lib/types/Product.ts`

So then we can reuse `productPriceWithVariations` in the frontend - in `product/[id]/+page.svelte`

src/lib/server/product.ts
```diff
@@ -155,3 +159,34 @@ export async function amountOfProductSold(productId: string): Promise<number> {
...
+					currency: product.price.currency
+			  }))
+			: [];
+	} else {
+		throw error(400, 'error matching on variations choice');
+	}
+
+	return product.hasVariations && chosenVariations
+		? sumCurrency(product.price.currency, [...variationPriceArray, product.price])
+		: 0;
```
coyotte508, line 191:
```suggestion
	return sumCurrency(product.price.currency, [...variationPriceArray, product.price]);
```

The check is useless or should have been done before this point.

(and we shouldn't return `0`)

src/lib/types/Product.ts
```diff
@@ -104,3 +106,34 @@ export function isPreorder(
 export function oneMaxPerLine(p: Pick<Product, 'standalone' | 'type'>) {
 	return p.standalone || p.type === 'subscription';
 }
+export function productPriceWithVariations(
+	product: Pick<Product, 'name' | '_id' | 'price' | 'variations'>,
+	chosenVariations: Record<string, string> | undefined
+) {
+	let variationPriceArray: Price[] = [];
+
+	variationPriceArray = chosenVariations
```
coyotte508, line 115: can be directly declared as `const variationPriceArray: Price[] = ...;`

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -129,6 +130,9 @@
...
+		customAmount = productPriceWithVariations(data.product, selectedVariations);
+	}
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:135]:
```suggestion
	$: if (data.product.hasVariations) {
		customAmount = productPriceWithVariations(data.product, selectedVariations);
	}
```

test should be consistent here & in the html code below

coyotte508 (approved):

PR #1499, opened 2024-10-11
🐛 track incrementations as events in database

ithiame:
🐛 🔬 Challenge gauge broken / flaky : add helpful tools https://github.com/B2Bitcoin/beBOP/issues/1496

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -1514,23 +1514,23 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
 						}))
 				  );
 
-		await collections.challenges.updateOne(
-			{ _id: challenge._id },
-			{
-				$inc: { progress: increase },
-				$push: {
-					...(increase && {
+		if (increase) {
+			await collections.challenges.updateOne(
+				{ _id: challenge._id },
+				{
+					$inc: { progress: increase },
+					$push: {
 						event: {
 							type: 'progress',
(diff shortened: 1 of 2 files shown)


src/lib/types/Challenge.ts
```diff
@@ -14,6 +15,11 @@ export type Challenge = Timestamps & {
 
 	beginsAt: Date;
 	endsAt: Date;
+	incrementation?: {
+		incrementedDate: Date;
+		order: Order['_id'];
+		increment: number;
+	}[];
```
coyotte508, line 22:
```suggestion
	events?: {
		type: "increment",
		at: Date;
		order: Order['_id'];
		amount: number;
	}[];
```

ok with this change in the data model

Tirodem:
Currently, on live pilot tests, tracks are "0", but that might be the rootcause symptom.

src/lib/server/orders.ts
```diff
@@ -1519,12 +1519,14 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
 			{
 				$inc: { progress: increase },
 				$push: {
-					event: {
-						type: 'progress',
-						at: new Date(),
-						order: order._id,
-						amount: increase
-					}
+					...(increase && {
```
coyotte508, line 1522: At this point you should just not do the DB call

coyotte508 (approved):

Issue #1157, opened 2024-03-14
🦋 Apply directly a filter for the last month on /admin/reporting and apply it to products/payments/orders details


Tirodem:
Following https://github.com/B2Bitcoin/beBOP/pull/1152
We need to fitler the view before it becomes too heavy to load for the page to open.

PR #1503, opened 2024-10-14
 Apply directly a filter for the last month on /admin/reporting and a…

ithiame:
…pply it to products/payments/orders details #1157

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts
@@ -4,7 +4,16 @@ import { sum } from '$lib/utils/sum';
 import { pojo } from '$lib/server/pojo.js';
 
 export async function load() {
-	const orders = await collections.orders.find().sort({ createdAt: -1 }).toArray();
+	const now = new Date();
+
+	const startOfCurrentMonth = new Date(now.getFullYear(), now.getMonth(), 1);
+	const startOfLastMonth = new Date(now.getFullYear(), now.getMonth() - 1, 1);
+	const endOfLastMonth = new Date(startOfCurrentMonth.getTime() - 1);
+
+	const orders = await collections.orders
+		.find({ createdAt: { $gt: startOfLastMonth, $lt: endOfLastMonth } })
+		.sort({ createdAt: -1 })
+		.toArray();
 
 	return {
 		orders: orders.map((order) => ({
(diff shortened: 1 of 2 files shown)


src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts
```diff
@@ -4,7 +4,16 @@ import { sum } from '$lib/utils/sum';
 import { pojo } from '$lib/server/pojo.js';
 
 export async function load() {
-	const orders = await collections.orders.find().sort({ createdAt: -1 }).toArray();
+	const now = new Date();
+
+	const startOfCurrentMonth = new Date(now.getFullYear(), now.getMonth(), 1);
+	const startOfLastMonth = new Date(now.getFullYear(), now.getMonth() - 1, 1);
+	const endOfLastMonth = new Date(startOfCurrentMonth.getTime() - 1);
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts:11]: Can you use the `startOfMonth`, `endOfMonth`, ... of `date-fns`?

coyotte508:
Does "last month" mean "current month" or "previous month" @Tirodem ?

Tirodem:
sorry : current month

src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts
```diff
@@ -2,9 +2,18 @@ import { collections } from '$lib/server/database';
...
 import { pojo } from '$lib/server/pojo.js';
+import { endOfMonth, startOfMonth, subMonths } from 'date-fns';
 
 export async function load() {
-	const orders = await collections.orders.find().sort({ createdAt: -1 }).toArray();
+	const orders = await collections.orders
+		.find({
+			createdAt: {
+				$gt: startOfMonth(subMonths(new Date(), 1)),
+				$lt: endOfMonth(subMonths(new Date(), 1))
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts:12]:
```suggestion
				$gte: startOfMonth(new Date(), 1)
```

coyotte508 (approved):

Tirodem:
"Not bad but"

1/ if we apply current month, I expect the reporting filter on the 2nd part of the page to be set as the current month
image
Currently, per default, it's 01/2023

2/ I can change the month for reporting synthesis, but order detail is now locked to the current month, and event by changing the later month selector, I'm unable to see earlier orders :
image

Limiting default-loaded range of orders if one good thing, but :
1/ if we display current month, the existing filter must be pre-filled with current month value
2/ we cannot only load orders from this month without being able to check previous months orders, so we need the month selector to apply also on order / product / payment detail, and not only on synthesis

Unless I missed something on this PR UI, I cannot have history detail of previous month and earlier.

Issue #1461, opened 2024-09-24
🦋 Manage decimals on price


Tirodem:
![image](https://github.com/user-attachments/assets/974840da-bdea-42aa-a9fd-3e3d2032ec42)
Current use (in FR & CH) is to display "5.90" instead of "5.9".
But some countries would rather "5.9" instead of "5.90".
We need to create a setting in /admin/config to allow the be-BOP owner to do as he/she wants.
Plus, we need to study if there's countries with 0 or 3 decimals on their currencies.

coyotte508:
We already have `FRACTION_DIGITS_PER_CURRENCY`

```
export const FRACTION_DIGITS_PER_CURRENCY = Object.freeze({
	BTC: 8,
	CHF: 2,
	EUR: 2,
	USD: 2,
	ZAR: 2,
	XOF: 2,
	XAF: 2,
	CDF: 2,
	SAT: 0
}) satisfies Record<Currency, number>;
```

It's something that needs to be edited when adding a new currency anyway.

We can base ourself on this for the number of digits

Tirodem:
Sure it's a super good base to work on.
We just need to avoid situations where we displayed that something cost 2.1200000000 BTC if price is 2.12 only (rare case I admin due to conversion rate and BTC amount, but on BTC as price reference currency store it can become a problem).

PR #1510, opened 2024-10-16
Manage decimals on price #1461

ithiame:


coyotte508:
I think the main problem in https://github.com/B2Bitcoin/beBOP/issues/1461 is that we already have `FRACTION_DIGITS: 2` but we display `5.9` instead of `5.90`

This PR doesn't solve that.

IMO we can start with:

- If there are decimals, display them (min 2, max `FRACTION_DIGITS`)
- If there are not decimals, don't display them

Then if we encounter a specific client that has specific needs (or a specific currency) we can adapt

Tirodem:
image
image
image
image
I don't see the change :(

Tirodem:
> I think the main problem in #1461 is that we already have `FRACTION_DIGITS: 2` but we display `5.9 €` instead of `5.90 €`
> 
> This PR doesn't solve that.
> 
> IMO we can start with:
> 
> * If there are decimals, display them (min 2, max `FRACTION_DIGITS`)
> * If there are no decimals, don't display them
> 
> Then if we encounter a specific client that has specific needs (or a specific currency) we can adapt the codebase with a more involved solution

yes, that's the target issue, 5.6€ should be 5.60€, while 5€ should stay 5€, but 0.02133BTC should stay 0.02133BTC and 123545 SATS should stay 123545 SATS

coyotte508 (approved):

coyotte508:
and @ithiame the other problem is for BTC, we don't want to display 8 fraction digits if I input `1.27` as @Tirodem explained

coyotte508 (approved):

PR #1511, opened 2024-10-16
Age limitation

ithiame:


src/routes/(app)/+layout.svelte
```diff
@@ -113,14 +113,41 @@
...
+				En validant ce formulaire pour continuer la navigation, vous admettez avoir l'âge légal dans
+				votre pays.
+			</p>
+			<p class="my-4">
+				On fait ça parce que <a href={data.ageLegalReason} class="underline">telle page</a> nous le demande
+			</p>
+			<label class="label-checkbox my-4">
+				<input type="checkbox" class="form-checkbox" bind:checked={ageWarning} />
+				Compris chef
+			</label>
```
coyotte508, line 142: Uhhh I don't think we want this as IS!!

This was just an example.

It should be configurable (maybe CMS page)

src/lib/server/runtime-config.ts
```diff
@@ -59,6 +59,8 @@ const baseConfig = {
 	 *
 	 * For Phoenixd max value is 60 minutes
 	 */
+	ageRestriction: false,
+	ageLegalReason: '',
```
coyotte508, line 63: should be a single object

src/lib/translations/en.json
```diff
@@ -13,6 +13,10 @@
 		"vatNumber": "VAT number",
 		"zipCode": "Zip code"
 	},
+	"ageWarning": {
+		"agreement": "Compris chef",
+		"navigate": "Continuer la navigation"
```
coyotte508, line 18: Those are english wordings in French. Check the other translations too, like italian.

Btw @Tirodem do we really want to use the "compris chef" in  the wording?

src/routes/(app)/+layout.svelte
```diff
@@ -113,14 +114,49 @@
...
+					roleId={data.roleId ? data.roleId : ''}
+					specifications={data.cmsAgewallData.specifications}
+					contactForms={data.cmsAgewallData.contactForms}
+					pageName={data.cmsAgewall?.title}
+					websiteLink={data.websiteLink}
+					brandName={data.brandName}
+					sessionEmail={data.email}
+					countdowns={data.cmsAgewallData.countdowns}
+					galleries={data.cmsAgewallData.galleries}
+					class={data.hideCmsZonesOnMobile ? 'hidden lg:contents' : ''}
```
coyotte508, line 145: btw for the props that have the same name you should be able to do `...data.cmsAgeWallData`

```suggestion
					{...data.cmsAgeWallData}
					roleId={data.roleId ? data.roleId : ''}
					pageName={data.cmsAgewall?.title}
					websiteLink={data.websiteLink}
					brandName={data.brandName}
					sessionEmail={data.email}
					class={data.hideCmsZonesOnMobile ? 'hidden lg:contents' : ''}
```

(and brandName & websiteLink could probably be put inside a global store at layout init, like currency exchange rates, to avoid passing them around)

src/lib/translations/en.json
coyotte508: so a sub-param in `runtimeConfig.ageRestriction` @ithiame

src/lib/server/runtime-config.ts
```diff
@@ -59,6 +59,10 @@ const baseConfig = {
 	 *
 	 * For Phoenixd max value is 60 minutes
 	 */
+	age: {
+		restriction: false,
+		legalReason: ''
+	},
```
coyotte508, line 65:
```suggestion
	ageRestriction: {
		enabled: false,
		legalReason: ''
	},
```

src/lib/translations/en.json
Tirodem: Ah, there's no translation for /config so it would be bothersome, so here's the translation key... ok.
Let me think of a wording.

Tirodem: "J'ai pris connaissance des conditions d'accès, les remplit et les respecte."
"I have read the access conditions, fulfill them and respect them."
+ other languages

Tirodem:
I tried :
- as a connected admin to check the checkbox in /admin/age-restriction then I landed on /admin/age-retriction?/update with this :
image
- as a new user on new session, same thing

Why this form ?
This should not be here.
And header / footer should be here (but with disabled links)

For memories, the spec :
image

Also, as it goes, /admin/login page should not be blocked either :
image
(in fact, no /admin page should be blocked)

ithiame:
`agewall `cms content image

image

ithiame:
Uploading Enregistrement 2024-10-17 163542.mp4…

ithiame:
ezgif-5-9e2d6c0430

src/lib/translations/fr.json
```diff
@@ -13,6 +13,10 @@
 		"vatNumber": "Numéro TVA",
 		"zipCode": "Code postal"
 	},
+	"ageWarning": {
+		"agreement": "J'ai pris connaissance des conditions d'accès, les remplit et les respecte.",
```
Tirodem, line 17: "remplis", not "remplit" (my bad)

Tirodem:
On my Chrome were I'm with my session, if I enable the agewall and go to the website outside the /admin, I have the /agewall, then check it and it's good.
On a private session, I always have the /agewall page.
Same for a buddy with Chrome on mobile (no session, never accessed to be-bop).
☣️ huge thing ☣️

Tirodem:
No more session bug from my side ✅ open to review

src/routes/(app)/+layout.server.ts
```diff
@@ -131,6 +132,28 @@ export async function load(params) {
 			  ).then((res) => filterUndef(res))
 			: null
 	]);
+	let cmsAgewall = null;
```
coyotte508, line 135: please type it appropriately

src/routes/(app)/+layout.server.ts
```diff
@@ -131,6 +132,28 @@ export async function load(params) {
...
+	const session = await collections.sessions.findOne({
+		sessionId: params.locals.sessionId
+	});
```
coyotte508, line 138: You should add the `acceptAgeLimitation` directly in `params.locals.acceptAgeLimitation` in `hooks.server.ts`, to avoid querying the DB once more.

eg near here:

```
		event.locals.email = session.email;
		event.locals.npub = session.npub;
		event.locals.sso = session.sso;
```

src/routes/(app)/+page.server.ts
```diff
@@ -40,3 +41,21 @@ export const load = async ({ locals }) => {
...
+export const actions = {
+	navigate: async ({ locals }) => {
+		await collections.sessions.findOneAndUpdate(
+			{
+				sessionId: locals.sessionId
+			},
+			{
+				$set: {
+					updatedAt: new Date(),
+					acceptAgeLimitation: true
```
coyotte508, line 54: please set the `expiresAt` field - like in the originall code from `hooks.server.ts`.

src/routes/(app)/+page.server.ts
```diff
@@ -40,3 +41,21 @@ export const load = async ({ locals }) => {
...
+				}
+			},
+			{ upsert: true }
```
coyotte508, line 57: Since you're upserting, you should add all fields with `$setOnInsert` so that a newly inserted session has all required fields.

(For example, `createdAt`)

src/routes/(app)/admin[[hash=admin_hash]]/age-retriction/+page.server.ts
```diff
@@ -0,0 +1,40 @@
...
+		const json: JsonObject = {};
+
+		for (const [key, value] of formData) {
+			if (value) {
+				set(json, key, value);
+			}
+		}
+		const result = z
+			.object({
+				ageRestriction: z.object({ enabled: z.boolean({ coerce: true }), legalReason: z.string() })
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/age-retriction/+page.server.ts:21]: hm can we do just two values instead of a nested object?

Then we can do regular form data, no need for JSON tricks

src/routes/(app)/+page.server.ts
```diff
@@ -40,3 +43,26 @@ export const load = async ({ locals }) => {
 		layoutReset: cmsPage.fullScreen
 	};
 };
+
+export const actions = {
+	navigate: async ({ locals }) => {
+		await collections.sessions.findOneAndUpdate(
```
coyotte508, line 49:
```suggestion
		await collections.sessions.updateOne(
```

you don't use the result

src/routes/(app)/+page.server.ts
```diff
@@ -40,3 +43,26 @@ export const load = async ({ locals }) => {
...
+			},
+			{
+				$set: {
+					updatedAt: new Date(),
+					acceptAgeLimitation: true,
+					expiresAt: addYears(new Date(), 1)
+				},
+				$setOnInsert: {
+					createdAt: new Date(),
+					_id: new ObjectId()
```
coyotte508, line 61: This field doesn't need to be auto-created

coyotte508 (approved):

Issue #1483, opened 2024-10-07
🦋 Make CTAs taller depending of translation


Tirodem:
Current behavior on cart addition in italian :
![image](https://github.com/user-attachments/assets/fb19ed64-9153-441a-a611-53835eeef334)

Same in french :
![image](https://github.com/user-attachments/assets/022db6f8-36f2-4f7b-8886-868bd9354e9c)

If that happens, make the button bigger, the neighbor button bigger if they are side by side, and also the container box bigger to fit size :
![image](https://github.com/user-attachments/assets/5e1514d3-f6e7-4299-9811-919b4262a56e)

Each CTA should be treated, so, check :
- product page (add to cart, checkout, custom cta)
- identity
- session
- history
- checkout
- order

Tag widgets are excluded, except product widget, has they have generic text, so have a look on them.

PR #1516, opened 2024-10-18
🦋 Make CTAs taller depending of translation #1483

ithiame:


ithiame:
image

image

src/routes/(app)/cart/+page.svelte
```diff
@@ -97,7 +97,7 @@
 			>
 				<div class="gap-4 flex flex-col md:flex-row">
 					<label class="form-label w-[20em]">
-						Fill product alias
+						{t('cart.fillProductAlias')}
```
coyotte508, line 100: Note that this is POS-only so it was OK to not translate it

coyotte508 (approved):

Tirodem:
I'll say it's OK as it's a good improvement but create more tickets for specific cases.

Issue #1515, opened 2024-10-16
🦋 Raise error on product form if 2 variations have the same name + value


Tirodem:
Imagine my "tee" product has these variations (name / value / price variation) :
- Color / Red / 0
- Color / Orange / 5
- Color / Yellow / 10

If I add this one and save :
- Color / Orange / 15

My product will now be :
- Color / Red / 0
- Color / Orange / 15
- Color / Yellow / 10

I's not so bad but it can leads to mistakes.

What I want in the future is : is two or more variations have the same variation name and value (here "Color" and "Orange", raise error on form UI, and highlight in red the variations inputs as error as long as the user doesn't manually delete the duplicate entries himself / herself.

PR #1531, opened 2024-10-23
Raise error on product form if 2 variations have the same name + value

ithiame:
🦋 Raise error on product form if 2 variations have the same name + value #1515

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -128,17 +128,13 @@
 			} else {
 				priceAmountElement.setCustomValidity('');
 			}
-			let seen = new Set();
-			for (let i = 0; i < variationLabelsValues.length; i++) {
-				const key = `${(
-					variationLabelsNames[i] ||
-					product.variations?.[i].name ||
-					''
-				).toLowerCase()}-${(
-					variationLabelsValues[i] ||
-					product.variations?.[i].value ||
-					''
-				).toLowerCase()}`;
+			const seen = new Set<string>();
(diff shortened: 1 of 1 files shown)


src/lib/components/ProductForm.svelte
```diff
@@ -126,7 +128,26 @@
 			} else {
 				priceAmountElement.setCustomValidity('');
 			}
+			let seen = new Set();
```
coyotte508, line 131: use `const`, and type it appropriately, `new Set<string>()`

src/lib/components/ProductForm.svelte
```diff
@@ -126,7 +128,26 @@
...
+			for (let i = 0; i < variationLabelsValues.length; i++) {
```
coyotte508, line 132: use `for const of`

src/lib/components/ProductForm.svelte
```diff
@@ -126,7 +128,26 @@
...
+				const key = `${(
+					variationLabelsNames[i] ||
+					product.variations?.[i].name ||
+					''
+				).toLowerCase()}-${(
+					variationLabelsValues[i] ||
+					product.variations?.[i].value ||
+					''
+				).toLowerCase()}`;
```
coyotte508, line 141: Use `JSON.stringify({name, value})` - that way if I have a variation with name "a-b" and val "c", and another with "a" and "b-c", it won't cause issues

coyotte508 (approved):

Tirodem:
image

Issue #1536, opened 2024-10-24
☣️ Delivery fees for products with physical component are displayed on /cart and /order but not anymore on /checkout


Tirodem:
![image](https://github.com/user-attachments/assets/808e97ad-b791-42ba-a6eb-1aa9f2e7faa2)
![image](https://github.com/user-attachments/assets/02651cc7-4e06-4e9e-aabf-0e001fa5f652)
![image](https://github.com/user-attachments/assets/fbf682ca-7324-4d1a-ad8d-42bae9105e49)
![image](https://github.com/user-attachments/assets/3bc0121a-4745-49be-a0f2-5baa5876dc2f)
As seen, currently, on /checkout, delivery fees aren't displayed anymore.
So price is different between cart, checkout and order.
This is a huge thing, so :
- so far, do no deploy on current pilot (this would bring huge problems with some live ones with a lot of delivery)
- find why we lost delivery fees on checkout cart content
- fix it ASAP and check it doesn't break anything else

Tirodem:
Most surprising is this case :
![image](https://github.com/user-attachments/assets/c8970e73-8a00-4822-b398-e9c688c2cd28)
- we detect there's delivery fees
- we create a 8.1% VAT rate line
- we don't display the delivery fees
- we consider that delivery fees are null
- so the displayed 8.1% VAT rate line for delivery fees is 0

PR #1537, opened 2024-10-24
Delivery fees for products with physical component are displayed on /…

ithiame:
…cart and /order but not anymore on /checkout #1536


coyotte508 (approved):
ok so @Tirodem it seems it happend for bebops where `		Default enable "this is a fully-paid on-location physical order"` from @ithiame 's PR

Issue #1535, opened 2024-10-24
🐛 "1 BTC = 1 BTC at the time of the order" if BTC is the main currency, whatever the accounting currency is, on invoice


Tirodem:
Test order :
![image](https://github.com/user-attachments/assets/f1e733f1-b1b8-4e08-a242-178820b1ee9e)
Bothersome thing :
![image](https://github.com/user-attachments/assets/354bc783-24a7-418e-8429-faa4b22f2a34)
![image](https://github.com/user-attachments/assets/8db29884-a715-4f5a-81bb-43ab419323bf)
That being the current be-BOP configuration :
![image](https://github.com/user-attachments/assets/9cb13e18-2281-4e8c-a86b-353b6abfa054)
And that being stored in database :
![image](https://github.com/user-attachments/assets/f8ad3268-8c7c-46c3-8ac2-0482dc7d58e8)
It should be better to use these informations to give the BTC value (main currency) in CHF (accountability currency) as the time.
It should not work only when BTC is a second currency, we should change the reference since the introduction of the accountability currency.

PR #1539, opened 2024-10-24
'1 BTC = 1 BTC at the time of the order' if BTC is the main currency,…

ithiame:
… whatever the accounting currency is, on invoice #1535


coyotte508:
I thought that `accounting` was purely internal, now it's exposed to the customers as well?

(no problem, PR looks good and you can merge it, it just seems weird to me)

Tirodem:
> I thought that `accounting` was purely internal, now it's exposed to the customers as well?
> 
> (no problem, PR looks good and you can merge it, it just seems weird to me)

Always have been, there's a link for that.
image

That's called a "Pro-forma" and is exposed to customer.

Tirodem:
Good for proforma.
image
So I'm merging as you said as it's functionnally good.

Tirodem (approved):

PR #1543, opened 2024-10-25
Adapt custom CTA on product page to content + limit content input

ithiame:


coyotte508 (approved):

Tirodem:
Chain breaking is not done
image
:x:

Tirodem:
Too brutal ❌ 
Here's my fill :
`dans ma valise il y a carlos et un sac de nougats`
Here's my CTA :
image
Ok, chain is broken, but there's no reason to put a break between "s" and "ac" on "sac".
We need a smarter way of breaking chains.

Tirodem:
And there's no character limit on input :
image

ithiame:
Before:
image

After:
image

PR #1551, opened 2024-10-28
Add pagination and filters on /admin/product#1532

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/product/+page.server.ts
```diff
@@ -5,18 +5,54 @@ import { set } from 'lodash-es';
...
+				break;
+			case 'with-stock':
+				query['stock.available'] = { $gt: 0 };
+				break;
+			case 'no-stock':
+				query['stock.available'] = { $lte: 0 };
+
+				break;
+		}
+	}
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/+page.server.ts:53]: why else/if ?

user should be able to specify several filter, at least thebackend should hadnle it

coyotte508 (approved):

Tirodem:
image
- I go to page 5
- I apply a filter, there's only 4 pages of result
- My page 5 is empty
Usually, on apps, it's worse, it breaks.
But applying a filter or resetting it should always reset pagination and bring to page 1.

Tirodem:
I'll merge it as it is because we have the same issue on /admin/order but we need a fix for that soon.
image
I'm creating the new ticket.

Issue #1558, opened 2024-10-29
🐛 When using filters on /admin/products and /admin/order, reset pagination and go to page 1


Tirodem:
![image](https://github.com/user-attachments/assets/f4d729d5-040e-4a8a-be8a-6d5e07bff588)
![image](https://github.com/user-attachments/assets/c0dcde9b-5cf1-4013-b88e-3637c0485e91)

PR #1560, opened 2024-10-29
When using filters on /admin/products and /admin/order, reset paginat…

ithiame:
…ion and go to page 1 #1558


src/routes/(app)/admin[[hash=admin_hash]]/order/+page.svelte
```diff
@@ -74,7 +77,7 @@
 			<input class="form-input" type="text" name="npub" placeholder="search order npub" />
 		</label>
 		<label class="form-label w-auto mt-8 flex flex-row">
-			<input type="submit" value="🔍" class="btn btn-gray" />
+			<input type="submit" value="🔍" class="btn btn-gray" on:click={() => (next = 0)} />
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/order/+page.svelte:80]: by the way it should be declared with `let next` no? Instead of `$: next`

coyotte508 (approved):

Issue #1569, opened 2024-10-30
🐛 On checkout, if "ceci est une commande professionnelle" is checked, "prénom" and "nom" inputs should not be required


Tirodem:
Owner first name and last name are now always required for billing a company.

PR #1572, opened 2024-10-31
 🐛 On checkout, if 'ceci est une commande professionnelle' is checked…

ithiame:
…, 'prénom' and 'nom' inputs should not be required #1569

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,7 +12,11 @@
 	},
 	"githubPullRequests.ignoredPullRequestBranches": ["main", "main"],
 	"i18n-ally.keystyle": "nested",
-	"i18n-ally.localesPaths": ["src/lib/translations"],
+	"i18n-ally.localesPaths": [
+		"src/routes/script/language",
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
coyotte508, line 1: revert?

ithiame: deleted

src/routes/(app)/checkout/+page.server.ts
```diff
@@ -185,8 +185,8 @@ export const actions = {
 			? z
 					.object({
 						billing: z.object({
-							firstName: z.string().min(1),
-							lastName: z.string().min(1),
+							firstName: z.string().optional().default(''),
+							lastName: z.string().optional().default(''),
```
coyotte508, line 189: you should check `isCompany` first, and then change this rule based on `isCompany`.

And you only need one of `optional()` and `default()`, not both

coyotte508 (approved):

Issue #1573, opened 2024-11-01
🐛 We cannot remove label from an order anymore


Tirodem:
An order has a label.
![image](https://github.com/user-attachments/assets/aa8cafe7-fed1-4c33-9e26-c78dd0a19711)

I delete and save it :
![image](https://github.com/user-attachments/assets/a1e7b4fb-307b-418c-880e-89183ff82915)

Result :
![image](https://github.com/user-attachments/assets/da062601-998d-4249-a17b-34961b9fb07d)

That is not supposed to happend, labels aren't mandatory.
Please fix it.

⚠️ ⚠️ ⚠️ ⚠️ ⚠️
**Also**, please use this ticket to check this error 422 implementation.
Since days, I had to open new issues to fix errors about it.
I want to know where it's used to determine if it's legit or not, it should not be triggered on non-mandatory information deletion, it's painful, it's a timewaster.

PR #1574, opened 2024-11-04
 🐛 We cannot remove label from an order anymore #1573

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/label/+page.server.ts
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/label/+page.server.ts:1]: it should unset labels when not defined - otherwise it will either do nothing or set it to `null`

coyotte508 (approved):

Issue #1458, opened 2024-09-23
⚡ Allow to generate a QR Code with Bolt 12 and get Bolt 12 address through Phoenix Server


Tirodem:
Resource : https://phoenix.acinq.co/server/api#get-bolt12-offer

1st thing : allow admin to get his/her Bolt12 address through Phoenix Server interface on be-BOP
Current :
![image](https://github.com/user-attachments/assets/e0bd9836-6941-4c63-82a6-a4d335a69b82)
After : 
![image](https://github.com/user-attachments/assets/568abc05-12fd-4544-8af8-be4d5047ab5f)
On click :
![image](https://github.com/user-attachments/assets/e3d9b7f9-fad8-4dd7-a35f-9dca1abfb289)

2nd thing : on CMS Page, implement a [QRCode=] code (other usages will be added on day).
If [QRCode=Bolt12], generate this QR Code from Phoenix Server (or make it stocked on database when reading it the first time) on CMS Page.
![image](https://github.com/user-attachments/assets/b1891c3e-abee-459f-8cb9-dcb450d27d37)

Tirodem:
Was ok on mainnet on this branch : https://github.com/B2Bitcoin/beBOP/tree/award-afrobitcoin-experiment ✅ 
please confirm to me that there's no diff or fix in the branch and not in this PR, if OK, I'll merge

PR #1578, opened 2024-11-04
Bolt 12 address phoenixD

ithiame:
⚡ Allow to generate a QR Code with Bolt 12 and get Bolt 12 address through Phoenix Server #1458


src/routes/+layout.server.ts
```diff
@@ -20,6 +21,13 @@ export async function load(event) {
 				return `width=${runtimeConfig.viewportContentWidth}`;
 		}
 	})();
+	let bolt12Address = '';
+
+	try {
+		bolt12Address = await phoenixdGetBolt12();
+	} catch (error) {
+		bolt12Address = 'Bolt12';
+	}
```
coyotte508, line 30: This does an API call to phoenixd on the backend for every single page render. It doesn't even check that phoenixd is configured (with `isPhoenixdConfigured`)

This will add latency to the app, especially if phoenixd has troubles.

Please call it only when necessary. Maybe it's there because of CMS page, but then you can call it when renderning CMS only if the bolt12 qr code is asked for (and only if the QR Code CMS Element is rendered)

coyotte508 (commented):
Looks nice!

Except the mistake

src/lib/server/phoenixd.ts
```diff
@@ -34,6 +34,17 @@ export async function phoenixdBalance(): Promise<{ balanceSat: number; feeCredit
...
+export async function phoenixdGetBolt12(): Promise<string> {
+	const res = await fetch(`${runtimeConfig.phoenixd.url}/getoffer`, {
+		headers: {
+			Authorization: `Basic ${Buffer.from(`:${runtimeConfig.phoenixd.password}`).toString(
+				'base64'
+			)}`
+		}
+	});
+
+	return await res.text();
```
coyotte508, line 46: It should check `res.ok` first, and throw an error if not.

Also, https://phoenix.acinq.co/server/api#get-bolt12-offer seems to say it's always the same address, interesting.

Can you check if the API call still works with a fresh node (without any data associated)?

If it doesn't work, what does an error look like? Sometimes the phoenixd server doesn't probably send an HTTP error code (If it works, then we can merge the PR as is, after the check on `res.ok`)

src/lib/server/phoenixd.ts
```diff
@@ -34,6 +34,21 @@ export async function phoenixdBalance(): Promise<{ balanceSat: number; feeCredit
...
+	const res = await fetch(`${runtimeConfig.phoenixd.url}/getoffer`, {
+		headers: {
+			Authorization: `Basic ${Buffer.from(`:${runtimeConfig.phoenixd.password}`).toString(
+				'base64'
+			)}`
+		}
+	});
+
+	if (!res.ok) {
+		throw error(500, `Error fetching Bolt12 offer: ${res.status} ${res.statusText}`);
```
coyotte508, line 47:
```suggestion
		throw error(500, `Error fetching Bolt12 offer: ${res.status} ${await res.text()}`);
```

ithiame:
Can be merged 

CMS Integration
image

PR #1579, opened 2024-11-04
lightning payment qr code clickable

ithiame:


src/lib/utils/lightningPaymentQr.ts
coyotte508, line 1: Please move this - and `bitcoinPaymentQrCodeString` to `$lib/types/Order.ts`

`$lib/utils` is for more generic util functions. (And the name of the file should match exactly the function, too)

Otherwise, looks good

coyotte508 (approved):

Issue #1567, opened 2024-10-30
🦋 Change max quantity of article by order


Tirodem:
It was 10 before.
![image](https://github.com/user-attachments/assets/f2f2cec7-bef3-4470-806c-56302c344031)

Now it's 99 (hardcoded).
https://github.com/B2Bitcoin/beBOP/commit/59ceffc3449d9a9259082396f13307f4a3737104
https://github.com/B2Bitcoin/beBOP/commit/916ab995bf1c976d2a8a4c60761acbe5beec2787
(but finally https://github.com/B2Bitcoin/beBOP/commit/4e9bb8faa50e7c68de9f6dc806829f9ffc58c3b8 , woopsie)

With our firsts B2B orders, future wine pilot, and future spicy sauce pilot, we need more than 10.

Having a too great max value can lead to misconfiguration, and lead to stock squatting.

BTW, we can make the default qty per order a setting on admin/config or admin/product
And for the max ox the max qty : remove the limiter. No need to put a limit, no need to do a setting.

AmadouAgiltoo:
The modification works; however, two remarks:
@Tirodem, in the end, should the input be set to 99 by default or left blank? 
Also, should we take the maximum stock quantity into account?

PR #1590, opened 2024-11-21
🦋 Change max quantity of article by order #1567

ithiame:


ithiame:
Can be merged ! @Tirodem

AmadouAgiltoo:
The modification works; however, two remarks:
@Tirodem, in the end, should the input be set to 99 by default or left blank? Also, should we take the maximum stock quantity into account?

PR #1598, opened 2024-11-29
Nostr PWYW

coyotte508:


Tirodem:
1/ nostr-cart and real order cart aren't sync, real cart is made with initial PWYW price
image
image

2/ add to cart confirmation message must be changed to take the set amount
image

coyotte508:
3/ seems impossible given the code, we explictly do `			if (product.payWhatYouWant) {` first

coyotte508:
1/ and 2/ should be fixed - the cart was created with real data, the problem happened when converting it to an order

Tirodem:
Indeed, 3 is good, test data issue.

2 is also good.

And so is 1 👏

Tirodem (approved):

Issue #1613, opened 2024-12-02
💵 Add some new currencies


Tirodem:
Currencies to add :
- Shilling Kényan (KES)
- Shilling Ougandais (UGX)
- Cédi Ghanéen (GHS)
- Naiga Nigérian (NGN)
- Shilling Tanzanien (TZS)
- Dirham marocain (MAD)
- Couronne tchèque (CZK)

PR #1618, opened 2024-12-03
 💵 Add some new currencies #1613

ithiame:


coyotte508 (approved):

Tirodem:
I have some huge doubts about results : ❌ 

image
image

image
image

image
image

image
image

image
image

image
image

image
image

For previous currencies, we're more or less OK (marging related to our exchange rate API) :
image
image

image
image

coyotte508:
ah yes @ithiame need to make sure this is valid:

```

			const resp = await fetch('https://api.coinbase.com/v2/exchange-rates?currency=BTC', {
				...{ autoSelectFamily: true }
			} as unknown as RequestInit);
```

from `currency-lock`

ithiame:
I have good results on my test

Config currencies
image

pricetag
image

image

image

image

coyotte508:
you can both check `exchangeRate` in `RuntimeConfig` - see if it's updated

ithiame:
image

coyotte508:
Seems good!

Maybe @Tirodem there are multiple processed on the same DB and there was a lock in place, so the currencies didn't update?

(otherwise, deploy again & wait a few minutes. Or run locally)

Tirodem:
I see the records on runtimeConfig, I gonna deploy it again.
Are the datas from src/lib/stores/exchangeRate.ts used at the beginning before we request the exchange API for fresh rates and stock them in runtimeConfig ?
What's the frequency of this call ?

Tirodem:
Amounts are good now.

Issue #1621, opened 2024-12-04
🦋 Allow admin to set a QR-code-less order on their be-BOP


Tirodem:
Context :
Belgian customers are used to flash a payment QR code with their bank app to pay online.
It's highly confusing with current be-BOP order page (where QR code is the order page url).
We need to allow be-BOP admin to set another display behavior of the order page to remove the QR code.

Current behavior :
![image](https://github.com/user-attachments/assets/8cf3ebea-5dcd-488d-afe9-69fb0c472467)

Alternative behavior :
![image](https://github.com/user-attachments/assets/9c817b3b-bb2e-47fd-ae42-359b5e8445e6)
- remove QR code
- allow to set a custom fill color for the Sum Up / Stripe SVG
- change text in span

Admin UI :
- on /admin/config
![image](https://github.com/user-attachments/assets/34fc7c60-fb8d-4da1-b5c1-c086763df114)
![image](https://github.com/user-attachments/assets/d64175ba-9d21-4ebd-afe2-7079bf80e364)
- on /admin/theme/ (on a edit or new theme form)
![image](https://github.com/user-attachments/assets/29d30208-3bf5-415c-b868-b998d4c04760)

Note :
This shouldn't, indeed, impact the PoS page

PR #1624, opened 2024-12-04
 🦋 Allow admin to set a QR-code-less order on their be-BOP #1621

ithiame:


src/routes/(app)/order/[id]/+page.svelte
```diff
@@ -294,6 +296,13 @@
 									/></a
 								>
 							{/if}
+							{#if payment.method === 'card'}
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/+page.svelte:299]:
```suggestion
							{#if payment.method === 'card' && !data.hideCbQrCode}
```

no need to render it in the DOM at all

src/lib/server/runtime-config.ts
```diff
@@ -181,6 +181,8 @@ const baseConfig = {
 	],
 	contactModes: ['email', 'nostr'],
 	posTouchTag: [] as Tag['_id'][],
+	hideCbQrCode: false,
+	overwriteCbSvgColor: false,
```
coyotte508, line 185: `cb` is a french abbrevation. Use `creditCard` instead

coyotte508 (approved):

Tirodem:
We still need to remove this text when we are QR code less :
image
And I don't find it in translation key.

Tirodem:
Also, order-creditCard-svg should be used for Stripe too.

Tirodem:
Otherwise, it's good.

Tirodem:
https://github.com/B2Bitcoin/beBOP/issues/1632 for stripe logo

Issue #410, opened 2023-08-27
🐛 Issue on subscription listing when no npub is filled


Tirodem:
I did 4 test subscription products on https://github.com/B2Bitcoin/B2BitcoinBootik/commit/268880fcc700d1b540bc57d7e67f20d4f48db813 commit :
- gold : https://dev-bootik2.pvh-labs.ch/product/gold-subscription
- silver : https://dev-bootik2.pvh-labs.ch/product/silver-subscription
- bronze : https://dev-bootik2.pvh-labs.ch/product/bronze-subscription
- platinum / BTC : https://dev-bootik2.pvh-labs.ch/product/platinum-btc-subscription
I also did 4 orders with these customer's contact information :
- https://dev-bootik2.pvh-labs.ch/order/0f376afb-3586-4afa-bd03-05750f843ed0 / npub15jqgr8thxgsfnhxufcu05ucja8kgr0guu4ud82cm2h36ml6ulmns25pxvw
- https://dev-bootik2.pvh-labs.ch/order/2711e0f7-f7e3-4ab5-a02f-900631d6a945 / tirodem@gmail.com
- https://dev-bootik2.pvh-labs.ch/order/a4c764ad-8dd5-4c05-b844-e031832fee1d / pvhlabs@gmail.com
- https://dev-bootik2.pvh-labs.ch/order/7b205f18-d173-4e29-b466-e015d218b6c9 / christophe.gerard@tirodem.com
Here's the result on /admin/product/{id}/subscribers page :
- gold subscription
![image](https://github.com/B2Bitcoin/B2BitcoinBootik/assets/50206014/472ccf22-d54e-4fcf-9963-e1ae55e16f87)
- silver subscription
![image](https://github.com/B2Bitcoin/B2BitcoinBootik/assets/50206014/8d766633-4e65-4ac6-847d-5ede2b7da8aa)
- bronze subscription
![image](https://github.com/B2Bitcoin/B2BitcoinBootik/assets/50206014/7258ca93-c9d8-473e-838a-ebeb94c7bfaa)
- platinum / BTC subscription
![image](https://github.com/B2Bitcoin/B2BitcoinBootik/assets/50206014/9697d48f-43df-4914-a1b2-ccc131f7f130)
In database :
- the 4 subscriptions in 1 cart are saved with npub info
[...]

PR #1638, opened 2024-12-10
 🐛 Issue on subscription listing when no npub is filled #410

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts
```diff
@@ -16,6 +16,10 @@ export const load = async ({ params }) => {
 		product: pojo(product),
 		subscriptions: subscriptions.map((subscription) => ({
 			...subscription,
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts:18]:
```suggestion
			...pojo(subscription),
```

this would have worked too

coyotte508 (approved):

Issue #1548, opened 2024-10-28
🐛 When I make mistake on my contact form on /checkout, the validate button is disabled until refresh


Tirodem:
If I submit a /checkout with a from format npub (happens with auto-filled address) :
![image](https://github.com/user-attachments/assets/b4b6e916-67c7-4da2-8f06-d55559ccc724)
The CTA will stay disabled even if I fix the form entry :
![image](https://github.com/user-attachments/assets/486e5544-0bb5-483a-b9cd-8eef74bd9991)
So I'm obliged to refresh the page and lose my informations.
That should now be the case.

PR #1641, opened 2024-12-11
 🐛 When I make mistake on my contact form on /checkout, the validate …

ithiame:
…button is disabled until refresh #1548

### DIFF
--- a/src/routes/(app)/checkout/+page.svelte
+++ b/src/routes/(app)/checkout/+page.svelte
@@ -91,6 +91,7 @@
 				input.reportValidity();
 
 				event.preventDefault();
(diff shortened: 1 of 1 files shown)


src/routes/(app)/checkout/+page.svelte
```diff
@@ -91,6 +91,7 @@
 				input.reportValidity();
 
 				event.preventDefault();
+				submitting = false;
```
coyotte508, line 94: put the whole `for` inside a `try` block, and add a `finally` block afterwards with `submitting = false`

coyotte508 (approved):

Issue #1591, opened 2024-11-25
🐛 LN orders with phoenixd don't seem to expire


Tirodem:
![image](https://github.com/user-attachments/assets/7750a628-e42d-4cb1-845c-83c8321391df)

From what I see, it works with every other payment mean (even BTC on-chain).
On pilots, orders don't expire when paid with phoenixd.
(sadly we don't have phoenixd sandbox)

PR #1642, opened 2024-12-11
 🐛 LN orders with phoenixd don't seem to expire #1591

ithiame:


coyotte508 (approved):
ah maybe phoenixd's api changed

coyotte508:
so I don't understand why the `.optional()` instead of `.nullable()`? Does the response change for expired invoices

Tirodem:
@ithiame no answer was given following Coyotte's last comment, can you take a few minutes here ?

ithiame:
With nullable we have this issue in the zod

```
ZodError: [
  {
    "code": "invalid_type",
    "expected": "number",
    "received": "undefined",
    "path": [
      "completedAt"
    ],
    "message": "Required"
  }
]
```

because completed is undefined when the order is pending

ithiame:
The response is not the same for pending invoice

coyotte508:
ok

Issue #1016, opened 2024-02-19
🐛 Manage alias duplicate error in front


Tirodem:
Actually, on /admin/product/alias, if two products or more have the same non-empty alias, beBOP goes 404.
- [ ] I'd like this to be checked before save and displayed on front

If I put an existing alias on /admin/product/{id}, the page goes 500 on save
- [ ] I'd like this to be checked before save and displayed on front

(Alias are needed to be unique as they must be used for alias addition in cart)

AmadouAgiltoo:
Test OK (test workflow)

PR #1644, opened 2024-12-12
🐛 Manage alias duplicate error in front #1016

ithiame:


AmadouAgiltoo:
Test OK (test workflow)

coyotte508 (commented):
IMO we should use `use:enhance` with the form and forward the error (without erasing all the data) in case of error

Instead of fetching the list of all aliases and passing it to the frtonend

src/routes/(app)/admin[[hash=admin_hash]]/product/alias/+page.svelte
```diff
@@ -1,11 +1,32 @@
 <script lang="ts">
+	import { enhance } from '$app/forms';
+
 	export let data;
+	let listAliases = data.products.flatMap((product) => product.alias?.[1]);
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/alias/+page.svelte:5]: This can be removed no?

src/routes/(app)/admin[[hash=admin_hash]]/product/alias/+page.svelte
```diff
@@ -1,11 +1,32 @@
...
+<form
+	class="flex flex-col gap-2"
+	method="post"
+	use:enhance={() => {
+		errorDuplicateAlias = false;
+		return async ({ result }) => {
+			loading = false;
+
+			if (result.type === 'error') {
+				errorDuplicateAlias = true;
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/alias/+page.svelte:22]: I think we should just display the error message from the backend (and make sure the backend shows a readable error message)

src/lib/components/ProductForm.svelte
```diff
@@ -168,6 +170,11 @@
 				// rerun all `load` functions, following the successful update
 				await invalidateAll();
 			}
+			if (result.type === 'error') {
+				errorMessage =
+					result.error.message === 'Internal Error' ? 'Duplicate alias' : 'Error on submission';
```
coyotte508, line 175: The error message should be forwarded as is, it's up to the backend to  produce a readable message.

There is a `isUniqueConstraintError` helper, you can do a try/catch with in the catch something like

```ts
if (isUniqueContraintError(err)) {
  throw error(400, 'A product with the same alias already exists');
} else {
	throw err;
}
```

Maybe you can even check the key with the unique constraint from the contents of `err` (but not required)

coyotte508 (approved):

AmadouAgiltoo:
It's partially OK.
When an alias already exists for a product line,
if the same alias is entered on another line, the message appears correctly, and the update button is disabled.
However, when the user clears their input to enter something else, sometimes the message remains displayed, and the button stays disabled, forcing them to reload the page.

AmadouAgiltoo:
The case is not handled either (e.g., pe and Pe).

AmadouAgiltoo:
PR retested after integrating back-end checks and sending the message from the front-end. It's OK now. The case sensitivity issue will be addressed in another ticket.

AmadouAgiltoo:
Done !

PR #1646, opened 2024-12-12
fix discount on receipt

ithiame:
https://discord.com/channels/1022592041709998080/1316717098357100575


coyotte508 (approved):
will fix for the exact case, still worth it to investigate further for orders with discounts that will have a wrong discount amount.

Still not sure why it happened, it's stranger that the order amount is so different from the price stored for the products. Did the products price update during the order, or from when it was in the cart?

Even if that was the case, it should be refetched/stored correctly in the order

Tirodem:
❌ Still KO
Here's my cart :
image
Here's my proforma :
image
Unit price is displayed as "Total TTC" price, and "Prix Unit. HT" is Unit price divided by the product quantity.
I hereby confirm there wasn't any product change during the test.

Tirodem:
BWT it's not only when there's a discount, check this new case :
image
image
image
It's more like we're going dumb if there's a product qty > 1.
When product qty = 1 : it's good
image

ithiame:
image

image

image

image

Tirodem:
Nice 👌

Issue #1553, opened 2024-10-29
🐛 Nostr-bot triggers its welcome message when tagued on non-MP publication


Tirodem:
Following talks on Nostr about the official account, here's what I have :
![image](https://github.com/user-attachments/assets/52f92260-91a7-457a-8e5c-9415bff603d9)
I disabled the welcome message, for now it seems good, but public messages should trigger the send of a private message to the user.
Bonus question : why "hello world" and "hello you", depending of the message ?

BTW, we're able to make difference between public and private messages :
![image](https://github.com/user-attachments/assets/9d9e9e15-de67-463e-84c8-28d86555c41a)
It's interesting to keep both in log, but trigger answers only on private messages.

PR #1648, opened 2024-12-17
 🐛 Nostr-bot triggers its welcome message when tagued on non-MP publi…

ithiame:
…cation #1553

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
@@ -132,10 +132,12 @@
 		<li class="break-words">
 			{#if message.kind === 4}
 				<span title="Encrypted message">'🔐'</span>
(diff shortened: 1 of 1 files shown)


coyotte508:
I don't understand, this is a frontend change, it doesn't change the behavior of the nostrbot.

(Not sure I understood the issue)

What do we want to do exactly?

ithiame:
I understand that nostr-bot send welcom message when we tag it in a discussion, so we don't want this message to be sent anymore (disabling welcom message do the job), and to list only private message.

coyotte508:
> I understand that nostr-bot send welcome message when we tag it in a public discussion, so we don't want this message to be sent anymore 

This PR only changes the frontend, what's displayed in the admin, it doesn't change the behavior of the bot

coyotte508 (approved):

Tirodem:
Thanks for catching the mistake.
Indeed it's not front only and welcome message should be enabled but should process only for MP, not for public @tag

Issue #1631, opened 2024-12-06
🦋 Disable "pay" button as long as the sum up / stripe iframe is not loaded


Tirodem:
Before having a proper display on /order/{id}/pay :
![image](https://github.com/user-attachments/assets/5e664358-fb23-4920-b0f1-f8f3953c06b8)

Here what we can have :
![image](https://github.com/user-attachments/assets/53872880-6fd8-46bc-b99d-ffdf8e356d60)
![image](https://github.com/user-attachments/assets/3d936202-8401-4fc4-8afa-71a8ada062ee)

It would be best to disable the button "Pay", either :
- for X seconds after page display
- until the iframe is fully loaded

That will remove a bit of customer frustration when they click too fast and avoid issues with submitted empty forms.

PR #1665, opened 2025-02-07
🦋 Disable pay button as long as the sum up / stripe iframe is not loa…

ithiame:
…ded #1631


coyotte508 (changes requested):
There is not a single `await` between `stripeLoading = true` and `stripeLoading = false`

image

So it doesn't do anything.

You need to find a way to wait for a callback or something

src/routes/(app)/order/[id]/payment/[paymentId]/pay/+page.svelte
```diff
@@ -215,6 +217,10 @@
 					layout: 'tabs'
 				})
 				.mount('#payment-element');
+			// Delay to give time for UI to render
+			setTimeout(() => {
+				stripeLoading = false;
+			}, 3000);
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/payment/[paymentId]/pay/+page.svelte:223]: Ideally you can listen to the DOM instead of hardcoding a delay. There is the `loaded` event for the document, and for the `iframe` too if you can find it.

It's what we do to download invoices or bulk invoices, listen to the dom event. For example here: https://github.com/B2Bitcoin/beBOP/pull/1660/files#diff-633ad8803814b1c6ca182165aabc08d46569cba063f15d1fc09e816072366bb0R184-R193

Searching on google I also get some answers: https://stackoverflow.com/questions/52954937/when-are-stripe-elements-ready-in-the-dom https://docs.stripe.com/js/element/events

Please take the time to look for good solutions

Comment (Tirodem) [reply src/routes/(app)/order/[id]/payment/[paymentId]/pay/+page.svelte:223]: The blame's on me here.
I proposed to make a 2 step implementation because this issue costs us (PVH Editions) money with issues on card payment.
I don't mind having a 3-seconds wait asap and a smart solution after.

coyotte508 (approved):
ok for the quickfix then I guess cc @Tirodem but please make a follow-up PR :)

Tirodem:
Yes cc @AmadouAgiltoo (but I'd be more into keeping #1631 open as the 2 version were listed on it, we can do 2 checkbox to say first part is done)

AmadouAgiltoo:
OK @Tirodem I'll indicate on the 331 that the quick fix by adding some seconds before display the buttom is done.
Then after we'll work for a clean implementation.

Issue #1675, opened 2025-02-11
Admin > Products - open the product in a new tab


AmadouAgiltoo:
In the admin panel, in the list of product : /admin/product, add an icon to open a product in a new tab. Currently, the product opens in the same tab.
In the admin panel, in the global product section, add an icon to open the product in a new tab by clicking on view. Currently, the product opens in the same tab.

AmadouAgiltoo:
Test failed. The product details is not open on a new tab.

PR #1676, opened 2025-02-11
Admin > Products - open the product in a new tab #1675

ithiame:


Tirodem:
The issue told about an icon to open in a new tab, in order to have the possibility of opening in the same tab or in a new tab.
Right now there's no icon and it open in a new tab only.
I take it as it is because it's practical but respect the initial request of the issue next time.
I'll create another ticket to complete this in the future.

Issue #1667, opened 2025-02-10
transaction > reporting - track the VAT amount and its total cost


AmadouAgiltoo:
As a Be-Bop store manager, I want to be able to track the VAT amount and its total cost in the Transaction > Reporting section.
Make sure to clearly indicate if the VAT is exclude if not applicable.

![Image](https://github.com/user-attachments/assets/e2a7cca2-df0d-4f77-88de-46c35d8d2dae)

PR #1680, opened 2025-02-12
transaction > reporting - track the VAT amount and its total cost #1667

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.svelte
```diff
@@ -57,6 +59,20 @@
 		),
 		averageCart: 0
 	};
+	$: orderVATSynthesis = {
+		orderQuantity: sum(paidOrders.map((order) => order.quantityOrder)),
+		orderNumber: paidOrders.length,
+		orderVATTotal: sum(
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.svelte:65]: Use the `sumCurrencies` helper

coyotte508 (approved):
It's still not perfect in case of currency changes over the lifetime of the shop, a better solution would be more elaborate

PR #1685, opened 2025-02-27
Community request - external resource (hyperlink) as a digital file #…

ithiame:
…1684


Tirodem:
Can you add these svg as icon before download and resource links ?
- for digital files : https://www.svgrepo.com/svg/514111/download-window
- for external resources : https://www.svgrepo.com/svg/493695/link-external-new-window-open

src/lib/types/Product.ts
```diff
@@ -19,6 +19,10 @@ export interface ProductTranslatableFields {
 		href: string;
 		fallback?: boolean;
 	}[];
+	externalResource?: {
```
coyotte508, line 22:
```suggestion
	externalResources?: {
```

src/routes/(app)/order/[id]/fetchOrderForUser.ts
```diff
@@ -137,7 +137,8 @@ export async function fetchOrderForUser(orderId: string) {
 				shipping: item.product.shipping,
 				paymentMethods: item.product.paymentMethods,
 				isTicket: item.product.isTicket,
-				variationLabels: item.product.variationLabels
+				variationLabels: item.product.variationLabels,
+				externalResource: item.product.externalResource
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/fetchOrderForUser.ts:141]: This should only be fetched if the order is paid.

You can remove the field after fetching the order if the order is not fully paid.

Otherwise someone can monitor the network call and access the resource without paying

coyotte508 (approved):

PR #1705, opened 2025-03-07
📅 Create Schedule widget datamodel & administration interface #1694

ithiame:


src/lib/types/Schedule.ts
```diff
@@ -0,0 +1,36 @@
...
+	location?: {
+		name: string;
+		link: string;
+	};
+	url?: string;
+	unavailabity?: {
+		label: 'postponed' | 'canceled' | 'soldOut';
+		isUnavailable: boolean;
+	};
+	is_archived?: boolean;
```
coyotte508, line 20: do not use `snake_case` when the rest is camelCase

src/routes/(app)/admin[[hash=admin_hash]]/schedule/[id]/schedule-schema.ts
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/schedule/[id]/schedule-schema.ts:1]: This file and the two below are empty

coyotte508 (commented):
you also need to update https://github.com/be-BOP-io-SA/be-BOP/blob/e2ec7664b90a576968b7a21c1be8a9ecc45263e6/src/routes/(app)/admin%5B%5Bhash%3Dadmin_hash%5D%5D/picture/%2Bpage.server.ts#L2-L10

Tirodem:
image
End date should be optional.

Tirodem:
✔️ End date check is right

Issue #1716, opened 2025-03-11
🦋 Allow to replace a payment on a pending order


Tirodem:
On pilots, sometime, on Point of Sale experience, employees fill the wrong payment mean.
Currently, if the payment is of 100% of the order amount (no deposit), admins have to cancel the payment, and so the order, and create a new order from scratch with the correct payment mean.

So, here's the proposition : add an option to replace the current payment.
![Image](https://github.com/user-attachments/assets/5df0f1f4-7c29-4302-8c90-4ca12185cf08)
On click, it displays a form to create a new payment call with another payment mean (it was taken from the multiple payment / payment call for order with deposit) :
![Image](https://github.com/user-attachments/assets/2c537768-1b75-497d-a33d-304beb62ae2a)

For now, I propose to pre-fill the payment call with the amount of the previous payment, but we can let an option to fill a lesser value (so it'll allow to add another payment mean).
Still, keep a way to easily switch from one to another, as there won't be a checkbox with order on depositf and acknowledgement of the obligation to pay the rest : I have a doubt about law compliance here in UI.
Safest option would probably to not allow lesser substitute payment, so a full order will have to be paid with a full payment, and a deposit order will have to be paid with the same deposit amount, but it might change in the future, depending of the law insights or of the feedbacks...

When a payment is replaced :
- cancel the payment, but not the order (keep it pending), and do not send a payment cancellation notification
- create a new pending payment, send a pending payment notification

PR #1717, opened 2025-03-12
🦋 Allow to replace a payment on a pending order #1716

ithiame:


Tirodem:
We have to change the new CTA label to cancel the substitution once clicked.
image
Expected :
image
I continue testing

Tirodem:
This error must be managed on form submission to keep the employee in UI (another ticket can be created later) :
image
(As a PoS user I try to edit a non-PoS order)

Tirodem:
Global receipt seems to have an issue :
image
image

Tirodem:
Otherwise, so far, seems good

src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts
```diff
@@ -136,5 +138,66 @@ export const actions = {
 			});
 		}
 		throw redirect(303, request.headers.get('referer') || `${adminPrefix()}/order`);
+	},
+	replacePaymentMethod: async ({ params, request, locals }) => {
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:142]: The function should only be called `replace` , because there is already `payment/[paymentId]` in the url

src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts
```diff
@@ -136,5 +138,66 @@ export const actions = {
...
+		const formData = await request.formData();
+		const parsed = z
+			.object({
+				amount: z.string().regex(/^\d+(\.\d+)?$/),
+				method: z.enum(methods as [PaymentMethod, ...PaymentMethod[]])
+			})
+			.parse({
+				amount: formData.get('amount'),
+				method: formData.get('method')
+			});
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:186]: The amount shouldn't be sent from the frontend, it should be the exact same amount as the replaced payment.

src/routes/(app)/order/[id]/+page.svelte
```diff
@@ -382,6 +383,66 @@
...
+												<label class="form-label">
+													{t('order.addPayment.amount')}
+													<input
+														class="form-input"
+														type="number"
+														name="amount"
+														min="0"
+														step="any"
+														max={data.order.currencySnapshot.main.totalPrice.amount}
+														value={data.order.currencySnapshot.main.totalPrice.amount}
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/+page.svelte:416]: This is not accurate when there are muliple payments (eg POS sale)

src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts
Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:186]: As it's a Point of Sale order, most of the time it'll be the exact same amount, but it's interesting to have the possibility to replace a payment with amount X by a payment with amount Y then create others calls for payment.

Comment (coyotte508) [reply src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:186]: For point of sale then the better way would be to allow cancelling payments (without cancelling the order), and adding back new payment methods

It would be a minimal code change 

In fact this function, replace payment on a pending order, is very similar to the other one in https://github.com/be-BOP-io-SA/be-BOP/pull/1719#issue-comment-box

They could share the same code (replace a pending/failed payment), and for POS, they could just cancel payments and add arbitrary new payments that may or may not be the full amount.

What do you think?

Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:186]: Why not, but one question subsist : as a PoS seller, if cancelling a payment doesn't make an order to expire, how would I be able to cancel / expire my order manually ?

Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/order/[id]/payment/[paymentId]/+page.server.ts:186]: Oh, we can make a "Cancel payment" button, and a "Cancel order" button.
If we do that, let's go with your way.

src/routes/(app)/order/[id]/payment/[paymentId]/+page.server.ts
```diff
@@ -41,20 +43,43 @@ export const actions = {
...
+				amount
+					? {
+							amount,
+							currency: order.currencySnapshot.main.totalPrice.currency
+					  }
+					: payment.price,
+				{
+					expiresAt: paymentMethodExpiration(parsed.method),
+					session
+				}
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/payment/[paymentId]/+page.server.ts:81]: I think it should be `payment.price` like before?

POS users can just cancel a payment (without cancelling order, need a PR for that probably) and add back an arbitrary payment afterwards.

coyotte508 (approved):

coyotte508 (approved):

Issue #1595, opened 2024-11-28
🦋 Allow payment retry when CB / paypal payment is KO as long as the order is not expired


Tirodem:
Following https://github.com/B2Bitcoin/beBOP/issues/1518
On paypal / CB (sum up, stripe) payment, if 3rd party feedback is "KO", instead of expiring the order directly, allow to retry the payment mean.

Ideally, customers would want to be able to choose another payment mean, but as I worked on this on a previous job, I know devil's in the details, it's not as simple as that, and can bring a lot of problems, so it'll be on one future with more time to think about it.

At least, this ticket will allow you to try with another payment card, or change your payment card credit (for payment max amount) or else without generating a new cart & order.

PR #1719, opened 2025-03-13
🦋 Allow payment retry when CB / paypal payment is KO as long as the o…

ithiame:
…rder is not expired #1595


ithiame:
image

If the payment fail i allow customer to add new paiement and retry

src/lib/server/orders.ts
```diff
@@ -345,7 +345,10 @@ export async function onOrderPayment(
 export async function onOrderPaymentFailed(
 	order: Order,
 	payment: Order['payments'][0],
-	reason: Exclude<OrderPaymentStatus, Exclude<OrderPaymentStatus, 'canceled' | 'expired'>>
+	reason: Exclude<
+		OrderPaymentStatus,
+		Exclude<OrderPaymentStatus, 'canceled' | 'expired' | 'failed'>
+	>
```
coyotte508, line 351: Actually you can use `Extract`


```suggestion
	reason: Extract<OrderPaymentStatus, 'canceled' | 'expired' | 'failed'>
```

src/routes/(app)/order/[id]/+page.svelte
```diff
@@ -508,6 +511,46 @@
...
+							<select name="currency" class="form-input" disabled>
+								<option value={data.order.currencySnapshot.main.totalPrice.currency}
+									>{data.order.currencySnapshot.main.totalPrice.currency}</option
+								>
+							</select>
+						</label>
+						<label class="form-label">
+							<span>{t('checkout.payment.method')}</span>
+							<select name="method" class="form-input">
+								{#each data.paymentMethods.filter( (pm) => ['card', 'bitcoin', 'lightning', 'paypal'].includes(pm) ) as paymentMethod}
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/+page.svelte:541]: This should be computed up top in the `script` tag. Because we don't want to show the help text or this block if the payment method list is empty

src/routes/(app)/order/[id]/+page.server.ts
```diff
@@ -97,5 +101,67 @@ export const actions = {
...
+				method: formData.get('method')
+			});
+
+		const amount = parsePriceAmount(parsed.amount, order.currencySnapshot.main.totalPrice.currency);
+		if (amount <= 0) {
+			throw error(400, 'Invalid amount');
+		}
+		if (amount > remainingAmount) {
+			throw error(400, 'Amount is greater than the remaining amount to pay');
+		}
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/+page.server.ts:153]: So the user can freely choose the amount of payment between `1` and `remaningAmount`?

It should match exactly the failed payment, and probably reference the failed payment in some way too

We only allow multiple payment methods (with a free choice amount) for POS role, and here this situation is only for customer role (from the frontend code)

src/routes/(app)/order/[id]/+page.server.ts
```diff
@@ -97,5 +101,67 @@ export const actions = {
...
+
+		await addOrderPayment(
+			order,
+			parsed.method,
+			{
+				amount,
+				currency: order.currencySnapshot.main.totalPrice.currency
+			},
+			{ expiresAt: null }
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/+page.server.ts:162]: `expiresAt` should be similar to here: https://github.com/be-BOP-io-SA/be-BOP/blob/66eee4d68ed507b8f5a13ff13f62256191493fb2/src/lib/server/orders.ts#L996-L998 ?

(instead of `null`)

coyotte508 (commented):
With the current code, if a payment fails, and the user does not do anything, the order will never expire

src/routes/(app)/order/[id]/+page.server.ts
```diff
@@ -97,5 +100,45 @@ export const actions = {
...
+			.object({
+				method: z.enum(methods as [PaymentMethod, ...PaymentMethod[]])
+			})
+			.parse({
+				method: formData.get('method')
+			});
+
+		await addOrderPayment(order, parsed.method, order.currencySnapshot.main.totalPrice, {
+			expiresAt: addMinutes(new Date(), runtimeConfig.desiredPaymentTimeout)
+		});
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/+page.server.ts:140]: For `expiresAt`, to be consistent, wse should use `paymentMethodExpiration` as in https://github.com/be-BOP-io-SA/be-BOP/blob/fc5bceb9b3b01e0d4fb117c0f58ee764ab83be4b/src/lib/server/orders.ts#L995-L998

For the amount, it should be the same as the cancelled payment. 

In fact it should be a replace payment call, like in https://github.com/be-BOP-io-SA/be-BOP/pull/1717, so the url should be something like `order/[id]/payment/[paymentId]/+page.server.ts`

And the user needs to provide the id of the failed payment.

The reason we can't just put the full amount, is that for example there are orders with deposits.

ithiame:
The form is moved here 
image

Tirodem:
:x: Still no order expiration despite the good value in database :
https://dev-bootik2.pvh-labs.ch/order/5248f4a0-22dd-4e6a-b861-5bb990a0096d
image

src/routes/(app)/order/[id]/payment/[paymentId]/+page.server.ts
```diff
@@ -0,0 +1,56 @@
...
+			throw error(400, 'No payment methods available');
+		}
+		const formData = await request.formData();
+		const parsed = z
+			.object({
+				method: z.enum(methods as [PaymentMethod, ...PaymentMethod[]])
+			})
+			.parse({
+				method: formData.get('method')
+			});
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/payment/[paymentId]/+page.server.ts:50]: This should be done before the `onOrderPaymentFailed` - to not leave the order in a state where payment is cancelled but nothign replaces it

coyotte508 (commented):
From what I see, if the payment fails (not "expires"), and the user doesn't replace the payment method, the order will never expire.

(Feel free to correct me)

ithiame:
> From what I see, if the payment fails (not "expires"), and the user doesn't replace the payment method, the order will never expire.
> 
> (Feel free to correct me)

Yes you're right

src/routes/(app)/order/[id]/payment/[paymentId]/+page.server.ts
```diff
@@ -0,0 +1,57 @@
...
+
+		await addOrderPayment(order, parsed.method, payment.price, {
+			expiresAt: paymentMethodExpiration(parsed.method)
+		});
+		await onOrderPaymentFailed(order, payment, 'canceled', true);
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/payment/[paymentId]/+page.server.ts:53]: They should be in the same session. `addOrderPayment` already takes a session param, `onOrderPaymentFailed` should also

coyotte508 (commented):
Looks good to me (after comment is handled)

With the caveat that orders with payment "failed" do not expire. I don't know if it's ok for you @Tirodem

coyotte508 (dismissed):

Tirodem:
The order has to expire otherwise the stock can be squatted.

ithiame:
> Looks good to me (after comment is handled)
> 
> With the caveat that orders with payment "failed" do not expire. I don't know if it's ok for you @Tirodem

Order with paiement "failed" will expire after expiration date or be canceled if there's a new paiement call

src/lib/server/locks/order-lock.ts
```diff
@@ -232,6 +239,8 @@ async function maintainOrders() {
 										order = await onOrderPaymentFailed(order, payment, 'failed');
 									} else if (checkout.status === 'EXPIRED') {
 										order = await onOrderPaymentFailed(order, payment, 'expired');
+									} else if (payment.expiresAt && payment.expiresAt < new Date()) {
+										order = await onOrderPaymentFailed(order, payment, 'expired');
```
coyotte508: Can remove IMO

src/lib/server/locks/order-lock.ts
```diff
@@ -29,20 +29,27 @@ async function maintainOrders() {
...
+				(p) => p.status === 'pending' || p.status === 'failed'
+			)) {
 				// Since we can overwrite order, we need to update payment too if needed
 				const updatedPayment = order.payments.find((p) => p._id.equals(payment._id));
-				if (!updatedPayment || updatedPayment.status !== 'pending') {
+				if (
+					!updatedPayment ||
+					(updatedPayment.status !== 'pending' && updatedPayment.status !== 'failed')
+				) {
 					continue;
```
coyotte508, line 53: Just do a separate loop for failed payments

```ts
for (let payment of order.payments.filter(p => p.status === "failed") {
  const updatedPayment = ...;
  if (updatedPayment.expiresAt && ...) {
		order = await onOrderPaymentFailed(...);
  }
}

for (let  payment of order.payments.filter(p => p.status === "pending") {
  ...
}
```

it's clearer that way

src/lib/server/locks/order-lock.ts
```diff
@@ -29,9 +29,11 @@ async function maintainOrders() {
 			await setTimeout(5_000);
 			continue;
 		}
-
 		const pendingOrders = await collections.orders
-			.find({ 'payments.status': 'pending', status: 'pending' })
+			.find({
+				'payments.status': { $in: ['pending', 'failed'] },
+				status: { $in: ['pending', 'failed'] }
```
coyotte508, line 35: Is there a need for this line? From what I see, `order.status` is never set to "failed", only to "canceled" or "expired"

ithiame: You're right

coyotte508 (approved):

Tirodem:
It's good now ✔️ 
We need another email template for an expiring order that doesn't contain any pending payment (like I make a 1st CB payment attempt, it goes expired / KO, I don't set new payment attempt, then the order expires).

Issue #1731, opened 2025-03-17
🦋 Allow email / npub collection even if no SMTP / nsec info is setted


Tirodem:
Currently, on /checkout :

- npub input is displayed if it's setted in admin > config + if this info is filled in .env.local :
```
NOSTR_PRIVATE_KEY=
```

- email is displayed if it's setted in admin > config + if these infos are filled in .env.local :
```
SMTP_FAKE=
SMTP_HOST=
SMTP_PORT=
SMTP_USER=
SMTP_PASSWORD=
```

![Image](https://github.com/user-attachments/assets/78f96ead-228d-4d10-8a64-0a953aaf5964)

All of this because we didn't plan to collect email or npub if be-BOP isn't able to contact the addresses back.
So, if be-BOP cannot send email, we don't collect email, and if be-BOP cannot send nostr DM, we don't collect npub.

Actually, some pilots want to do it, and TBH, I don't see why not.

So, on Admin > Config, add this option :

![Image](https://github.com/user-attachments/assets/2b1cbae6-7827-45f8-b203-21ff19f0564c)

So, logical decision might be something like :
- `address type is set in Contact Modes in Admin > Config > Notifications` AND (`address type is set in .env.local` OR `Force option display is check in Admin > Config > Notifications`)

This is useful for pilots with no smtp access (for example, Protonmail doesn't offer smtp services on free plan, pilots don't want to use GMail, pilots don't have smtp services with hosting and smtp is hard to have in a full self-hosted configuration) or with pilots who want to make changes without being able to connect to their VPS through SSH.

Tirodem:
Best thing would also be, in case email (for example) are collected through "Force display" and with no SMTP configuration, to not try to send an email, to avoid spam or errors in log or flood of old emails when setting correct smtp parameters in .env.local.
And also same for npub collection.

ithiame:
> Best thing would also be, in case email (for example) are collected through "Force display" and with no SMTP configuration, to not try to send an email, to avoid spam or errors in log or flood of old emails when setting correct smtp parameters in .env.local. And also same for npub collection.

We save emails to a database. There's also an async export function, queueEmail(), that we use often.
Perhaps a @coyotte508  opinion before we fix it.

PR #1735, opened 2025-03-17
🦋 Allow email / npub collection even if no SMTP / nsec info is setted…

ithiame:
… #1731


Tirodem:
image
image
:x: The back seems to require a valid notification target (here on this pilot, smtp is not configured, nsec is)

Tirodem:
✔️ The 400 error is no more around

Issue #1753, opened 2025-03-24
📆 Allow to manage event color on calendar display and hide from event list option on Schedule widget


Tirodem:
Currently, on Schedule widget with `display=calendar`, having an event put an orange date on calendar :

![Image](https://github.com/user-attachments/assets/1d17588c-5430-4798-8d64-0676e1043275)

Combined with #1752, we'll need :
- a checkbox "Hide event from list" to show an event only in calendar view of widget, and not in an event list
- an "Event color on calendar" input to choose an event color on calendar view of schedule widget (hexadecimal + color selector, like in theme manager)

That'll allow to display unavailable dates, open dates, and also special events, like that :

![Image](https://github.com/user-attachments/assets/596def66-a2d2-4489-9d89-cf7a08b7d030)

Also, on Schedule widget global configuration, allow to choose color for date selector.

PR #1758, opened 2025-03-25
📆 Allow to manage event color on calendar display and hide from event…

ithiame:
… list option on Schedule widget #1753

### DIFF
--- a/src/lib/types/Schedule.ts
+++ b/src/lib/types/Schedule.ts
@@ -13,6 +13,8 @@ export interface EventSchedule {
 		link: string;
 	};
 	url?: string;
+	hideFromList?: boolean;
+	colorOnCalendar?: string;
 	unavailabity?: {
 		label: string;
 		isUnavailable: boolean;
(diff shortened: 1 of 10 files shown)


Tirodem:
✔️ Tests are good for me

coyotte508 (approved):

src/lib/types/Schedule.ts
```diff
@@ -13,6 +13,11 @@ export interface EventSchedule {
 		link: string;
 	};
 	url?: string;
+	hideFromList?: boolean;
+	calendar?: {
+		hasCustomColor: boolean;
+		color: string;
```
coyotte508, line 19: tbh we should never need the `hasXxx` variables in DB, only make the `color?: ` variable optional here for example.

This comes up a lot

Tirodem:
❌ one of my test schedule goes to /error on cms page when I display it without ?display=calendar (more infos on Discord)

Tirodem:
✔️ Unrelated to this PR

PR #1765, opened 2025-03-28
Improve cms page slug

ithiame:
>  Allow slug edition on creation but forbid uppercases, spaces and special characters in input
>     BTW, pages already created with uppercases and spaces previously to this PR should still work

### DIFF
--- a/src/lib/components/CmsForm.svelte
+++ b/src/lib/components/CmsForm.svelte
@@ -59,7 +59,7 @@
 	function validateSlug(event: SubmitEvent) {
 		const value = slugElement.value;
 		const result = slugSchema.safeParse(value);
-		if (!result.success) {
+		if (!result.success && !cmsPage?._id) {
 			slugElement.setCustomValidity(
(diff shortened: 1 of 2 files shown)


Tirodem:
✔️ Tests are OK
Except for these :
image
As seen, we'll make them in a 2nd step.
Can we unlink this PR to the original issue ? We'll be able to merge as it is once validated and we'll keep track for phase 2 with slug edition.

src/lib/components/CmsForm.svelte
```diff
@@ -6,6 +6,8 @@
 		TINYMCE_TOOLBAR
 	} from '../../routes/(app)/admin[[hash=admin_hash]]/cms/tinymce-plugins';
 	import { MAX_CONTENT_LIMIT } from '$lib/types/CmsPage';
+	import { generateId } from '$lib/utils/generateId';
+	import { z } from 'zod';
```
coyotte508, line 10: Let's not add this dependency on the frontend (and increase bundle size), it's only used in the backend

Issue #1592, opened 2024-11-25
🦋 Pre-fill a default price for a PWYW product


Tirodem:
TBD
product PWYW
currently, what is pre-filled as price on product page is the minimum amount
it could be interesting to pre-fill a price **that is not the minimum amount** and that can be changed by the customer
design to come

PR #1776, opened 2025-04-01
🦋 Pre-fill a default price for a PWYW product #1592

ithiame:


coyotte508 (approved):

Tirodem:
❌ Big regression
A product PWYW with minimal amount and no pre-fill amount goes like this :
image
In that case we should apply the minimal amount and not the non-setted pre-fill amount.
Now, the minimal option isn't optional anymore :
image
If we want to go like that we should add a sub-option "This article has a pre-filled price"

coyotte508 (approved):

PR #1780, opened 2025-04-02
A schedule with event without any image goes to error on CMS page #1779

ithiame:


Tirodem:
✔️ Tests OK, merge if PR review is good

coyotte508 (commented):
you should fix the typing at the source

For example:

```
export let pictures: Picture[] | [];
```

why does it not throw an error if `pictures` can be undefined? There's a wrong typing somewhere upstream.

(also you can do `export let pictures: Picture[] = [];`  maybe, but the typing should still be fixed at the source)

coyotte508 (approved):
Ok the root cause is from lodash's `groupBy` which returns a `Record<...>` instead of  a `Partial<Record<...>>` 

We should really remove lodash from the frontend btw

Issue #1781, opened 2025-04-02
🦋 Display html block on CMS admin page only if user enables a checkbox


Tirodem:
To lighten the page a bit

![Image](https://github.com/user-attachments/assets/39ad2ed2-4537-4b1f-ae01-0076955892f8)

![Image](https://github.com/user-attachments/assets/64d424d6-885d-416f-b093-dc70626e8bb6)

Do it also on Product admin page CMS zone

PR #1783, opened 2025-04-03
🦋 Display html block on CMS admin page only if user enables a checkbo…

ithiame:
…x #1781


coyotte508 (approved):

Tirodem:
❌ big no
no raw html displayed then save of the product = no saving + 422 error "content: Required"

coyotte508 (approved):

PR #1791, opened 2025-04-07
🦋 In admin CMS page, make a "Show tips" option to display or not all …

ithiame:
…the "To include" advices #1784

### DIFF
--- a/src/lib/components/CmsForm.svelte
+++ b/src/lib/components/CmsForm.svelte
@@ -40,6 +40,7 @@
 	let employeeContent = cmsPage?.employeeContent || '';
 	let slugElement: HTMLInputElement;
 	let formElement: HTMLFormElement;
+	let showTips = false;
 
 	function confirmDelete(event: Event) {
 		if (!confirm('Would you like to delete this CMS page?')) {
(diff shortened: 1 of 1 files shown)


Tirodem:
✔️ Tests OK, merge if PR good

src/lib/components/CmsForm.svelte
```diff
@@ -185,82 +186,89 @@
...
-				option like this:
-				<code class="font-mono">[Gallery=slug?display=var-1]</code>
-			</li>
-		</ul>
+		<label class="checkbox-label my-2">
+			<input type="checkbox" name="showTips" bind:checked={showTips} class="form-checkbox" />
+			Show tips
+		</label>
+		{#if showTips}
+			<ul class="text-gray-700 my-3 list-disc ml-4">
```
coyotte508, line 194: you can do it in CSS-only:

```suggestion
		<label class="checkbox-label my-2 peer">
			<input type="checkbox" name="showTips" bind:checked={showTips} class="form-checkbox" />
			Show tips
		</label>
		<ul class="text-gray-700 my-3 list-disc ml-4 hidden peer-checked:block">
```

With no need for an extra var (and it works without JS in the client)

coyotte508 (approved):

src/lib/components/CmsForm.svelte
```diff
@@ -186,81 +186,88 @@
...
-				>, where
-				<code class="font-mono">slug</code> is the slug of your gallery. You can specify the display
-				option like this:
-				<code class="font-mono">[Gallery=slug?display=var-1]</code>
-			</li>
-		</ul>
+		<label class="checkbox-label my-2 block">
+			<input type="checkbox" name="showTips" class="form-checkbox peer" />
+			Show tips
+			<ul class="text-gray-700 my-3 list-disc ml-4 hidden peer-checked:block">
```
coyotte508, line 192: Don't put everything in the `<label>`.

If `peer-has-checked` does not work (maybe we need to upgrade tailwind) go back to the previous version

PR #1795, opened 2025-04-07
Fix discount

coyotte508:
image

Still:

- needs to display stricken price on product page?
- need to account in cart page & order page & receipts


coyotte508:
>   can we make something like "if the price is too long, only display final price" ?

In a follow up PR

coyotte508:
Note @ithiame 

This PR adds a `discountPercentage` in the `Order["items"]` schema. The `Order["items"]["currencySnapshot"]` stays as before, the full price of a single product.  So to display the discounted price for a product, it needs to be multiplied by the discount percentage, like in the receipt page, order summary, reporting...

I updated everywhere I could think of (including the reporting tab)

Tirodem:
KO when Discount = 100

image

image

image

coyotte508:
hmm can't reproduce locally 

image

Two deploys on `bootik2` at 2am separated by 15 minutes, maybe it was the frontend of thsi PR and the back of another PR? (idk)

Will deploy on bootik2 and see if I can reproduce there

coyotte508:
hmm note it's only 90% off, maybe the discount was updated when you had the checkout page already loaded and the error happened because the discount wasn't 100% anymore?

image

It worked for me: https://dev-bootik2.pvh-labs.ch/order/1bf927ae-ca92-479e-9e52-1542e58dcacd

Tirodem:
Ok, strange, I'll recheck it.
BTW, I spotted that if you visit the order and check the receipt while not being logged, we display the product full price :
image
image

We should make something in order to keep a snapshot of the product price at the time and session context of the order, otherwise it might be bothersome.

Tell me if it can be done here of if it deserves a new separate ticket.

coyotte508:
> Ok, strange, I'll recheck it. BTW, I spotted that if you visit the order and check the receipt while not being logged, we display the product full price : image image
> 
> We should make something in order to keep a snapshot of the product price at the time and session context of the order, otherwise it might be bothersome.
> 
> Tell me if it can be done here of if it deserves a new separate ticket.

Did you check the receipt while on the PR or different code?

It should display the same for everyone

Maybe we can just merge this PR and open tickets if issues remain?

Tirodem:
Indeed, it's good. Might have been a branch / deploy mistake of mine.

Tirodem (approved):

Tirodem:
https://github.com/be-BOP-io-SA/be-BOP/issues/1806
https://github.com/be-BOP-io-SA/be-BOP/issues/1807
https://github.com/be-BOP-io-SA/be-BOP/issues/1808

Issue #1778, opened 2025-04-02
📅 Add RSVP option to event


Tirodem:
# Overview

This option will add a button to an event that'll allow a user to notify his / her participation to the event.

For now, the RSVP should be done through email or nostr notification.

# Design

## Admin

![Image](https://github.com/user-attachments/assets/789c0902-6c04-428c-be6b-c4f406a2ae67)

![Image](https://github.com/user-attachments/assets/ab8d1037-0290-46a0-8085-7b5be44054e4)

## Visitor

Clic on icon should bring this :
[no screenshot yet, do this in a way similar to shedule subscription)

### Original view

![Image](https://github.com/user-attachments/assets/980f3b84-0133-4953-8633-290173f4f185)

### Light view

![Image](https://github.com/user-attachments/assets/4ed066f7-2e8a-470f-90b9-a2df44f00c7f)

### List view

![Image](https://github.com/user-attachments/assets/34b3ddcd-3516-4976-b9c6-8c5215fbbb0b)
Icon : 🙋

### Mobile view

![Image](https://github.com/user-attachments/assets/9d908060-bc0d-483c-8c2a-993d0946547f)

## Notification

For email, add this in Admin > Config > Template.
For nostr, hardcoded for now, will be templated in the future.

### For admin

```
{{brandName}} - RSVP for {{eventName}}
-----
<p>A new person confirmed participation to this event : {{eventName}}</p>
<p>{{participantContact}}</p>
```

`{{participantContact}}` is the email / npub used by the visitor while he/she RSVP'd.

### For user

```
{{brandName}} - Participation confirmation to {{eventName}}
-----
<p>This message was sent to you because you confirmed your participation to this event : {{eventName}} by {{brandName}}
<p></p>
<p><img src="{{eventPicture}}" style="max-width: 500px"></p>
<p>{{eventName}} - {{eventDate}}</p>
<p>{{eventShortDescription}}</p>
<p>{{eventDescription}}</p>
<p>{{eventLocationName}}</p>
<p><a href="{{eventLocationLink}}">{{eventLocationLink}}</a></p>
<p></p>
<p>If this an error or if you don't want to participate anymore to the event, please notify the organizer through this link :</p>
<p><a href="{{eventCancellationLink}}">Cancel participation</a></p>
```

`{{eventCancellationLink}}` will be a link setted with a user contact (email / npub) and event.

# Cancellation link

PR #1797, opened 2025-04-08
📅 Add RSVP option to event #1778

ithiame:


src/routes/(app)/schedule/[id]/rsvp/[slug]/+page.server.ts
```diff
@@ -0,0 +1,90 @@
...
+		);
+		if (address.includes('@')) {
+			await queueEmail(address || SMTP_USER, 'schedule.rsvp.user', {
+				brandName: runtimeConfig.brandName,
+				eventName: eventSchedule.title,
+				eventDate: format(eventSchedule.beginsAt, 'EEEE dd MMM yyyy HH:mm'),
+				eventShortDescription: eventSchedule.shortDescription,
+				eventDescription: eventSchedule.description,
+				eventLocationName: eventSchedule.location?.name,
+				eventLocationLink: eventSchedule.location?.link,
```
Comment (coyotte508) [inline src/routes/(app)/schedule/[id]/rsvp/[slug]/+page.server.ts:65]: note: we should probably html escape this properly but it's already not handled correctly. Would need a separate issue to do a pass and see if it can be fixed at the source (the templating engine)

src/routes/(app)/schedule/[id]/rsvp/[slug]/+page.server.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const actions = {
+	notify: async function ({ request, params }) {
+		const schedule = await collections.schedules.findOne({ _id: params.id });
+		if (!schedule) {
+			throw error(404, 'Schedule not found');
+		}
+		const eventSchedule = schedule.events.find((e) => e.slug === params.slug);
+		if (!eventSchedule) {
+			throw error(404, 'Event not found');
```
Comment (coyotte508) [inline src/routes/(app)/schedule/[id]/rsvp/[slug]/+page.server.ts:36]: Note that anything that sends emails should call `rateLimit`

coyotte508 (approved):

Tirodem:
❌ Opening my older schedules in admin goes to /error.

Tirodem:
- :x: `{{eventCancellationLink}}` is still no good
- :x: admin notification is sent to be-BOP admin email and not to target npub
Otherwise it's OK.

Tirodem:
✅ Seems good.
Fix conflict, check PR (last review) and we can go !

Issue #1796, opened 2025-04-08
🆕 Challenge with ratio


Tirodem:
# Overall

When you make a crowdfunding on platform, people believe you have all of the money.

Content creators have to justify to their supporters :
- they pay taxes
- they pay furnitures
- they pay copyrights
- etc

Idea here is to allow to make that directly, in a dynamic way, through Challenge gauge.

So, 3 use-cases :
- Classic (current mode, 100%)
- Global ratio (set 10% as global ratio, and only 10% of the global value of sold products will be added to the Challenge current score)
- Per product ratio (same but 10% for product A, 15% for product B, 20% for product C)

In the end, what's incrementing the Challenge score is product price * ratio (global or per product) / 100, otherwise (with Total), it's like ratio = 100.

# Design

All is done in Challenge widget admin interface :

Total : default, no change.
![Image](https://github.com/user-attachments/assets/d1244856-ab98-4220-97fb-769244984b7d)

Ratio (global) :
![Image](https://github.com/user-attachments/assets/854282ec-451f-46cf-bd9f-c0308b70f780)
You fill the ratio value (%) in Challenge widget admin interface.

Ratio (per product) :
![Image](https://github.com/user-attachments/assets/743fee23-bf50-4d9a-a43a-15aacff490e5)
![Image](https://github.com/user-attachments/assets/52376fea-98a3-4b5f-aa9a-34d1c478c8c7)
Note that ratio is % but should be 12.53 (for example), so manage decimals.

A few rules :
- Ratio type should be defined on Challenge creation and cannot be updated after
- Ratio amount (global) should be filled on creation but can be updated after (with a lock like we did on product admin page for release date)
- Ratio amount (per product) should be filled AFTER creation (you create a challenge with Ratio per product type then you have the tab with an input for each product set in the challenge). Custom ratios are prefilled with 100 value.

coyotte508:
I think that in that case, in the challenge, we should store the amount collected for each product (regardless of challenge type) - so it's easier to adjust things later on.

No particular comment, other than it may be weird for a user to contribute X and see increase of Y

PR #1798, opened 2025-04-09
🆕 Challenge with ratio #1796

ithiame:


coyotte508:
As mentioned in https://github.com/be-BOP-io-SA/be-BOP/issues/1796#issuecomment-2785919432

Would be good to store the amount collected for each product in the challenge in DB

src/lib/server/orders.ts
```diff
@@ -1537,10 +1537,22 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
...
 							currency: item.customPrice?.currency || item.product.price.currency
 						}))
 				  );
+		const amountPerProduct = items.map((item) => ({
+			amount:
+				(item.customPrice?.amount || item.product.price.amount) *
+				item.quantity *
+				((challenge.globalRatio || challenge.perProductRatio?.[item.product._id] || 100) / 100),
+			currency: item.customPrice?.currency || item.product.price.currency,
+			productName: item.product.name
```
coyotte508, line 1554: btw we should really use `??` instead of `||` in all of these

src/lib/server/orders.ts
```diff
@@ -1551,7 +1563,8 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
 							type: 'progress',
 							at: new Date(),
 							order: order._id,
-							amount: increase
+							amount: increase,
+							amountPerProduct
```
coyotte508, line 1567: This is vulnerable to race conditions.

Also we want to store the full amount in that array, in `challenge.goal.currency` IMO.

So you can do  `updateOne(challengeId, {$inc: {[`amountPerProduct.${slug}]`: convertedAmount}})` (test if it works)

If you want to store in an array, it's possible too, see https://www.mongodb.com/docs/manual/reference/operator/update/positional/.

You can also check the result and `insert` if it wasn't updated.

ithiame: > Also we want to store the full amount in that array, in challenge.goal.currency IMO.
> 
> So you can do updateOne(challengeId, {$inc: {[amountPerProduct.${slug}]: convertedAmount}}) (test if it works)

I don't understand this two lines

coyotte508: > Also we want to store the full amount in that array, in challenge.goal.currency IMO.

Instead of doing `amountPerProduct += product.price * ratio`, it should just be `amountPerProduct += product.price`. That way if the ratio changes later we can recompute if we want to (at least we have the option)

> So you can do updateOne(challengeId, {$inc: {[amountPerProduct.${slug}]: convertedAmount}}) (test if it works)

Instead of resetting the whole `amountPerProduct` each time, you can type it as `Record<string, number>` and do a DB update operation that looks like this:

```ts
await db.challenges.updateOne({_id: ...}, {
  $inc: {
    [`amountPerProduct.${product.slug}`]:  toCurrency(challenge.goal.currency, price),
  }
})
```

src/lib/types/Challenge.ts
```diff
@@ -24,8 +24,12 @@ export type Challenge = Timestamps &
 			type: 'progress';
 			at: Date;
 			order: Order['_id'];
+			amountPerProduct?: Record<string, number>;
```
coyotte508, line 27: Oh I meant for this to be in the global zone, near  `ratio` and `globalRatio`

This is why I suggested using `$inc`

coyotte508 (approved):

Issue #1773, opened 2025-03-29
🦋 Add an "Disable cart icon" on Admin > Config > Config


Tirodem:
As surprising as it seems, someone asked me for a be-BOP to be used as a Point-of-Sale and a blog / website but not as a e-shop.
(And in fact it's kinda logical.)
So, to hide the e-shop aspects to the customer, we need to add a "Disable cart icon" option on Admin > Config > Config, near the "Disable language selector".

PR #1800, opened 2025-04-10
🦋 Add an 'Disable cart icon' on Admin > Config > Config #1773

ithiame:


coyotte508:
shluld also block from going to `/cart` (at least for regular users?)?

src/lib/server/runtime-config.ts
```diff
@@ -98,6 +98,7 @@ const baseConfig = {
 	collectIPOnDeliverylessOrders: false,
 	isBillingAddressMandatory: false,
 	disableLanguageSelector: false,
+	disableCartIcon: false,
```
coyotte508, line 101: `hideCartInToolbar` is more appropriate (if that's all we want to do)

cc @Tirodem to confirm that's all we want to do

Tirodem:
yes

src/routes/(app)/cart/+page.server.ts
```diff
@@ -2,11 +2,16 @@ import { checkCartItems } from '$lib/server/cart';
...
 import { refreshAvailableStockInDb } from '$lib/server/product.js';
+import { runtimeConfig } from '$lib/server/runtime-config.js';
 import { userIdentifier, userQuery } from '$lib/server/user.js';
 import { CUSTOMER_ROLE_ID } from '$lib/types/User';
 import { error, redirect } from '@sveltejs/kit';
 
 export async function load({ parent, locals }) {
+	if (runtimeConfig.hideCartInToolbar) {
+		throw error(403, 'This page is not accessible ');
+	}
```
coyotte508, line 13: Should only throw if user is a customer

coyotte508 (approved):

PR #1801, opened 2025-04-11
Physical cart minimum amount

ithiame:


src/routes/(app)/cart/+page.svelte
```diff
@@ -346,7 +353,17 @@
 				</div>
 			{/if}
 			<div class="flex justify-end">
-				<a href="/checkout" class="btn body-cta body-mainCTA">{t('cart.cta.checkout')}</a>
+				{#if !physicalCartCanBeOrdered}
+					<p class="text-red-500 font-light">Your cart doesn't contain minimum physical amount</p>
```
coyotte508, line 357: Should be translated?

coyotte508 (commented):
you should add a check in `createOrder` in backend. It's the central point called by everything when creating an order

Also add screenshots :)

ithiame:
Admin config:
image
Cart:
image
Cart Popup
image
Product
image
Product with quantity > min amount
image

src/routes/(app)/cart/+page.svelte
```diff
@@ -346,7 +353,17 @@
...
+					<p class="text-red-500 font-light">{t('cart.minimumPhysicalAmountText')}</p>
+				{/if}
+			</div>
+			<div class="flex justify-end">
+				<a
+					href="/checkout"
+					class="btn body-cta body-mainCTA {!physicalCartCanBeOrdered ? 'opacity-50' : ''}"
+					style="pointer-events: {!physicalCartCanBeOrdered ? 'none' : ''};"
+					>{t('cart.cta.checkout')}</a
+				>
```
coyotte508, line 366: make it a `button` with `disabled` attribute instead in that case, instead of using `pointer-events` and `opacity`

By the way here too: https://github.com/be-BOP-io-SA/be-BOP/blob/2c170ca9bb7129859995cd39ac179bf9fa2c6182/src/lib/components/ProductWidget/ProductWidgetPOS.svelte#L58-L61

That way it benefits from being more correct for accessibility purposes and from the theming too

src/routes/(app)/checkout/+page.server.ts
```diff
@@ -14,13 +14,42 @@ import type { JsonObject } from 'type-fest';
...
+				!!runtimeConfig.physicalCartMinAmount &&
+				priceInfoInitial.partialPriceWithVat >=
+					toCurrency(
+						priceInfoInitial.currency,
+						runtimeConfig.physicalCartMinAmount,
+						runtimeConfig.mainCurrency
+					);
+			if (!physicalCartCanBeOrdered) {
+				throw error(403, `Can't order a cart with amount < ${runtimeConfig.physicalCartMinAmount}`);
+			}
```
coyotte508, line 52: I think this error should be shown on the checkout page itself (like the cart error message) instead of crashing the page

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -479,7 +488,14 @@
...
 									formaction="?/addToCart"
-									disabled={!acceptRestriction || loading}
+									disabled={!acceptRestriction ||
+										(!!data.physicalCartMinAmount &&
+											data.product.price.amount * quantity <=
+												toCurrency(
+													data.product.price.currency,
+													data.physicalCartMinAmount,
+													data.currencies.main
+												))}
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:498]: you removed the ` || loading` here

Also I think we're still able to add to cart? So it shouldn't be disabled here?

Tirodem:
:x: Nope :
image
image
As it is, if I merge and deploy to an existing be-BOP, no cart cannot be ordered anymore.
Backward compatibility is mandatory.

Tirodem:
image
image
:x: The input is misplaced

Tirodem:
✅ Otherwise (+ conflict), it's OK.

pnpm-lock.yaml
coyotte508, line 1: This file should not be changed, package.json is not changed

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -492,8 +492,17 @@
...
+									class="btn body-cta body-mainCTA"
+									disabled={!acceptRestriction ||
+										loading ||
+										(!!data.physicalCartMinAmount &&
+											data.product.price.amount * quantity <=
+												toCurrency(
+													data.product.price.currency,
+													data.physicalCartMinAmount,
+													data.currencies.main
+												))}>{t(`product.cta.${verb}`)}</button
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:505]: extract this to a var at least, eg  `{@const cannotOrderPhysicalProduct = ...}` and here
```suggestion
										cannotOrderPhysicalProduct}>{t(`product.cta.${verb}`)}</button
```

coyotte508 (approved):

PR #1803, opened 2025-04-14
Analytic script snippet

ithiame:


src/lib/server/runtime-config.ts
```diff
@@ -121,6 +121,7 @@ const baseConfig = {
 		[TICKET_CHECKER_ROLE_ID]: false
 	},
 	plausibleScriptUrl: '',
```
coyotte508, line 123: you can remove this line

coyotte508 (approved):

Issue #1741, opened 2025-03-18
🦋 Improve and add new mode in Discount interface


Tirodem:
# Overview

![Image](https://github.com/user-attachments/assets/4200df23-7720-4a63-9bd7-6e6cc42dbf06)

Currently, our Discount interface allow to associate 1 to n subscriptions to 1 to n products (or whole catalog) and apply X% discount on them.
(still we have to manage #1740 )

We want to add an alternative mode that, instead of giving a discount on a product available as long as the subscription is active, will give X free copy of 1 to N products.

# Design

## Discount mode

![Image](https://github.com/user-attachments/assets/21bca79e-2fa8-4a2f-b6a8-75f98303cb73)

## Free copies mode

![Image](https://github.com/user-attachments/assets/5f82101b-8342-4060-be65-3f788208ffa0)

⚠ : note that free, subscriptions and PWYW products should not be added here.

PR #1814, opened 2025-04-16
🦋 Improve and add new mode in Discount interface #1741

ithiame:


src/lib/types/Discount.ts
```diff
@@ -7,8 +7,10 @@ export interface Discount extends Timestamps {
 	subscriptionIds: string[];
 	/* If empty, works on all products */
 	productIds: Product['_id'][];
-	percentage: number;
+	percentage?: number;
 	wholeCatalog: boolean;
 	beginsAt: Date;
 	endsAt: Date | null;
+	mode?: 'productPercentage' | 'freeProductQuantity';
+	quantityPerProduct?: Record<string, number>;
```
coyotte508, line 15:
```suggestion
	freeProductsById?: Record<string, number>;
```

src/lib/server/orders.ts
```diff
@@ -1617,7 +1646,8 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
 					paidUntil: add(new Date(), { [`${runtimeConfig.subscriptionDuration}s`]: 1 }),
 					createdAt: new Date(),
 					updatedAt: new Date(),
-					notifications: []
+					notifications: [],
+					...(freeProductsById && { freeProductsById })
```
coyotte508, line 1650:
```suggestion
					...(!isEmpty(freeProductsById) && { freeProductsById })
```

because `freeProductsById` is initialized wiht an empty object

src/routes/(app)/+layout.server.ts
```diff
@@ -203,7 +203,9 @@ export async function load(params) {
 			.filter((d) => d._id !== null)
 			.map((d) => [
 				d._id,
-				wholeDiscount !== undefined && wholeDiscount > d.discountPercent
+				wholeDiscount !== undefined &&
+				d.discountPercent !== undefined &&
+				wholeDiscount > d.discountPercent
```
coyotte508, line 208: this is not correct, if there's a whole discount and no individual discount, it will fallback to `d.discountPercent` which is `undefined` in that case

src/lib/server/orders.ts
```diff
@@ -626,9 +627,11 @@ export async function createOrder(
 			.filter((d) => d._id !== null)
 			.map((d) => [
 				d._id,
-				wholeDiscount !== undefined && wholeDiscount > d.discountPercent
+				wholeDiscount !== undefined &&
+				d.discountPercent !== undefined &&
+				wholeDiscount > d.discountPercent
 					? wholeDiscount
-					: d.discountPercent
+					: d.discountPercent ?? 0
```
coyotte508, line 634: This is still not correct.

Delete all the discounts you have.

Create just a whole discount of 50%.

Does the product still get the discount or not? (before / after this PR)

src/lib/server/orders.ts
```diff
@@ -1583,6 +1584,15 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
...
 		.toArray();
+	const discounts = await collections.discounts
+		.find({
+			subscriptionIds: {
+				$in: order.items
+					.filter((item) => item.product.type === 'subscription')
+					.map((sub) => sub.product._id)
+			}
+		})
+		.toArray();
```
coyotte508, line 1595: The DB call should only be run if some items are subscriptions

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -144,7 +145,7 @@ export const load = async ({ params, locals }) => {
 				}
 		  )
 		: null;
-
+	console.log(discount);
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:148]: can you remove?

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -294,7 +294,7 @@
 					<PriceTag
 						currency={data.product.price.currency}
 						amount={(data.product.hasVariations ? customAmount : data.product.price.amount) *
-							(data.discount ? 1 - data.discount.percentage / 100 : 1)}
+							(data.discount ? 1 - (data.discount.percentage ?? 0) / 100 : 1)}
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.svelte:297]:
```suggestion
							(data.discount?.percentage ? 1 - data.discount.percentage / 100 : 1)}
```

src/routes/(app)/admin[[hash=admin_hash]]/discount/new/+page.server.ts
```diff
@@ -29,32 +30,56 @@ export const load = async () => {
...
-				endsAt: data.get('endsAt') || undefined
+				wholeCatalog: formData.get('wholeCatalog'),
+				percentage: formData.get('percentage') || '0',
+				beginsAt: formData.get('beginsAt'),
+				endsAt: formData.get('endsAt') || undefined,
+				quantityPerProduct: quantityPerProductRecord,
+				mode: formData.get('mode')
 			});
-
+		console;
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/discount/new/+page.server.ts:82]: remove

src/routes/(app)/admin[[hash=admin_hash]]/discount/new/+page.server.ts
```diff
@@ -64,6 +89,8 @@ export const actions: Actions = {
 			wholeCatalog,
 			percentage: Number(percentage),
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/discount/new/+page.server.ts:90]: This can be NaN

Should either be unset or set to `null`

src/lib/types/Discount.ts
```diff
@@ -11,4 +11,6 @@ export interface Discount extends Timestamps {
 	wholeCatalog: boolean;
 	beginsAt: Date;
 	endsAt: Date | null;
+	mode?: 'productPercentage' | 'freeProductQuantity';
```
coyotte508, line 14:
```suggestion
	mode?: 'percentage' | 'freeProducts';
```

src/lib/types/Discount.ts
```diff
@@ -11,4 +11,6 @@ export interface Discount extends Timestamps {
...
+	quantityPerProduct?: Record<string, number>;
 }
```
coyotte508, line 16:
```suggestion
} & ({
	mode: 'percentage';
	percentage: number;
} | {
	mode: 'freeProducts';
	quantityPerProduct: Record<string, number>;
})
```

and a migration to add `mode` to existing discounts

you can use `z.union` on just the two fields when you need to use zod / clean up typing

src/routes/(app)/+layout.server.ts
```diff
@@ -140,7 +139,7 @@ export async function load(params) {
 		? await collections.discounts
 				.aggregate<{
 					_id: Product['_id'] | null;
-					discountPercent: Discount['percentage'];
+					discountPercent: number;
```
coyotte508, line 142: Should add `mode: 'percentage'` in the query

src/routes/(app)/admin[[hash=admin_hash]]/discount/[id]/+page.server.ts
```diff
@@ -46,30 +47,51 @@ export const actions = {
...
 			.object({
 				name: z.string().min(1).max(MAX_NAME_LIMIT),
-				productIds: z.string().array(),
+				productIds: z.string().array().optional(),
 				subscriptionIds: z.string().array(),
-				percentage: z.string().regex(/^\d+(\.\d+)?$/),
+				percentage: z
+					.string()
+					.regex(/^\d+(\.\d+)?$/)
+					.optional(),
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/discount/[id]/+page.server.ts:75]: It shouldn't be put as optional.


You parse one or the other, depending on `discount.mode`.

But currently on the frontend I can put all those as empty even for "percent" discounts

src/routes/(app)/admin[[hash=admin_hash]]/discount/new/+page.server.ts
```diff
@@ -29,45 +30,90 @@ export const load = async () => {
...
+		} = z
 			.object({
 				name: z.string().min(1).max(MAX_NAME_LIMIT),
 				productIds: z.string().array(),
 				subscriptionIds: z.string().array().min(1),
-				percentage: z.string().regex(/^\d+(\.\d+)?$/),
+				percentage: z
+					.string()
+					.regex(/^\d+(\.\d+)?$/)
+					.optional(),
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/discount/new/+page.server.ts:60]: Same comment this shouldn't be optional.

You can do a follow up `z.parse` below depending on `mode`

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -112,16 +113,32 @@ export const load = async ({ params, locals }) => {
...
 	const discount = subscriptions.length
 		? await collections.discounts.findOne(
 				{
-					$or: [{ wholeCatalog: true }, { productIds: product._id }],
+					$or: [
+						{ wholeCatalog: true },
+						{ productIds: product._id },
+						{ mode: 'percentage' },
+						{ mode: 'freeProducts' }
+					],
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:141]:
```suggestion
					$or: [
						{ wholeCatalog: true },
						{ productIds: product._id },
					],
					mode: 'percentage',
```

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -112,16 +113,32 @@ export const load = async ({ params, locals }) => {
...
+	const freeProductSubscriptions = await collections.paidSubscriptions
 		.find({
 			...userQuery(userIdentifier(locals)),
-			paidUntil: { $gt: new Date() }
+			paidUntil: { $gt: new Date() },
+			[`freeProductsById.${product._id}`]: { $exists: true }
 		})
+		.sort({ [`freeProductsById.${product._id}.available`]: -1 })
+		.limit(1)
 		.toArray();
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:124]: you can remove this and just use a `.find()` from the results of the below db calls.

Eg  `subscriptions.find(s => (s.freeProductsById?.[product._id]?.available ?? 0) > 0)`

You should also sum the available amount from all the free product subscriptions.

Maybe you can just pass the available amount to the frontend instead of the whole subscriptions

coyotte508 (commented):
ok with the two changes mentioned

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -113,36 +113,25 @@ export const load = async ({ params, locals }) => {
...
-		: await collections.paidSubscriptions
-				.find({
-					...userQuery(userIdentifier(locals)),
-					paidUntil: { $gt: new Date() }
-				})
-				.toArray();
+	const freeProductSubscriptions = subscriptions.filter(
+		(s) => (s.freeProductsById?.[product._id]?.available ?? 0) > 0
+	);
+	const percentageSubscriptions = freeProductSubscriptions ? [] : subscriptions;
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:125]: `freeProductSubscriptions` is always defined, you should use  `freeProductSubscriptions.length` or even compute the sum directly here.

By the way is there a reason for this line? In my opinion we still want to show the discounted price (so if they go beyond their free products, they already know the prices they will get)

src/routes/(app)/product/[id]/+page.server.ts
```diff
@@ -118,6 +119,9 @@ export const load = async ({ params, locals }) => {
 			paidUntil: { $gt: new Date() }
 		})
 		.toArray();
+	const freeProductSubscriptions = subscriptions.filter(
+		(s) => (s.freeProductsById?.[product._id]?.available ?? 0) > 0
+	);
```
Comment (coyotte508) [inline src/routes/(app)/product/[id]/+page.server.ts:124]:
```suggestion
	const freeProductsAvailable = sum(subscriptions.map(
		(s) => (s.freeProductsById?.[product._id]?.available ?? 0)
	));
```

coyotte508 (approved):

Tirodem:
Not working for me (detail communicated elsewhere)
Only thing working is the discount amount display on product page.

coyotte508:
> Not working for me (detail communicated elsewhere) Only thing working is the discount amount display on product page.

Maybe this

> For now it doesn't seem to actually allow the user to get free products, that's fine, but please indicate this in your PR (that this part needs to be done in next PR)

(it's not a full PR, the discount shows on the product page, but it's not propagated to order / cart)

ithiame:
Capture d’écran du 2025-04-28 12-30-30
Capture d’écran du 2025-04-28 12-29-51
Capture d’écran du 2025-04-28 12-27-20

Tirodem:
A bit of exploit here ☢️ 

`freeProductsById.id.available` is currently decremented and `freeProductsById.id.used` is currently incremented when an order is paid, seemingly.
It should be :
- decremented when an order is created as pending (or paid in case of full gratuity)
- incremented when the said order expired

As said, if I use all my free products + something non-free on a pending order, I can also use my free products on individual orders, then I can came back to my pending order and finalize it, that way I'll have more free products than expected.

We also have an issue with "quickdraw order" (no more detail here) but it's not related to this PR directly.

src/lib/server/cart.ts
```diff
@@ -163,6 +163,19 @@ export async function addToCartInDb(
...
+			{
+				...userQuery(params.user),
+				'items.product._id': product._id,
+				status: 'pending'
+			},
+			{ limit: 1 }
+		))
+	) {
+		throw error(400, 'You already have a pending order for this free product: ' + product.name);
+	}
```
coyotte508, line 178: I think what @Tirodem referred to is decreasing the available stock (maybe add it to a "pending" object) as soon as the order is created.

Then the `onOrderPaid` would move the stock from "pending" to "used", and on order cancel would move the stock from pending to available.

src/lib/server/orders.ts
```diff
@@ -371,6 +371,38 @@ export async function onOrderPaymentFailed(
...
+						$set: {
+							updatedAt: new Date(),
+							notifications: []
+						},
+						$unset: { cancelledAt: 1 }
+					}
+				);
+			}
+		}
+	}
```
coyotte508, line 405: this should only happen if the order is fully cancelled.

Like in this section: https://github.com/be-BOP-io-SA/be-BOP/pull/1847/files#diff-68e2d47d872b8e27c8f8a5ae4ccdb5409fe481e6c03a49b3c7f323530c1107c0R401-R417

```ts
	if (
		order.status !== ret.value.status &&
		(ret.value.status === 'canceled' ||
			ret.value.status === 'expired' ||
			ret.value.status === 'failed')
	) {
		await collections.scheduleEvents.updateMany(
			{
				orderId: order._id
			},
			{
				$set: {
					status: 'canceled'
				}
			}
		);
	}
```

src/lib/server/orders.ts
```diff
@@ -371,6 +371,46 @@ export async function onOrderPaymentFailed(
...
+									[`freeProductsById.${item.product._id}.available`]: +item.quantity
+								})
+							},
+							$set: {
+								updatedAt: new Date(),
+								notifications: []
+							},
+							$unset: { cancelledAt: 1 }
+						}
+					);
```
coyotte508, line 409: At the moment if I cancel an order with a paid product, if I have two subscriptoins with that product, both will get an increase of 1. Infinite free item glitch :)

src/lib/server/orders.ts
```diff
@@ -371,6 +371,34 @@ export async function onOrderPaymentFailed(
...
+								[`freeProductsById.${item.product._id}.used`]: -source.quantity,
+								[`freeProductsById.${item.product._id}.available`]: source.quantity
+							},
+							$set: {
+								updatedAt: new Date(),
+								notifications: []
+							},
+							$unset: { cancelledAt: 1 }
+						}
+					);
```
coyotte508, line 397: add the `{session}` param

src/lib/server/orders.ts
```diff
@@ -371,6 +371,34 @@ export async function onOrderPaymentFailed(
...
+					await collections.paidSubscriptions.updateOne(
+						{ _id: source.subscriptionId },
+						{
+							$inc: {
+								[`freeProductsById.${item.product._id}.used`]: -source.quantity,
+								[`freeProductsById.${item.product._id}.available`]: source.quantity
+							},
+							$set: {
+								updatedAt: new Date(),
+								notifications: []
```
coyotte508, line 393: why update the notifications? It's not related at all

src/lib/server/orders.ts
```diff
@@ -639,6 +670,48 @@ export async function createOrder(
...
+				await collections.paidSubscriptions.updateOne(
+					{ _id: sub._id },
+					{
+						$inc: {
+							[`freeProductsById.${item.product._id}.used`]: toUse,
+							[`freeProductsById.${item.product._id}.available`]: -toUse
+						},
+						$set: {
+							updatedAt: new Date(),
+							notifications: []
```
coyotte508, line 706: do not update the `notificatoins`

src/lib/server/orders.ts
```diff
@@ -639,6 +670,48 @@ export async function createOrder(
...
+						},
+						$unset: { cancelledAt: 1 }
```
coyotte508, line 708: not related? why this update?

src/lib/server/orders.ts
```diff
@@ -639,6 +670,48 @@ export async function createOrder(
...
+					}
```
coyotte508, line 709: add a `{session}` param

src/lib/server/orders.ts
```diff
@@ -1594,44 +1668,76 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
...
+									total: 0,
+									available: 0,
+									used: 0
+								};
+							}
+							freeProductsById[productId].total += quantity;
+							freeProductsById[productId].available += quantity;
+						}
+					}
+				}
```
coyotte508, line 1725: This whole diff can be removed no? Since we don't care what happens after the order is paid, because the changes happen when the order is created

coyotte508: ok but another problem probably, when you renew the subscription it should add new free products, currently the code doesn't do that (renewing the subcription doesn't add new products)

src/lib/server/orders.ts
```diff
@@ -639,6 +670,49 @@ export async function createOrder(
 
 	for (const item of items) {
 		item.discountPercentage ??= discountByProductId.get(item.product._id) ?? wholeDiscount;
+		if (item.freeQuantity) {
+			let quantityToConsume = Math.min(item.quantity, item.freeQuantity);
+			const freeProductSubscriptions = await collections.paidSubscriptions
+				.find({
+					[`freeProductsById.${item.product._id}`]: { $exists: true },
+					[`freeProductsById.${item.product._id}.available`]: { $gt: 0 },
```
coyotte508, line 678:
```suggestion
					[`freeProductsById.${item.product._id}.available`]: { $gt: 0 },
```

you should also restrict the subscriptoin to the user making the order, not all subscriptoins

src/lib/server/orders.ts
```diff
@@ -1594,44 +1669,95 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
...
+				const result = await collections.paidSubscriptions.updateOne(
+					{ _id: existingSubscription._id },
+					{
+						$set: {
+							paidUntil: add(max([existingSubscription.paidUntil, new Date()]), {
+								[`${runtimeConfig.subscriptionDuration}s`]: 1
+							}),
+							updatedAt: new Date(),
+							notifications: [],
+							...(Object.keys(freeProductsById).length !== 0 && { freeProductsById })
```
coyotte508, line 1717: you can use the `isEmpty` or `isEmptyObject` helper

coyotte508: also here you are "resetting" the free products, it should be added to the existing inside the subscriptoin

ithiame: It renews the subscription why not resetting free products ?

coyotte508 (approved):
:crossed_fingers:

Tirodem:
```
It should be :

- decremented when an order is created as pending (or paid in case of full gratuity)
- incremented when the said order expired
```
This is not done (or not working) :x:
Decrementation is only made on payment.

Tirodem:
:x: Standalone free product can be added ad nauseam one by one and still be free.
image
image

Tirodem:
Way better !
Sadly there's a bug left.
If I have 1 free product on my standalone product.
I add one, it's free ✅ 
I add another, it's not free ✅ 
I delete the free one from my cart and go to my product page : I have 0 free product left :x:
I delete the second one (not-free) from my cart and go to my product page : I have 1 free product left 🤷

coyotte508 (approved):

PR #1818, opened 2025-04-22
Add label directly in order page

ithiame:
Discord issue: 
https://discord.com/channels/1022592041709998080/1088417396232028170/1363575986540511442


src/routes/(app)/order/[id]/+page.server.ts
```diff
@@ -109,7 +111,9 @@ export async function load({ params, depends, locals }) {
 			)
 		}),
 		overwriteCreditCardSvgColor: runtimeConfig.overwriteCreditCardSvgColor,
-		hideCreditCardQrCode: runtimeConfig.hideCreditCardQrCode
+		hideCreditCardQrCode: runtimeConfig.hideCreditCardQrCode,
+		labels,
+		adminPrefix: adminPrefix()
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/+page.server.ts:116]: This exposes the admin prefix to every user that makes an order.

The admin prefix is supposed to be a secret, it's the whole reason for why it exists.

By the way you can check the data sent by the bebop to the user by refreshing the page and doing Ctrl+U

src/routes/(app)/order/[id]/fetchOrderForUser.ts
```diff
@@ -191,6 +191,7 @@ export async function fetchOrderForUser(orderId: string) {
 			userLogin: order.user.userLogin,
 			userAlias: order.user.userAlias
 		},
-		onLocation: order.onLocation
+		onLocation: order.onLocation,
+		orderLabelIds: order.orderLabelIds
```
Comment (coyotte508) [inline src/routes/(app)/order/[id]/fetchOrderForUser.ts:195]: Do we want to expose the labels to regular users? I don't think so :thinking: 

For `receiptNote` it's fine, the note is in the receipt

For `notes`, there's a link to the notes inside the order page so it's fine too

For `orderLabelids` it's purely internal I think.

The function is called `fetchOrderForUser`, it's to fetch only the data a regular user has access to. (feel free to add a doc comment to the function to make it clearer)

Maybe you can add a role extra param to `fetchOrderForUser` so it adds additional info depending on the role.

Comment (ithiame) [reply src/routes/(app)/order/[id]/fetchOrderForUser.ts:195]: Can we add this check ?
`		...(order.user.userRoleId !== CUSTOMER_ROLE_ID && { orderLabelIds: order.orderLabelIds })`

Comment (coyotte508) [reply src/routes/(app)/order/[id]/fetchOrderForUser.ts:195]: It won't work, `order.user`  is the user that made the order. Not the user getting the data at the moment

Tirodem:
Tests OK, ready for review before merge

coyotte508 (approved):

Issue #1745, opened 2025-03-21
🔌 Create visitor session from Wordpress account


Tirodem:
A pilot need for his customers to be logged on be-BOP through Wordpress.
As it seems there's no Wordpress SSO (everything I've found is about SSO from others platforms to Wordpress), I don't really know what to do here :
- use Wordpress REST API ?
- create a custom link to generate a custom token that'll be used in be-BOP to pre-fill the user session ?
- create a Wordpress plugin ?
- directly write on be-BOP database with a script to import Wordpress users ? 
If a generic solution is found, use https://github.com/be-BOP-io-SA/be-BOP/issues/1732 instead.

coyotte508:
The Wordpress needs to install this plugin: https://wordpress.org/plugins/miniorange-oauth-20-server/

Then on be-bop side we can add the possibility of a custom oauth provider, currently we have:

https://github.com/be-BOP-io-SA/be-BOP/blob/53aa6d27f7a9fee996c965499ab8c9461b36476b/src/hooks.server.ts#L390-L450

We can add a custom oauth provider there.

Tirodem:
Perfect 🔥 
Let's go !

PR #1821, opened 2025-04-22
Add dynamic SSO config

coyotte508:
Add multiple SSO config directly in /admin/oauth

Possibility to fold the current GITHUB/GOOGLE/... login into this

Fix https://github.com/be-BOP-io-SA/be-BOP/issues/1745

Example config: 
image

Steps:

- Install a wordpress **with https** (required with current lib I'm using, I could do additional development and stop using a lib)
- Add the miniOrange OAuth 2.0 Server/Provider plugin
- Create a  "Custom OpenID Connect" application with the correct redirect url (shown in /admin/oauth/new in bebop)
- Copy the issuer and create the bebop oauth app

I can handle other WP plugins / ... but it works with this setup at the moment

image

image

image


Tirodem:
It's good.
Only thing : the "Connecter avec Wordpress" button in /login triggers on hover (no need to clic to activate the CTA)

coyotte508:
> It's good. Only thing : the "Connecter avec Wordpress" button in /login triggers on hover (no need to clic to activate the CTA)

Maybe it's fixed automatically by upgrading sveltekit 2. Or we can do a quick fix for this PR. What do you thikn?

Tirodem (approved):

PR #1822, opened 2025-04-22
multi alias employee filter

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts
```diff
@@ -14,12 +14,24 @@ export async function load({ url }) {
...
-		employeeAlias: z.string().optional()
+		employeesAlias: z.string().array()
 	});
 	const queryParams = Object.fromEntries(url.searchParams.entries());
-	const result = querySchema.parse(queryParams);
-	const { beginsAt, endsAt, paymentMethod, employeeAlias } = result;
-
+	const result = querySchema.parse({
+		...queryParams,
+		employeesAlias: JSON.parse(String(url.searchParams.get('employeesAlias') ?? '[]')).map(
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.server.ts:22]: We should not have `?employeeAlias=[{value: '...', name: '...'}, ...]` in the url because of an implementation detail of `MultiSelect`, when we just need something like `?employeeAlias=X&employeeAlias=y`

It should be something like this:

```suggestion
		employeesAlias: url.sarchParams.getAll('employeesAlias')
```

You can use `input type="hidden` to fix it, like here: https://github.com/be-BOP-io-SA/be-BOP/blob/73208de6dc82b9b121d58aec4d9b756756518421/src/routes/(app)/admin%5B%5Bhash%3Dadmin_hash%5D%5D/order/%5Bid%5D/label/%2Bpage.svelte#L15-L27 in

(and you don't even need to add  `[{...}]` suffix to the query name this time)

coyotte508:
Also please link the issue in the PR description :)

coyotte508 (approved):

Issue #1589, opened 2024-11-16
🦋 Allow product to be added to cart in Nostr while they cannot be in website


Tirodem:
With this setting :
![image](https://github.com/user-attachments/assets/3e426cc8-4071-40bc-a225-ac12aa016074)
We have this :
![image](https://github.com/user-attachments/assets/e7304aad-35b2-4dba-98bc-10b64d0807c5)
By enabling eshop option, it goes OK again.
Permissions to add a product on Nostr should not depend of the eshop setting and be independant for nostr-bot exclusive products.

PR #1823, opened 2025-04-23
🦋 Allow product to be added to cart in Nostr while they cannot be in …

ithiame:
…website #1589

### DIFF
--- a/src/lib/server/cart.ts
+++ b/src/lib/server/cart.ts
@@ -69,12 +69,12 @@ export async function addToCartInDb(
 ) {
 	if (
 		params.user.userRoleId === POS_ROLE_ID
-			? !product.actionSettings?.retail?.canBeAddedToBasket
+			? !product.actionSettings.retail.canBeAddedToBasket
 			: params.mode === 'eshop'
-			? !product.actionSettings?.eShop?.canBeAddedToBasket
+			? !product.actionSettings.eShop.canBeAddedToBasket
 			: params.mode === 'nostr'
-			? !product.actionSettings?.nostr?.canBeAddedToBasket
-			: !product.actionSettings?.retail?.canBeAddedToBasket
+			? !product.actionSettings.nostr.canBeAddedToBasket
+			: !product.actionSettings.retail.canBeAddedToBasket
 	) {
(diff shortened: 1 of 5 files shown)


src/lib/server/cart.ts
```diff
@@ -69,7 +69,8 @@ export async function addToCartInDb(
 	if (
 		params.user.userRoleId === POS_ROLE_ID
 			? !product.actionSettings?.retail?.canBeAddedToBasket
-			: !product.actionSettings?.eShop?.canBeAddedToBasket
+			: !product.actionSettings?.eShop?.canBeAddedToBasket &&
+			  !product.actionSettings?.nostr.canBeAddedToBasket
```
coyotte508, line 73: The mode ('eShop', 'nostr', 'pos'...) should be added as a param to  `addToCartInDb` and the check should depend on that mode.

Because with this PR a user could still add a nostr-exclusive item using the API instead of using NostR

src/lib/server/cart.ts
```diff
@@ -64,14 +64,21 @@ export async function addToCartInDb(
 		customPrice?: { amount: number; currency: Currency };
 		deposit?: boolean;
 		chosenVariations?: Record<string, string>;
+		mode?: 'eshop' | 'nostr' | 'pos';
```
coyotte508, line 67: can you make this parameter mandatory?

```suggestion
		mode: 'eshop' | 'nostr' | 'pos';
```

you can use Ctrl+Shift+F in VS code or just shift+F12 when on the function to find all references.

Currently all references have the "mode" added.So it can be made mandatory

src/lib/server/cart.ts
```diff
@@ -64,14 +64,21 @@ export async function addToCartInDb(
...
 	}
 ) {
 	if (
 		params.user.userRoleId === POS_ROLE_ID
 			? !product.actionSettings?.retail?.canBeAddedToBasket
-			: !product.actionSettings?.eShop?.canBeAddedToBasket
+			: params.mode
```
coyotte508, line 73: This line can be removed once it's made mandatory

coyotte508 (approved):

PR #1825, opened 2025-04-23
Remove lodash

coyotte508:
I think the only way to prevent it being used more is to remove the package altogether.

Better typing, smaller footprint. Avoid issues with `groupBy`, there was a bug about it, where the typing is not correct (`Record` instead of `Partial<Record>`) and values can be undefined

cc @ithiame 

Also fixes:

- Some optimizations (DB calls)
- Do not expose digital links when item is added to cart
- Do not expose internal note to non-POS users (although I don't think there was a problem here in the first place)

tests:

- [x] Digital files icon still show
- [x] intenal note still works in POS
- [x] updating product / ... still works in admin
- [ ] tag product sorting - actually I don't think this one worked correctly @Tirodem so I rewrote the sorting
- [x] test custom translations


coyotte508:
Merging, I tested manually a few things that could've been impacted, added code unit tests, this is not a general bug fix or new feature, unless we want to do a 100% retest.

Issue #1837, opened 2025-04-28
🦋 Add an option in Admin > Pos to pre-fill term of use checkbox in /checkout (for PoS account only)


Tirodem:
self-explaining
this should not apply to employee non-PoS account and, obviously, neither to customers.

PR #1839, opened 2025-04-29
🦋 Add an option in Admin > Pos to pre-fill term of use checkbox in /c…

ithiame:
…heckout (for PoS account only) #1837

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts
@@ -62,6 +62,7 @@ export const actions = {
 				upsert: true
 			}
 		);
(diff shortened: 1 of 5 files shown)


Tirodem:
Tests are OK, to merge if PR review is good.

src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts
```diff
@@ -45,6 +48,20 @@ export const actions = {
 				upsert: true
 			}
 		);
+		await collections.runtimeConfig.updateOne(
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts:51]: should also do `runtimeConfig.posPrefillTermOfUse = ...`

coyotte508 (approved):

Issue #1256, opened 2024-05-09
🦋 On tag widget description, allow basic rich text on long description


Tirodem:
At least :
- bold
- italic
- underline
- hyperlink

Topic to discuss (markdown ? html ? something else ?)

Tirodem:
Drop issue, it works already with markdown :

![Image](https://github.com/user-attachments/assets/49de044a-8ba7-4231-bcb7-af1f8d11799e)

![Image](https://github.com/user-attachments/assets/89970363-3f8c-451e-87b7-a908c4cb423e)

PR #1851, opened 2025-04-30
🦋 On tag widget description, allow basic rich text on long descriptio…

ithiame:
…n #1256

### DIFF
--- a/src/lib/components/TagWidget/TagWidgetVariation1.svelte
+++ b/src/lib/components/TagWidget/TagWidgetVariation1.svelte
@@ -24,7 +24,7 @@
 				<h2 class="text-6xl body-title pb-2 {titleClassNames}">{tag.title}</h2>
 				<h2 class="text-md md:text-xl">
 					<!-- eslint-disable svelte/no-at-html-tags -->
-					{@html marked(tag.content.replaceAll('<', '&lt;'))}
+					{@html tag.content}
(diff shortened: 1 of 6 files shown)


src/lib/components/TagWidget/TagWidgetVariation1.svelte
```diff
@@ -24,7 +23,7 @@
 				<h2 class="text-6xl body-title pb-2 {titleClassNames}">{tag.title}</h2>
 				<h2 class="text-md md:text-xl">
 					<!-- eslint-disable svelte/no-at-html-tags -->
-					{@html marked(tag.content.replaceAll('<', '&lt;'))}
```
coyotte508, line 27: You need to do something like this: https://github.com/be-BOP-io-SA/be-BOP/blob/5ad1de08eeedae9aea8ad4603279ecde13c2a8c2/src/lib/server/cms.ts#L183

then

to the tag's content when you fetch it from DB

In general be very careful whenever using `@html` or it can add big security holes to the project.

Tirodem:
> Also the tag content supports markdown currenrtly @Tirodem , do we need this PR? I see that in thee issue it says markdown is ok.

Does it ? :o
I didn't know, we'll test it !

Tirodem:
Indeed, it works. We can drop this PR.
image

Issue #594, opened 2023-11-10
Dark / Light style selector v2 - add a 3rd option "follow OS / browser mode"


Tirodem:
Following #593 & https://github.com/B2Bitcoin/beBOP/issues/593#issuecomment-1806351494

Add the 3rd option to follow the browser mode if the shop owner allows it.

Tirodem:
A identifier : quel est l'élément donné par le browser qui nous permet de connaître cela.
https://amourycodes.com/notes/how-to-detect-if-the-user-prefers-dark-mode/297651110ac24c0ab17389fb0c53904a
https://ultimatecourses.com/blog/detecting-dark-mode-in-javascript
https://www.hanselman.com/blog/how-to-detect-if-the-users-os-prefers-dark-mode-and-change-your-site-with-css-and-js

PR #1863, opened 2025-05-02
Dark / Light style selector v2 - add a 3rd option 'follow OS / browse…

ithiame:
…r mode' #594

### DIFF
--- a/src/lib/stores/theme.ts
+++ b/src/lib/stores/theme.ts
@@ -2,7 +2,9 @@ import { browser } from '$app/environment';
 import { writable } from 'svelte/store';
 
 const initialValue = browser
-	? document.querySelector('html')?.classList?.contains('dark')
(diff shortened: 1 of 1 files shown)


coyotte508 (dismissed):

coyotte508 (approved):
nice, if test ok, code ok too :)

Tirodem:
image
:x: "fo" > for
:x: "Brower /OS" > "Browser / OS"
:x: it "works" but it should not replace the theme light / night selector. What I imagined is for the visitor / user or for the employee to land with the parameter "light" or "dark" pre-selected. So, if mode is "dark mode", I land with "dark mode" but I can change it through selector. And if mode is "corresponding to browser / os" and if my browser mode is "dark mode", I land with "dark mode" but I can change it through selector.

Tirodem:
Tests are OK for me ✅

Issue #1869, opened 2025-05-05
🦋 Add a VAT calculator to admin product page


Tirodem:
# Overview

Include a calculator in admin product page (creation & edition) to help employees from a be-BOP to generate their product HT price.

# Design

![Image](https://github.com/user-attachments/assets/9f762866-ca0b-4e0d-8349-0a2927e76493)

![Image](https://github.com/user-attachments/assets/5b2a298c-2889-4db4-baad-37ceeb6c208e)

![Image](https://github.com/user-attachments/assets/eb4ef732-b2dc-477c-80d1-0770a46526e3)

PR #1870, opened 2025-05-05
🦋 Add a VAT calculator to admin product page #1869

ithiame:


coyotte508 (approved):
it's just frontend, as long as it behaves ok, ok for me

Issue #1217, opened 2024-04-16
🦋 Improve [Picture=] options


Tirodem:
To define

tl;dr (each time, "add option or document / explain current option if it can be done") :
- right align
- center align
- expand on whole body width

PR #1875, opened 2025-05-05
🦋 Improve [Picture=] options #1217

ithiame:


Tirodem:
Does not work under 1023px.
Otherwise, it's OK.
We can go like that I think.

coyotte508 (approved):

Issue #1672, opened 2025-02-10
Admin > Reporting - invoice generation flow


AmadouAgiltoo:
When refreshing or changing filters:

- If a "Prepare PDF" was done on the first search,

- On the second search, the "Print" button still shows the PDF generated from the first search,

- Instead, the "Prepare PDF" button should be shown again to generate a new export based on the updated search results.

PR #1886, opened 2025-05-09
Admin > Reporting - invoice generation flow #1672

ithiame:


coyotte508 (approved):

Tirodem:
On Admin > Transactions > Reporting :
image
dev-bootik2

Tirodem:
@roosemberth ping me when you have time next week to check logs on sandbox (if you have access to testnet and mainnet nodls -otherwise, we'll ask for it-)

Tirodem:
It "works" but the CTA becomes "Print" before the end of this call :
image
So, after refreshing your filters, you can print orders again... but the new batch of orders is not loaded yet.
While the call to `__data.json?blablabla` is processing, the CTA must become "Print" but be disabled.
Once the call is finished, then, the CTA must become enabled, not earlier.

Tirodem:
It's not perfect.
I landed on the Admin > Transactions > Reporting.
I launched a search with 71 results.
I prepare the pdf.
Then I changed the search criteria to go down to 10 results and validated.
The "Prepare PDF" CTA was disabled : ✅ 
Then, once the call was finished, the CTA was enabled, and I clicked :
- "Preparing invoices 1/71" instead of 10 :x:
- 71 results displayd in the reporting instead of 10 :x:
- The the page reloaded / went to the top of the page
- Then there was 10 results and the reporting was good ✅ 

So, the timing is still wrong.
If we reload the page with new results (did we do that before this PR ?), we must keep the "Prepare PDF" disabled until that.

Tirodem:
Functional is now fine ✅ 
@roosemberth for PR review

Issue #1879, opened 2025-05-06
🦋 Add margin-left to bulk editors


Tirodem:
Following #1878 
To make bulk editors more readable, add the same margin-left on each entry after title on these pages :
- /admin/product/prices
- /admin/picture/name

PR #1887, opened 2025-05-09
🦋 Add margin-left to bulk editors #1879

ithiame:


Tirodem:
Thanks, you did it on product aliases, I forgot :
image

But it's not done on product prices :
image

Issue #1746, opened 2025-03-23
⚡ Swiss Bitcoin Pay integration


Tirodem:
TBD

https://docs.swiss-bitcoin-pay.ch/

As we are already able to cash-in in Lightning, this integration will allow :
- to give SBP users a better e-shop & Point of Sale solution than Prestashop & co
- to give a quick answer to NFC management in Point of Sale experience

Tirodem:
https://github.com/be-BOP-io-SA/be-BOP/issues/1902 for NFC features, I'm closing this one.

PR #1894, opened 2025-05-13
Swiss bitcoin pay integration

roosemberth:
This PR partially addresses #1746 by enabling the use of [Swiss Bitcoin Pay](https://swiss-bitcoin-pay.ch) as a lightning payment processor.

Please see commit messages.


src/lib/server/orders.ts
```diff
@@ -2310,3 +2328,35 @@ function compareBookingsAndThrow(
...
+}> {
+	const accountingNote = `Order #${params.orderId}`;
+	const device = runtimeConfig.brandName;
+	try {
+		const checkout = await sbpCreateCheckout({
+			title: params.label,
+			amount: params.toPay.amount,
+			unit: params.toPay.currency === 'SAT' ? 'sat' : params.toPay.currency,
+			extra: {
+				customNote: accountingNote,
```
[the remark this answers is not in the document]
roosemberth: Applied suggestion. Thanks!

src/lib/server/swiss-bitcoin-pay.ts
```diff
@@ -0,0 +1,89 @@
...
+		throw new Error('Swiss Bitcoin Pay is not enabled');
+	}
+
+	const response = await fetch(`${apiUrl()}/checkout`, {
+		method: 'POST',
+		headers: {
+			'Content-Type': 'application/json',
+			'api-key': runtimeConfig.swissBitcoinPay.apiKey
+		},
+		body: JSON.stringify(request)
```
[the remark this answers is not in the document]
roosemberth: I recon it is; according to [the MDN documentation on `fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#setting_a_body), the object would otherwise be converted using `toString()`.

The examples provided by MDN also use JSON.stringify.

src/lib/server/locks/order-lock.ts
```diff
@@ -149,42 +150,80 @@ async function maintainOrders() {
...
-
-								if (invoice.isPaid) {
-									order = await onOrderPayment(
-										order,
-										payment,
-										{
-											amount: invoice.receivedSat,
+							if (!payment.processor) {
+								throw new Error('Missing processor on lightning payment');
+							}
```
coyotte508, line 157: Should add a migration in `migrations`, to add "lnd" processor to each lightning payment without processor. As in the past when there was only lnd, `processor` was not filled.

The migration file is here: https://github.com/be-BOP-io-SA/be-BOP/blob/42c59cc81b31e8894f5137c113b0aeb12d964368/src/lib/server/migrations.ts

Note for the `_id` of the migration, you can use `ObjectId.createFromTimestamp(getUnixTime(new Date("2025-05-13T08:59:51.997Z")))`.

Not sure if there are any pilots using the lnd client @Tirodem (if there are not, the migration is not needed)

src/lib/server/orders.ts
```diff
@@ -1590,6 +1591,23 @@ async function generatePaymentInfo(params: {
...
+						amount: satoshis,
+						currency: 'SAT'
+					}
+				});
+				return {
+					address: invoice.payment_address,
+					invoiceId: invoice.invoiceId,
+					processor: 'swiss-bitcoin-pay'
+				};
+			} else {
```
coyotte508, line 1609: I think you can make this higher priority than LND maybe (put the if branch before lnd's)

cc @Tirodem

coyotte508 (approved):
Nice!

I think the UI on the order/payment page is missing but can be done in another PR.

Just the migration to add

src/lib/server/locks/order-lock.ts
roosemberth: Added the migration.

src/lib/server/orders.ts
roosemberth: I've moved SBP higher up.

src/lib/server/migrations.ts
```diff
@@ -509,6 +509,17 @@ const migrations = [
 				{ session }
 			);
 		}
+	},
+	{
+		_id: new ObjectId('68230a070000000000000000'),
+		name: 'Ensure payment processor is set on lightning payments',
+		run: async (session: ClientSession) => {
+			await collections.orders.updateMany(
+				{ 'payments.method': 'lightning', 'payments.processor': '' },
```
coyotte508, line 518:
```suggestion
				{ 'payments': { $elemMatch: {method: 'lightning', processor: { $exists: false } } } },
```

When we want to add multiple conditions that a specific element in an array needs to match, we need to use `$elemMatch`: https://www.mongodb.com/docs/manual/reference/operator/query/elemMatch/

Also I think the absent processor would be `{$exists: false}`

coyotte508 (approved):
looks good, feel free to merge after suggestion taken care of

src/lib/server/migrations.ts
roosemberth: Applied (and tested) suggestion.

Issue #1896, opened 2025-05-13
🐛 Theme selector crashes mobile display if used


Tirodem:
![Image](https://github.com/user-attachments/assets/5019870b-0eef-42eb-8f87-ba266ebc3a70)

PR #1900, opened 2025-05-13
🐛 Theme selector crashes mobile display if used #1896

ithiame:


Tirodem:
Before :
image
After :
image

It's better but (especially here with no navbar entry so no burger menu) it'll be best to align the cart icon, lang selector and theme selector to the right.

Can you take the opportunity of this PR to do that please ?

Tirodem:
Tests OK ✅

Issue #1832, opened 2025-04-28
🦋 Allow multiple pictures upload at the same time


Tirodem:
TBD
Especially for product pictures upload

PR #1904, opened 2025-05-14
🦋 Allow multiple pictures upload at the same time #1832

ithiame:


src/routes/(app)/admin[[hash=admin_hash]]/product/new/+page.server.ts
```diff
@@ -136,135 +136,139 @@ export const actions: Actions = {
...
-							name: parsed.name,
-							isTicket: parsed.isTicket,
-							price: {
-								currency: parsed.priceCurrency,
-								amount: priceAmount
+		await withTransaction(async (session) => {
+			parsed.pictureIds.map(async (pictureId) => {
+				if (pictureId) {
+					await generatePicture(pictureId, {
+						productId: parsed.slug
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/new/+page.server.ts:143]: with this option, you should pass the `session` parameter to `generatePicture`

coyotte508 (commented):
LGTM

Tirodem:
- Doesn't seem to work for existing product
- Works on new product

Please allow it for existing product.
In the future it would be interesting to allow multiple file upload on a product, too.

Tirodem:
Tests are OK for me ✅

src/routes/(app)/admin[[hash=admin_hash]]/picture/new/+page.server.ts
```diff
@@ -33,8 +39,16 @@ export const actions: Actions = {
 		});
 
 		if (fields.productId) {
+			await Promise.all(
+				fields.pictureIds.slice(1).map((pictureId) =>
+					generatePicture(pictureId, {
+						productId: fields.productId
+					})
+				)
+			);
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/picture/new/+page.server.ts:48]: - Why is are the images only generated if the product is set?
- Why only the first image slider, tag and schedule are set?

Comment (ithiame) [reply src/routes/(app)/admin[[hash=admin_hash]]/picture/new/+page.server.ts:48]: - because i only handle for now multi images upload for product only. So if the product is set and we upload multi image i generate the other image instead if  slider, tag or schedule is set we generate only one image for now.

roosemberth (approved):

Issue #1889, opened 2025-05-11
🦋 Portugal is missing as delivery / billing country in country list


Tirodem:
Self explaining.
Check if other countries are missing.
We need an exhaustive listing to prepare future issues like #1855 & #1857

PR #1905, opened 2025-05-14
Portugal is missing as delivery / billing country in country list #1889

ithiame:


coyotte508 (commented):
what you don't like "le Portugal" xD?

LGTM

Tirodem:
Oh, it was a "Le" issue ?
We probably have to do something for these contries so :
image

Issue #1842, opened 2025-04-29
🦋 Improve !catalog option in nostr-bot


Tirodem:
# Overall

For glamorous demo purpose and future real use, let's add a feature in nostr-bot to search in a more practical way into the catalog.
As Nostr doesn't have pagination, it'll be helpful.

# Detail

Add options to `!catalog` command :
- `!catalog s:searchterm` : the `s` option will only show product having `searchterm` in their title, short description or long description
- `!catalog t:type` : the `t` option will only show product with the said type, types being :
  - digital (without delivery shipping component)
  - ship (with delivery shipping component)
  - free (amount = 0)
  - pwyw (pay what you want)
  - event (event ticket)
  - preorder (future release date with preorder option)
  - sub (subscription)
  - tip (donation)
- `catalog m:mode` : the `m` option will change the display mode :
  - `m:light` : provide the product name, price and slug
  - `m:mid` : provide the product name, price, slug and url ; serves as default mode for !catalog
  - `m:full` : provide the product name, short description, long description, price, web URL

Allow to combine parameters.
For example : `!catalog s:bitcoin t:event t:free m:light`

PR #1908, opened 2025-05-15
🦋 Improve catalog option in nostr-bot #1842

ithiame:


coyotte508 (changes requested):
So far all commands take only "positional" arguments, but they should be able to take options. And `s:xxx` or `t:xxx` is a valid syntax (there are other like `-s blabla`. But it requires some development to support that new featureset.

There's also the concept of shorthands, eg the user could `search:xxx` or `s:xxx`.

Finally, the help text really needs to be on point. Here it's not understandable at all with just the "s", "t", and "m". If the args are positional the minimum is to use full names for the args. But they're going to _not_ be positional, so they should have a short name and a full name.

Tirodem:
I +1 about the help text, it's not understandable.
And only answers I have now is "Catalog is empty" or, worse, nothing.
Maybe it's due to buffer between 2 identical commands  : !catalog and !catalog with-paramaters should be treated as different orders and should not be ignored even if done both in a small time period.

src/lib/server/locks/handle-messages.ts
```diff
@@ -259,101 +259,120 @@ const commands: Record<
...
-				name: 't',
-				enum: ['digital', 'ship', 'free', 'pwyw', 'event', 'preorder', 'sub', 'tip', 'all'],
-				default: 'all'
-			},
-			{
-				name: 'm',
-				enum: ['light', 'mid', 'full'],
-				default: 'light'
+				name: 'options',
+				default: 's:all t:all m:light'
```
coyotte508, line 267: How does the user know that the mode can be "light", "mid" or "full"?

IMO a real "option" support should be implemented (eg non-positional), with listing each arg value.

Except that compared to traditional CLI, where it's `--argname value`, here it's `argName:value`.

ithiame: > IMO a real "option" support should be implemented (eg non-positional), with listing each arg value.

for all commands or just for catalog ?

coyotte508: for all commands, eg the args are always positoinal at the moment but they should have `option: true` or `positional: true` in their definition or `type: 'positional' | 'option'`  - although @roosemberth is the more appropriate to chime in for this since he'll have to review

ithiame: > IMO a real "option" support should be implemented (eg non-positional), with listing each arg value.

It's not positionnal actually, with the last commit

ithiame:
With one arg i can handle the non position issue and i don't need to make change for all commands.

Actually i tried to use 3 args but it can make it non-positionnal maybe if you have clear suggestion @coyotte508 

For now the latest commit takes into account what we want to do on this issue

src/lib/server/locks/handle-messages.ts
roosemberth: I agree with coyotte: We should simply define a grammar for all commands:

```
commandline = command, { option }, varargs, newline ;
command     = "!", identifier ;
option      = space, ( shortOpt | longOpt ), ":", value ;
varargs     = space, { character - newline } ;
```

Looking at the current parser it's becoming hard to understand it's grammar. I think we should move the parsing out of the `handle-message` module into its own file (e.g. `nostr-message-parser`.

(In the future, we should start looking at DSLs like [Ohm](https://ohmjs.org/), but I don't think there's need for that yet.)

Once the message has been parsed (e.g. `{ command: "foo", opts: {"s": "bar"}, arg: "baz"}`), we give it to the command handler (ie. `execute`) and the command handler does whatever they want with the options (including refusing "unknown" options).

---

I think it's best to implement hitherto **before** augmenting the `catalog` command. For example, the first commit can add the `nostr-message-parser` module and update `handleReceivedMessage` to use it; then the second commit can augment the `catalog` command.

Issue #1759, opened 2025-03-25
📆 Set GMT option on Schedule


Tirodem:
tl;dr when you fill a Schedule widget form and you're not on the same timezone, you have 08:00 on form and 10:00 on CMS widget, it's bothersome

![Image](https://github.com/user-attachments/assets/c0579daf-c963-4f5d-b966-9966298b1cc4)

![Image](https://github.com/user-attachments/assets/aad8e8c0-6f2e-4255-b24e-7bfc03511a0a)

![Image](https://github.com/user-attachments/assets/3b85096b-4cb3-4504-bda9-6b7d5b258c18)

GMT timezone should be pre-setted with employee system current timezone when checking the box, if possible.
Once setted, it should not be reset with default value and stay to the setted value.

PR #1911, opened 2025-05-16
Set GMT option on Schedule #1759

ithiame:


src/lib/components/ScheduleWidget/ScheduleWidgetCalendar.svelte
```diff
@@ -172,26 +176,38 @@
 					{event.title}
 					{#if event.endsAt && isSameDay(event.endsAt, event.beginsAt)}
 						{t('schedule.dateText', {
-							beginTime: event.beginsAt.toLocaleTimeString($locale, {
+							beginTime: toZonedTime(
+								event.beginsAt,
+								schedule.timezone ?? Intl.DateTimeFormat().resolvedOptions().timeZone
```
roosemberth, line 181: The expression `schedule.timezone ?? Intl.DateTimeFormat().resolvedOptions().timeZone` is repeated multiple times, it's long and clutters the code.

We could probably extract the pattern `toZonedTime(_, schedule.timezone ?? Intl.DateTimeFormat().resolvedOptions().timeZone)` into a function; this would make the intent clear and improve the readability.

I won't comment other instances, but they should all be addressed.

roosemberth (changes requested):

roosemberth (approved):
LGTM. Thank you!

roosemberth:
@Tirodem for QA

Tirodem:
- [ ] On the list, I cannot search through test (like in the countries list in /checkout for example, where I can do it). Considering the number of entries, it's a big pain. Can something be done here ? > edit : in fact, it's because it's Continent/Country, it's a bit confusing. Can we change that ? Either we should allow country / state, or order by (and display) the timezone, it's more usual, and usable because it includes less entries.
image
- [ ] Do we manage winter / summer time variations on a same GMT timezone when it changes ?
Otherwise, it seems functionnal, but I need change on the 1st point please.

Issue #1616, opened 2024-12-02
🦋 Add PoS specific options to any employee profile


Tirodem:
#1049 
(to allow any role to have PoS specific features like discount, VAT & co, and to allow to create different kinds of PoS role variations with more or less access)

Tirodem:
For memories, point of sales action are :
- on /checkout 
  - TVA exemption
  - discount
  - freeing delivery fees
  - pre-fill of delivery target address depending of your setting

PR #1916, opened 2025-05-20
🦋 Add PoS specific options to any employee profile #1616

ithiame:


coyotte508 (commented):
SGTM

It's annoying to have two admin accounts, one for POS and one for admin

src/lib/server/orders.ts
```diff
@@ -563,7 +563,8 @@ export async function createOrder(
 		!canBeNotified &&
 		paymentMethod !== 'point-of-sale' &&
 		paymentMethod !== null &&
-		params.user.userRoleId !== POS_ROLE_ID
+		params.user.userRoleId !== POS_ROLE_ID &&
+		!params.user.userHasPOSOptions
```
roosemberth, line 567: Can you please move the whole "if" condition into a variable?

It's really hard to understand what the intent of this check is. This can easily be fixed by moving the condition into a variable and naming the variable accordingly.

src/hooks.server.ts
```diff
@@ -221,7 +221,8 @@ const handleGlobal: Handle = async ({ event, resolve }) => {
 						_id: user._id,
 						login: user.login ? user.login : '',
 						roleId: user.roleId,
-						alias: user.alias
+						alias: user.alias,
+						hasPOSOptions: user.hasPOSOptions
```
roosemberth, line 225: Since "any" user can now have access to the POS options; it's redundant to keep the `POS_ROLE_ID` role.

I think we should either remove `POS_ROLE_ID` in favor of `hasPOSOptions` or replace `hasPOSOptions` with a method `canAccessPosOptions()` that takes care of doing the projection, regardless of how the user gained the capability.

src/lib/server/orders.ts
```diff
@@ -648,7 +649,10 @@ export async function createOrder(
 	}
 
 	const discountInfo =
-		params.user.userRoleId === POS_ROLE_ID && params?.discount?.amount ? params.discount : null;
+		(params.user.userRoleId === POS_ROLE_ID || params.user.userHasPOSOptions) &&
```
roosemberth, line 652: Per my comment above, simplify this and leave the projection at the `User` interface.

I will not comment on other occurrences, but the pattern "userRoleId == POS_ROLE || userHasPOSOptions` should be removed.

src/lib/server/orders.ts
```diff
@@ -814,7 +818,11 @@ export async function createOrder(
 		currency: Currency;
 		amount: number;
 	} | null = null;
-	if (priceInfo.discount && params.discount && params.user.userRoleId === POS_ROLE_ID) {
+	if (
+		priceInfo.discount &&
+		params.discount &&
+		(params.user.userRoleId === POS_ROLE_ID || params.user.userHasPOSOptions)
+	) {
```
roosemberth, line 825: Please move this condition into a variable and name the variable accordingly.

src/routes/(app)/admin[[hash=admin_hash]]/arm/role/new/+page.svelte
```diff
@@ -14,7 +14,10 @@
 		Role name
 		<input class="form-input" type="text" name="name" required placeholder="Role name" />
 	</label>
-
+	<label class="checkbox-label">
+		<input class="form-checkbox" type="checkbox" name="hasPOSOptions" />
+		Has POS
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/arm/role/new/+page.svelte:19]:
```suggestion
		Has POS access
```

Even better, "Point-of-sale access". I don't know if we already use "POS" elsewhere; in that case, it's fine to use POS.

I won't comment other occurrences of "Has POS" in UI files, but they should all be addressed.

roosemberth (changes requested):
I find `hasPOSOptions` a bit awkward since the camel case [convention is to treat acronyms like regular words](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case); for example "HTML parser" -> "htmlParser".

`hasPosOptions` would be less awkward.

src/lib/server/orders.ts
ithiame: Have deleted POS_ROLE

ithiame:
I rollback the POS_ROLE_ID && add migration  hasPosOptions for any existing user whose role was POS_ROLE_ID.

roosemberth (approved):

Tirodem:
It's good now.

PoS existing accounts have the option :
image

To be fair it should be done on role account also :
image

src/routes/(app)/admin[[hash=admin_hash]]/arm/+page.svelte
```diff
@@ -209,6 +214,13 @@
 						</option>
 					{/each}
 				</select>
+				<input
+					class="form-checkbox place-self-center"
+					type="checkbox"
+					name="hasPosOptions"
+					bind:checked={user.hasPosOptions}
+					disabled={data.roles.find((rol) => rol._id === user.roleId)?.hasPosOptions}
+				/>
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/arm/+page.svelte:223]: If the user role has POS access, the user checkbox will be disabled (as expected), but unchecked.

It would be best if the checkbox was checked and disabled; a tooltip could be displayed on hover explaining the user this user has POS access because of their role.

src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts
```diff
@@ -57,6 +65,7 @@ export const actions = {
 							...(parsed.recoveryEmail && { email: parsed.recoveryEmail }),
 							...(parsed.recoveryNpub && { npub: parsed.recoveryNpub })
 						},
+						hasPosOptions,
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:68]: In line 56 of this file:

```
const hasPosOptions = parsed.hasPosOptions || role.hasPosOptions;
```

On line 217-223:

```
				<input
					class="form-checkbox place-self-center"
					type="checkbox"
					name="hasPosOptions"
					bind:checked={user.hasPosOptions}
					disabled={data.roles.find((rol) => rol._id === user.roleId)?.hasPosOptions}
				/>
```

If `role.hasPosOptions`, but not the user; this hunk will incorrectly set `hasPosOptions` in the **user**. This is problematic, since when the role is changed, the user will still have POS access.

roosemberth (changes requested):

src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts
Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:68]: I'll wait for a last review following the last changes, before doing a new QA round

Comment (roosemberth) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:67]: @ithiame Did you forget to stage/push some changes?

I don't see how the last changes address this discussion.

> Now the user hasPosOptions will change if the role is changed

Why do we need to modify the user's `hasPosOptions`? I think we shouldn't update `hasPosOption` unless the admin explicitly sets it in the user's profile.

Best would be to improve the login in line 68 (where this discussion started).

Comment (ithiame) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:67]: @roosemberth If the role of the user changes i think we want to update the `hasPosOptions` that's why i use this line 

`const hasPosOptions = parsed.hasPosOptions || role.hasPosOptions;`

Comment (ithiame) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:67]: > since when the role is changed, the user will still have POS access.

the last change i push was to address this issue. and now if the role change the user will not still have POS access

Comment (roosemberth) [reply src/routes/(app)/admin[[hash=admin_hash]]/arm/user/[id]/+page.server.ts:67]: Consider the following code block:

```ts
const hasPosOptions = parsed.hasPosOptions || role.hasPosOptions;
[...]
await collections.users.updateOne(
	{ _id: user._id },
	{
		$set: {
			login: parsed.login,
			...(parsed.alias && { alias: parsed.alias }),
			recovery: {
				...(parsed.recoveryEmail && { email: parsed.recoveryEmail }),
				...(parsed.recoveryNpub && { npub: parsed.recoveryNpub })
			},
			hasPosOptions,
			disabled: parsed.status === 'disabled',
			roleId: parsed.roleId ?? user.roleId
		}
	}
);
```

Consider a user that does not have `hasPosOptions` set, but `role.hasPosOptions`
is set to true.

- `user.hasPosOptions` is not set (or `false`).
- The admin loads the page, the hasPosOptions checkbox is not ticked.
- When the admin submits the form:
  - `parsed.hasPosOptions` is `false`, `role.hasPosOptions` is true; thus `hasPosOptions` is `true`.
  - We set the user `hasPosOptions`, in this case to `true`.
- The transaction is committed.
- `user.hasPosOptions` is now `true`.

This results in the admin unadvertedly setting `hasPosOptions` in the user.

roosemberth (approved):
Thank you!

roosemberth:
I've removed my approval, I'm working on something different based on this branch and I think I found a regression. Will report back once I've pin it down.

roosemberth:
Yeah, I see the issue now:

- When creating a POS account, the newly created account does not have access to POS features.
- An admin can remove POS access to the POS role.

I think we can simply:

- Set `hasPosOptions` when we create a new account with POS role;
- Prevent unticking attribute `POS` for the POS role in the UI.

roosemberth:
Actually, moving the user to the POS role does not work either. I realize now the thought process in [my previous comment](https://github.com/be-BOP-io-SA/be-BOP/pull/1916#discussion_r2126366120) is flawed, since it does not make sense for the `hasPosOptions` checkbox to be unchecked.

This is rather annoying from a data-consistency point of view, since now we have to ensure the user does not end-up being in the POS role without `hasPosOptions` being unchecked.

Tirodem:
I've just deployed : "point of sale" role should have checkbox "POS" checked and disabled by default :x:

:x: Trying to enable the option on a super admin user leads to this :
image

It's OK with users from other roles.

Tirodem:
It's OK for pos option on super-admin employee.
Checkbox is disabled but not checked for PoS role :x:

Tirodem:
Is functionally good now.
@roosemberth for PR Review (merge if OK)

roosemberth (approved):

Issue #1912, opened 2025-05-19
🦋 Allow be-BOP owner to disable zoom on product picture


Tirodem:
🤷‍♂ 
It's the first time it's requested by one of our pilots so why not.
In merch / layout, add an option (disable by default) to allow to disable zoom on product pictures.

PR #1919, opened 2025-05-21
🦋 Allow be-BOP owner to disable zoom on product picture #1912

ithiame:


roosemberth (approved):

Tirodem:
✅ OK but ⚠️ desktop only.
BTW, mobile need to click, not just hover, so it's good for me.

PR #1920, opened 2025-05-22
Schedule command nostrbot

ithiame:


Tirodem:
rsvp: and m: doesn't seem to work.
Are there missing elements in the PR depending of 1908 ?

ithiame:
To be continued after #1908 done.

Issue #1877, opened 2025-05-06
🦋 Remove Product Price on pop-in when adding to Cart by enabling an option on Admin Config


CypherJack:
TBD

PR #1925, opened 2025-05-23
 Remove Product Price on pop-in when adding to Cart by enabling an op…

ithiame:
…tion on Admin Config #1877


Tirodem:
✅ Functional is OK
Waiting for the PR review

Issue #1291, opened 2024-05-31
🏪 Create a specific interface (restaurant-like) for POS session


Tirodem:
# Context
be-BOP works now with physical scanners for both barcode & QR code and has been texted on POS-like configuration IRL.
But we want to go deeper : we want a optimized interface, made for use with a tactical monitor, and adapted to fast-service-like environments (restauration for example).

# Design
At first, you choose a pool for your ticket.
"Ticket" is a particular cart that be preserved as long as it's not manually cancelled. Contrary to classic carts, a "Ticket" (don't confuse with event ticket) can be created, opened, saved for later and opened again.
"Pool" is manually created in admin interface, and allow you to group Tickets by some common information. Main use-case for this : a table #number in a restaurant.
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/89389fc3-df86-4797-91d2-f497c38986dc)
After selecting the pool, the seller lands on an empty ticket, and has access to "Favoris" products.
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/9d0a95d3-e871-4ca1-8770-5647595ff6e5)
Interface here allows to :
- make direct add to cart (with no preview nor confirmation) by clicking on a Product Widget (products presented here are on a [Productlist=] with the tag POS-favorite)
- display product list from a particular tag that has to be setted in admin interface
- display all articles
The selected list (favorites, all, or user defined ones) is highlighted by a 🟢
Long list are displayed with a pagination system.
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/3543c5b7-10ae-42ec-888f-1648d7a0ae8a)
When products are added to cart by clicking in tile, a ticket-like list is displayed on the left side (I hope soon we'll be able to print it on a ticket printer).
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/bb23a2a9-f225-461b-8849-823a8be15ce8)
An empty ticket cannot be saved ("Save" button has no effect).
While not saved, ticket has the indication `n° tmp`.
A non-saved ticket cannot be paid.
Once ticket is saved, cart cannot expire.
Ticket can be opened later (imagine a restaurant when you can order more stuff while eating, a coffee, a dessert or whatever, this has to be done multiple times).
![image](https://github.com/B2Bitcoin/beBOP/assets/50206014/bf59606a-e7cb-4f2e-89d5-5409e31114d2)
"Tickets" button allow to see non-deleted, non-paid tickets.
[...]

PR #1931, opened 2025-05-26
Improve PoS Touchscreen interface

Tirodem:
Related to #1291

### DIFF
--- a/src/lib/components/ProductWidget/ProductWidgetPOS.svelte
+++ b/src/lib/components/ProductWidget/ProductWidgetPOS.svelte
@@ -52,7 +52,7 @@
 				/>
 			</div>
 			<div class="p-3 flex items-start text-left">
-				<h2 class="text-3xl break-word">{product.name}</h2>
+				<h2 class="text-3xl break-all break-words">{product.name}</h2>
 			</div>
 		</div>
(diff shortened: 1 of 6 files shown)


Tirodem:
@ithiame 2 things here :
image
- [ ] We display more product on list but as you can see we still have room for at least 2 more lines. Beware ! It's not about putting a fix number of lines, it's mean to adapt the number of lines depending of the height of the window.
- [ ] While clicking on "All products", we have an undesired padding (right & bottom). It's not present on other product lists.
Can you check tomorrow please ?

ithiame:
> it's mean to adapt the number of lines depending of the height of the window.

That's what i'm doing. 

Can i have your screen size (the size of pos touch screen?) in case the last commit doesn't work

Tirodem:
Better but still not OK ❌ 
image
image

Tirodem:
@ithiame can you check conflit please ?

Tirodem:
@roosemberth I'll merge as it is and I'll keep #1291 open for a future PR.
In its current state it brings enough improvements of the PoS touch interface for employees.
Can you help with the conflict before merging ?

roosemberth:
@Tirodem rebased.

PR #1939, opened 2025-05-27
🦋 Allow to define default product picture in admin for product withou…

ithiame:
…t picture #1809


Tirodem:
Picture file should not be limited to this :x:
image
We should accept the same picture formats we accept on a product picture upload.

Tirodem:
Works for me ✅ 
Go for PR review

src/routes/(app)/admin[[hash=admin_hash]]/product/default-picture/+page.server.ts
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		const buffer = Buffer.from(await file.arrayBuffer());
+
+		const allowedTypes = ['image/jpeg', 'image/png', 'image/webp'];
+		if (!allowedTypes.includes(file.type)) {
+			throw error(400, 'Unsupported file type.');
+		}
+
+		const assetsDir = join(rootDir, 'src/lib/assets');
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/default-picture/+page.server.ts:25]: The `src` directory does not exist in a production build.

Today, this works by accident because production deployments start by building from source.

`rootDir` also assumes the project metadata exists (ie. `package.json`). So, more things will break once we improve the deployment strategy.

Let's leave this as-is for now.

PR #1945, opened 2025-05-28
📆🦋 Allow to define default event picture in Schedule widget #1810

ithiame:


Tirodem:
Same, with limiting the picture format more than when adding a picture to an event ?
image
:x:

src/lib/components/ScheduleWidget/ScheduleWidgetMain.svelte
```diff
@@ -7,6 +7,7 @@
 	import { addMinutes, isSameDay } from 'date-fns';
 	import IconRssFeed from '../icons/IconRssFeed.svelte';
 	import IcsExport from './IcsExport.svelte';
+	import DEFAULT_EVENT_PICTURE from '$lib/assets/event-default-picture.svg';
```
roosemberth: This requires for the image to be available at compile-time, for the svelte pipeline to process the asset.

It would be cumbersome to have to rebuild the project to change the image. This must be loaded dynamically.

roosemberth (changes requested):

src/lib/assets/event-default-picture.png
roosemberth, line 1: Why do we have to add both a PNG and an SVG version of the same image to the repository source?

Can't we use the build system to prepare the necessary versions of the image?

src/lib/components/ScheduleWidget/ScheduleWidgetMobile.svelte
```diff
@@ -17,6 +18,10 @@
 		pictures.map((picture) => [picture.schedule?.eventSlug, picture])
 	);
 	const { t, locale } = useI18n();
+	let cacheBuster = 0;
+	onMount(() => {
+		cacheBuster = Date.now();
+	});
```
roosemberth: Combined with the snipped in the next hunk,

```
src="/asset/event-default-picture.png?t{cacheBuster}"
```

this pattern is problematic: Using this will force clients to always request and download the image regardless of whether it has changed.

Instead, the resource server (in this case, be-BOP) should return adequate _Cache-Control headers_ when serving the resource. For instance `Cache-Control: no-cache` (or even [`max-age=0; must-revalidate`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#must-revalidate) would be the preferred way.

Using a more adequate cache configuration enables efficient use of cache and save bandwidth. For instance, the user agent (browser) will include an [`If-None-Match` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match) when fetching the resource; this enables the server to respond with a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304) response if the resource has not changed and avoid the transfer.

For this mechanism to work, the [`ETag` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) must be set when serving the resource.

roosemberth (changes requested):

src/lib/assets/event-default-picture.png
ithiame: We can delete the svg it's an old uploaded file.

src/routes/(app)/asset/[filename]/+server.ts
```diff
@@ -1,22 +1,38 @@
 import { rootDir } from '$lib/server/root-dir';
+import { createHash } from 'crypto';
 import fs from 'fs';
 import { join } from 'path';
 
-export async function GET({ params }) {
+export async function GET({ params, request }) {
 	const assetDir = join(rootDir, 'src/lib/assets');
 	const filePath = join(assetDir, params.filename);
```
Comment (roosemberth) [inline src/routes/(app)/asset/[filename]/+server.ts:8]: I just realized we introduced a [path-traversal](https://owasp.org/www-community/attacks/Path_Traversal) vulnerability here.

I'll make a separate PR addressing this.

roosemberth (approved):

roosemberth:

Tirodem:
As long as we have a default picture, we should remove this requirment on product creation :
image
Do that when we'll be able to merge

Issue #1940, opened 2025-05-27
🐛 Unreadable content on CMS management page on dark theme


roosemberth:
On `/admin-[hash]/cms`, in light theme:

![Image](https://github.com/user-attachments/assets/dce253fa-e074-406e-9c87-2326a0528989)

In dark theme:

![Image](https://github.com/user-attachments/assets/235cdd93-e127-419d-bfbb-f9d644a00acb)

The links to edit the pages (among other content) are not visible since they are the same color as the background.

PR #1948, opened 2025-05-28
🐛 Unreadable content on CMS management page on dark theme #1940

ithiame:


Tirodem:
You removed body class from admin so it's readable on white background, wether you're in dark or light mode.
But I asked for this :( "*We should apply the body theme (from theme configurator) to back-office now.*" :x:
The issue was here because we applied body dark theme colors but kept a white background on admin interface.
We need to apply theme body style to the admin, background included

Tirodem:
:x: We need to be able to change the `kdb` :
image

💡 We'll probably need to harmonise the forms in the future, like the ARM on dark background :
image

:x: The `btn-gray` should call the body main cta background color I think, unless we create admin cta parts in the theme selector, which is also possible
image

I'm tempted about managing the admin side on theme manager, but let's put that on the burner for later.
At need change every occurrence of `btn-gray` to `body-mainCTA` and fix the `kbd` issue (for this you can keep `color:back` ; otherwise, you can use `body-secondaryCTA` font and background color as we're sure it'll be contrasted).

roosemberth (approved):
I think this is not a good approach.

I think the best would be to use global styles and design tokens, as recommended per [standard practices](https://developer.atlassian.com/platform/forge/design-tokens-and-theming/); this ensures the UI elements remain consistent without worrying whether we've applied the correct theme settings.

The changes in this patchset demand for each style unit to be manually and religiously added to any link, button, input and kbd. Adding maintenance burden.

Given that:
- This is an admin-only feature;
- You've already spent time implementing it this way;
- I think it's not worth spending more time on this.

I'm approving this patchset.

@Tirodem please take notice of my remarks above, if you agree with my points above, merge; else let me know and I'll we'll discuss a different approach.

Tirodem:
Dully noted, vision about it is shared.
It should be the basis of the future admin backoffice reforge.

Tirodem:
:x: We have an issue with tag selector component :
image

Tirodem:
:x: On some admin page with short content, we have a white background :
image

Tirodem:
:x: Also, some page have light background body color on bottom AND side :
image

roosemberth:
I've rebased the branch to resolve the conflict with the repository's main branch.

@Tirodem I think this is ready to review.

Tirodem:
No because I still have this issue :
image

We need to apply the same style to the dropdown menu as to the input.

Tirodem:
Functional test is good for me

roosemberth (approved):

Issue #1952, opened 2025-05-31
🦋 Made PoS checkout prefill with shop information optional in /admin/pos


Tirodem:
Following what was done here : https://github.com/be-BOP-io-SA/be-BOP/issues/1218
Make the prefill optional through a new option "Do no prefill checkout with shop address on PoS session" (default : disable, to keep the actual behavior).
Context : one of our pilot sells on PoS only by collecting customer address, for warranty purpose. A PoS empoyee is a bit to quick to validate orders and it's bothersome, they have invoices to cancel on shop's name & address.

PR #1956, opened 2025-06-03
🦋 Made PoS checkout prefill with shop information optional in /admin/…

ithiame:
…pos #1952


Tirodem:
Sorry, it works, but I forgot something.
We also need to make some fields required again even in PoS session :
- First Name
- Name
- Address
- Country (pre-fill, select menu)
- Town
- Zipcode

Otherwise it'll save the adress as "COUNTRY" and it does not answer properly the pilot's need.

It should not be a problem as in 99% case the address will be prefilled with shop info.

Tirodem:
Also, with PoS / Employee session, do not prefill address with informations from `personalInfo` collection, otherwise it'll schmuk the stuff.

src/lib/server/runtime-config.ts
```diff
@@ -212,6 +212,7 @@ const baseConfig = {
 	contactModesForceOption: false,
 	posTouchTag: [] as Tag['_id'][],
 	posPrefillTermOfUse: false,
+	notPrefillCheckoutAddress: false,
```
roosemberth: The feature is POS-specific, how about `posDisablePrefillCheckoutAddress`?

src/routes/(app)/checkout/+page.svelte
```diff
@@ -171,6 +171,7 @@
 			? priceInfo.partialPriceWithVat >=
 			  toCurrency(priceInfo.currency, data.physicalCartMinAmount, data.currencies.main)
 			: true;
+	let prefillWithShopInfo = data.roleId === POS_ROLE_ID && !data.notPrefillCheckoutAddress;
```
roosemberth: We also want this to be available to employees with access to POS (not using an account with role POS).

I think we should finish with that PR first: https://github.com/be-BOP-io-SA/be-BOP/pull/1916 and them add this logic here.

src/routes/(app)/checkout/+page.svelte
```diff
@@ -232,7 +233,7 @@
 							class="form-input"
 							name="shipping.lastName"
 							autocomplete="family-name"
-							required={data.roleId !== POS_ROLE_ID}
+							required={data.roleId !== POS_ROLE_ID || data.notPrefillCheckoutAddress}
```
roosemberth: This expression is a bit obtuse and repeated around. I think it would be best to make a variable (e.g. `enforceCheckoutAdress`). This would improve the readability of the code.

Tirodem:
@ithiame for review & conflit resolution

Tirodem:
When I check the box and save, the check is not saved :x:

Tirodem:
Checkbox save is OK ✅ 
"Also, with PoS / Employee session, do not prefill address with informations from personalInfo collection, otherwise it'll schmuk the stuff." is not OK :x:

When logged as a employee :
- if "Pre-fill term of use checkbox in /checkout" is unchecked : do not pre-fill the form except with the country
- if "Pre-fill term of use checkbox in /checkout" is checked : pre-fill with shop informations

For memory, employee informations from /identity are those from `personalInfo`

Tirodem:
🧊  @roosemberth to put on the burner for after Riga for last improvements

PR #1961, opened 2025-06-04
🦋 Improve catalog option in nostr-bot #1842

ithiame:
https://github.com/be-BOP-io-SA/be-BOP/pull/1908


Tirodem:
🧊 to schedule after Riga

PR #1976, opened 2025-06-12
🦋 Reorder, duplicate and remove product variations on the admin product page

### LINKED ISSUE #1702: 📅 Improve product variation management on admin product page
# Overview

To make the product variation feature easier to use, especially with incoming https://github.com/B2Bitcoin/beBOP/issues/1474 and https://github.com/B2Bitcoin/beBOP/issues/1687, we need to do some improvements on admin product page for variations.

# Description

## Allow product variation deletion & cloning

Currently :
- deleting a variation is done by deleting the inputs
- adding a variation is done by clicking on "Add variation"

Image

We want to :
- allow a one-click deletion of a variation through a ➖ icon
- allow to duplicate a variation line (minus the "Value" input) through a ➕ icon
  - that way, it'll help to be sure to have a variation with the same "Name", to avoid typo and create an underisable new variation family

While we're at it, remove the "Name", "Value" and "Price difference" on each line (only put them on the first line) :

Image
When clicking on ➕ , the new line must not be created at the end of the list, but just after the duplicated line.

If "Name" field is empty, set ➕ opacity to 0.5 and disable it.

Image

⚠ : Consider the next point that'll change the "Value" format ; in case of variation duplication, it should also take that in account.

## Allow to change variation list order

Image
Through 🔺 and 🔻 (with `filter: opacity(0.5)` when disabled for 1st and last line), the admin will be able to change the order of the lines.

## Checkbox restriction

Currently we cannot check "Product has light variations (no stock difference)" if "This is a standalone product" is unchecked :

Image

Image

We need to change that and make like "This is a pay-what-you-want product" where checking the option will automatically check the "This is a standalone product" option :

Image

ithiame:
Products with variations could only have them listed in the order they were
first entered. A new variation always landed at the end of the list, and the
only way to place one between two others was to delete everything after it and
type it all again.

Variations can now be arranged from the product page. Each row carries controls
to move it up or down, to duplicate it — the copy appears directly under the
original, ready to be renamed — and to remove it. The order set here is the
order customers see on the product page.

A product that has variations is now standalone by default, so its stock is
counted per variation rather than shared with a parent item. Existing products
with variations are migrated on startup.

Fixes #1702

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -235,20 +235,27 @@
 			}
 
 			const seen = new Set<string>();
-			for (const [i, value] of variationLabelsValues.entries()) {
-				const key = JSON.stringify(
-					`${variationLabelsNames[i] || product.variations?.[i].name}, ${
-						value || product.variations?.[i].value
-					}`
-				).toLowerCase();
+			const duplicateIndex = Array.from({
+				length: Math.max(product.variations?.length || 0, variationLabelsValues.length)
+			}).findIndex((_, i) => {
+				const name = product.variations?.[i]?.name || variationLabelsNames[i];
+				const value = product.variations?.[i]?.value || variationLabelsValues[i] || '';
 
-				if (seen.has(key)) {
-					variationInput[i].setCustomValidity(`Duplicate variations found ${key}`);
-					variationInput[i].reportValidity();
-					event.preventDefault();
-					return;
+				if (!name || !value) {
+					return false;
 				}
+
+				const key = `${name},${value}`.toLowerCase();
+				const isDuplicate = seen.has(key);
 				seen.add(key);
+				return isDuplicate;
+			});
+
+			if (duplicateIndex !== -1) {
+				variationInput[duplicateIndex]?.setCustomValidity('Duplicate variation found');
+				variationInput[duplicateIndex]?.reportValidity();
+				event.preventDefault();
+				return;
 			}
 			if (!duplicateFromId && isNew) {
 				if (!files) {
@@ -369,6 +376,10 @@
 		product.standalone = true;
 	}
 
+	$: if (product.hasVariations) {
+		product.standalone = true;
+	}
+
 	$: if (product.free) {
 		allowDeposit = false;
 	}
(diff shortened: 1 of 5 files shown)


Tirodem:
Fine for me.
@roosemberth for PR

roosemberth:
@ithiame Can you please rebase?

roosemberth:
When I add the following product variations to an existing product:

image

When I try to update the product, I get the following error:

image

src/lib/components/ProductForm.svelte
```diff
@@ -299,6 +299,41 @@
...
+			variationLabelsNames.push(variation.name);
+			variationLabelsValues.push(variation.value);
+		} else {
+			const newVariationValue = crypto.randomUUID();
+			if (product.variationLabels) {
+				product.variationLabels.values[variation.name][newVariationValue] =
+					product.variationLabels.values[variation.name][variation.value];
+			}
+			product.variations?.splice(index + 1, 0, { ...variation, value: newVariationValue });
+		}
```
roosemberth: I don't understand what's the relationship between `variationLabelsValues, variationLabelsNames` and `product.variationLabels`.

if anything I'd expect the two former to be lists of the keys and values of the later one. But this code reveals that they don't even have the same shape. What's the UUID seemingly being used as a key?

src/lib/components/ProductForm.svelte
```diff
@@ -629,20 +664,24 @@
 				type="checkbox"
 				bind:checked={product.hasVariations}
 				name="hasVariations"
-				disabled={!product.standalone}
 			/>
 			Product has light variations (no stock difference)
 		</label>
 		{#if product.hasVariations}
 			{#each [...(product.variations || []), ...Array(variationLines).fill( { name: '', value: '', price: 0 } )].slice(0, variationLines) as variation, i}
```
roosemberth: I know this was not introduced in this commit, but please clean this up... You can bind this expression to a symbol explaining right bellow the `{#if product.hasVariations}` explaining what it is.

src/lib/components/ProductForm.svelte
```diff
@@ -629,20 +664,24 @@
...
 				<div class="flex gap-4">
 					{#if variation.name && variation.value}
+						<button
+							class={i === 0 ? 'mt-8' : ''}
```
roosemberth: Why put an extra margin specifically in the first element of any branch?
Wouldn't it be easier to read if it were set in the parent container?

src/lib/components/ProductForm.svelte
```diff
@@ -652,19 +691,39 @@
...
+							<input
 								type="text"
 								name="variationLabels.values[{variation.name}][{variation.value}]"
 								class="form-input"
 								value={product.variationLabels?.values[variation.name]?.[variation.value]}
 								bind:this={variationInput[i]}
-								on:input={() => variationInput[i]?.setCustomValidity('')}
+								on:input={() => (
+									variationInput[i]?.setCustomValidity(''), (variationInput[i] = variationInput[i])
+								)}
```
roosemberth: First, why does this return a tuple instead of using a regular function body?

```suggestion
								on:input={() => {
									variationInput[i]?.setCustomValidity('');
									variationInput[i] = variationInput[i];
								}}
```

Second, what's the purpose of `variationInput[i] = variationInput[i]`?

src/lib/components/ProductForm.svelte
```diff
@@ -652,19 +691,39 @@
...
 								required={!!product.variationLabels?.names[variation.name]}
 							/>
 						</label>
 					{:else}
+						<button
+							class="{i === 0 ? 'mt-8' : ''} {!variationLabelsNames[i] && !variationLabelsValues[i]
```
roosemberth: Please bind the expression `!variationLabelsNames[i] && !variationLabelsValues[i]` to a variable after `{:else}` explaining what this represents. There are more instances of the same expression and it's really hard to understand.

Tirodem:
@roosemberth That shit happens from time to time even on main. Let me recheck if with this version.

roosemberth (changes requested):

Tirodem:
Ok, seen.
`variationLabels.values.undefined: Expected object, received array` happens when you have one empty entry on variations form.
We can imagine to delete empty entry before submitting the form.
From what I see, it does not depend of this PR.

Tirodem:
https://github.com/be-BOP-io-SA/be-BOP/issues/1994 as side-dish

Tirodem:
🧊 after Riga

roosemberth (changes requested):
Can you please squash both commits? (The first commit is incomplete without the second one; and the second one fixes what's in the first.) If you want to preserve the authorship information, see [here](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors).

Can you also explain in the commit message why the migration is required?

The code is much much better (thank you!) than before, but the expressions in the `<input value=...` could be refactored for clarity (cascading ternary expressions is utterly unreadable.)

roosemberth (approved):
Thank you!

roosemberth:
@Tirodem please merge when convenient.

Tirodem:
:x: Need peer testing next week (no emergency on this one), I have various issues :
- moving a variation horizontally triggers some undesired "Duplicate variation found"
- duplicating a variation A then changing its name also change variation A name, then after save, if I delete variation A', variation A is also deleted

It was on an existing product with existing variations.

gntw:
@Tirodem  , I made the approach more reliable, now it works well

Tirodem:
➕ should create the duplicated variation just after the original one, not at the end of the list, otherwise it loses a bit of interest in term of UX.
(btw it's not urgent)

Tirodem:
Otherwhise, it's works like a charm 👌

gntw:
> ➕ should create the duplicated variation just after the original one, not at the end of the list, otherwise it loses a bit of interest in term of UX.
> (btw it's not urgent)

DONE , @Tirodem

Tirodem:
image

@gntw fix conflict then go merge

roosemberth (approved):
I reviewed the changes (not the changeset) with `range-diff` 👍

Tirodem:
@gntw beware conflict, also for incoming #2246 , and future #2065 (that'll need this ticket)

Tirodem:
@gntw can you refacto this ticket to be compliant with the newly merged admin product page please ?

gntw:
@Tirodem its done, works good at my local

Tirodem:
Here we go again :

image

@roosemberth for final review if you want.

roosemberth:
LGTM.

roosemberth:
The CI is taking a bit longer, but merge when convenient.

Tirodem (approved):

Issue #1921, opened 2025-05-23
🦋 Allow to mute pending / expiring order notifications for be-BOP Owner


Tirodem:
TBD
To avoid spam and noise.
That way the owner will be notified of paid / completed orders only.
Take caution of pending bank transfer orders that do not expire.

Tirodem:
@ithiame how I see this :
- let's list every notification related to order (those that are sent with the activation of the option "Copy order emails to mail@provider.com (set in identity section)" (I think we can use the email templates related to orders as a workbase to list them)
- list them in Admin > Node managements > Emails
- add 2 columns with checkboxes : "Send to user" / "Send copy to admin"
- only send email if the option is enable

For now we'll send copy email to admin only if "Copy order emails to blablabla" is enable in Admin > Settings General **AND** if the option is checked in Admin > Node managements > Emails.
Maybe in the future we'll remove the "Copy order emails to blablabla" option.
BTW, also, move this option to Admin > Node managements > Emails.

If this works, we'll do a proper management for other non-templated emails related to admin (challenge and leaderboard update, for example).

In the future we'll do the same for Nostr.

PR #1979, opened 2025-06-13
🦋 Allow to mute pending / expiring order notifications for be-BOP Own…

ithiame:
…er #1921

### DIFF
--- a/src/lib/server/locks/order-notifications.ts
+++ b/src/lib/server/locks/order-notifications.ts
@@ -12,7 +12,12 @@ import { ORIGIN } from '$env/static/private';
 import { Kind } from 'nostr-tools';
 import { toBitcoins } from '$lib/utils/toBitcoins';
 import { getUnixTime, subHours } from 'date-fns';
-import { refreshPromise, type EmailTemplateKey, runtimeConfig } from '../runtime-config';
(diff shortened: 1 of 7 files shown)


Tirodem:
After deploy :
image
Few remarks :
- :x: every email for user should be enabled after deploy
- :x: if "Copy order emails to christophe@be-bop.io (set in admin/identity section)" is checked, every email for admin should be enabled after deploy
- :x: Some admin-only emails are missing (leaderboard & challenge change notification, for example)
- one "update" CTA per line seems a bit overkill, we should have only one CTA for global save, or, if we keep one CTA per line, a disk option (like /admin/arm) should do the trick

Tirodem:
When I uncheck one or many options like this then save :
image

Here's what I have after save :x:
image

Tirodem:
Checkbox is fixed but :
- :x: when "copy email to admin" main option is unchecked, no email is sent to admin
- :x: when "copy email to admin" main option is checked, all sub-options for admin notifications are enabled (if you try to uncheck some, on save, they are checked again)

Tirodem:
@roosemberth considering current behavior, we shall schedule a peer-testing session
---> 🧊 Riga

Issue #1932, opened 2025-05-26
🦋 Increment/Decrement Product quantity on POS Touch


CypherJack:
When clicking on a Ticket line in POS Touch, user can already add comment thanks to a pop-in.
Extend this feature to allow for quick quantity decrement if done a mistake or client adjusting their order. While providing an Increment Button.
Also having the possibility to change product's quantity using a dropdown menu.

![Image](https://github.com/user-attachments/assets/49ca61d0-96dd-430d-9d36-28b3fe9b5fc6)

PR #1981, opened 2025-06-16
🦋 Increment/Decrement Product quantity on POS Touch #1932

ithiame:


Tirodem:
"Comment" field is always pre-filled with the previous command (scenario : add article 1, comment it, add article B, open qty/command popin > the input should be empty or pre-filled with the current comment, it shouldn't be pre-filled with the comment on another product) :x:
image

Tirodem:
"+" and "-" CTA are a bit laggy, because the quantity change is done directly and impact calculs and prices :x:
It shouldn't, best way here is to make those CTA update the qty on the select menu, and apply change only if the "Save" button is clicked.
Otherwise it's too laggy to be used on real life.

Tirodem:
On the contrary, change on product quantity through select menu is saved but not applied to the ticket :x:
image
(I've made change, saved then opened it again, it's 8 on select menu qty but 5 on ticket and cart)

Tirodem:
Also "Communiquer la ligne article" should be in fact "Commenter la ligne article" ou "Ajouter un commentaire"

Tirodem:
Same issues :
- "+" and "-" cta are to laggy ❌ 
- qty select menu does not work ❌ 
- comment is persistent on input field ❌ 

@ithiame I think we need a live demo to show you the issues.

Tirodem:
- "+" and "-" ctas work now ✅ 
- qty select menu does not work ❌
- comment is no more persistent ✅

Tirodem:
I still have issues with qty select menu that is not sync with qty amount and +/- CTAs :
image
@roosemberth 🧊 after Riga

Tirodem:
image

on dev-bootik2 with "Röstis Chaux-de-Fonniers"
Here's the admin config :

image

So :
- :x: there's no reason to trigger this error for a qty of 9 (10 I would understand, 9 I don't)
- 🦋 : we should plan an error management on PoS touchscreen for stock error / max qty error
  - 🦋 "+" CTA should be disabled when clicking on it while editing the quantity
  - 🦋 Product background should be lighter when adding more is impossible, and onclick, we display a warning "we cannot add more"
    - 💡 We might even plan to add a warning when available qty is inferior to a certain treshold
    - 💡 Shouldn't we disable the max qty per order of a product when on PoS touchscreen ?

Issue #1938, opened 2025-05-27
🦋 On customer PoS screen, display order url through QR code after payment to allow customer to collect order for history (optional)


Tirodem:
Currently, when a PoS order is paid, here's what we have :
![Image](https://github.com/user-attachments/assets/68382ce0-7635-4c86-92b1-85258c47c753)
Then it goes to the welcome page.
Idea is here :
- add an option in Admin > Settings > Pos "Display order QR code on PoS after payment" (default : disabled)
  - if checked :
    - add an input to allow to set a time before redirecting to home (default : 10s)
    - add an option "Display the customer CTA on PoS after payment" (default : disabled)
    - add an option "Remove be-BOP logo from POS after payment QR code" (default : disabled)
- when an order is paid, display this page and stay on it for the setted time (default : 10s)

![Image](https://github.com/user-attachments/assets/90af7a53-82ac-476b-bd06-7e1dae127880)
Instead of "Scan me" :
- if "Remove be-BOP logo from POS QR after payment code" is enabled, display plain QR code
- if "Remove be-BOP logo from POS QR after payment code" is disabled, display be-BOP "B" logo

The CTA is only here if "Display the customer CTA on PoS after payment" is enabled.
In that case, when customer click on CTA, the PoS goes to the waiting / welcome page without waiting for the timeout.

PR #1982, opened 2025-06-17
🦋 On customer PoS screen, display order url through QR code after pay…

ithiame:
…ment to allow customer to collect order for history (optional) #1938

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts
@@ -2,7 +2,9 @@ import { adminPrefix } from '$lib/server/admin';
 import { collections } from '$lib/server/database';
 import { runtimeConfig } from '$lib/server/runtime-config';
 import type { Tag } from '$lib/types/Tag';
+import { set } from '$lib/utils/set';
 import { error, redirect } from '@sveltejs/kit';
+import type { JsonObject } from 'type-fest';
 import { z } from 'zod';
 
 export const load = async ({}) => {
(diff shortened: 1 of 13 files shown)


Tirodem:
Works finely for me ✅ 
@roosemberth for PR review, merge if OK.

src/routes/(app)/order/[id]/qrcode/+server.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ORIGIN } from '$env/static/private';
+
+const bebopLogoSvg = building ? '' : readFileSync(join(rootDir, 'assets/bebop-b.svg'), 'utf-8');
+
+const BEBOP_LOGO_RATIO = 0.2;
+
+export async function GET({ params, url }) {
+	const address = `${ORIGIN}/order/${params.id}`;
+
+	let qrcodeString = (await qrcode.toString(address, { type: 'svg' })).trim();
```
Comment (roosemberth) [inline src/routes/(app)/order/[id]/qrcode/+server.ts]:
```suggestion
	let qrcodeString = (await qrcode.toString(address, { type: 'svg' }));
```

Why do you need to trim here? That's a bit odd.

src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts
```diff
@@ -28,12 +32,39 @@ export const actions = {
...
-				posPrefillTermOfUse: formData.get('posPrefillTermOfUse')
+				posPrefillTermOfUse: formData.get('posPrefillTermOfUse'),
+				posDisplayOrderQrAfterPayment: formData.get('posDisplayOrderQrAfterPayment')
 			});
+		const json: JsonObject = {};
+
+		for (const [key, value] of formData) {
+			set(json, key, value);
+		}
+		const parsed = z
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/pos/+page.server.ts]: It's confusing to see two "parses" in this code, I think this can be fixed by improving the name.

```suggestion
		const parsedOptsForPosQrCodeAfterPayment = z
```

Since we're parsing from a JSON now, wouldn't it also make sense to remove the `forData.get(...)` calls above and use the same `json` object to parse all options?

PR #1983, opened 2025-06-18
🆕 Customer touch interface #1805

ithiame:


Tirodem:
- Shop logo must be `object-fit: contain`
- "Menu", "Burgers", "Salades" & co should not be SVG onboarded on code, they'll be pictures setted on admin > settings > CTI
- in /admin/cti, category entry picture upload does not work properly
- "Ajouter" CTA on product is not great...
- when the design we'll be finished, we'll remove a lot of padding here and here

Despite that, nice progression 👌

Tirodem:
Thanks @ithiame for your last push on be-BOP !
See ya later I hope.

Final review on latest version :
- [ ] theme manager integration
- [ ] CMS pages integration (ex: /touch-intro-1)
- [ ] "Sur place" / "A emporter" info to stock on database
  - [ ] note that we can be a bit smart and consider something to work in the future with a click and collect option from web sales
  - [x] so, we can consider to track the order origin (currently we log the employee session, otherwise system, but we can track the interface, like "web" for standard eshop UI, "nostr" for nostr-bot, "pos" -web + employee login-, "waiter" -order through /pos/touch-, and "cti" -through the CTI interface) on those databases informations : `carts.origin` and `orders.origin`
    - [ ] done for carts but not for the orders :x:
    - [ ] done for onsite and take-away, not the others
  - [x] and in the same way, we can consider `carts.target` and `orders.target` : target would be "web" for web order, "customer" for web order with delivery fees, "onsite" for PoS order with free shipping, while CTI orders would either be "eat-in" or "take-away" ; and in the future we'll have "clickncollect" for web order with delivery but on shop
    - [ ] done for carts but not for the orders :x:
    - [ ] done for "touch-selfcare" but not the others
    - [ ] not every cart can carry a `carts.target` information as this will be processed depending of the checkout options ; it'll be usefull especially for "eat-in" and "take-away" I think ; when `carts.target` is filled, it sets `orders.target` with the same value ; otherwise, `orders.target` will be processed on validated checkout when creating the order object
- [ ] Input email / npub validity check > 🧊 iceberg as we did something with  receipt notification after payment
  - [ ] npub / QR code camera integration > 🧊 iceberg as we did something with nostr DM OTP for receipt notification after payment
- [ ] timeout custom setting on /admin/CTI
- [ ] display should not be bigger than screen height
image
- [ ] removal of "Recherche" in left menu (it'll be a CMS page with a search widget)
- [ ] "Enable Customer Touch Interface" on /admin/CTI should disable the CTI when unchecked
- [ ] on /admin/cti allow upload of .svg (it's possible if you choose "all files")
- [ ] every category list target link is to /pos/customer-touch/list/home
- [ ] the product list widget that was done by ithiame is interesting. it's supposedly CMS content, but we should make it a [ProductList=] widget or something like that. I'll think about it.
- [ ] product page on CTI should have another "Add to cart" CTA (a rounded one at least, and maybe with another position)
  - [ ] on the same topic, note that the position differs depending of the product type (here's a non-standalone product)
image
- [ ] we should change "Ajouter le premier produit" CTA because it's confusing when landing on a product page for the first time on CTI
image
- [ ] on the same topic, this CTA doesn't change if the cart includes free products only (it only changes when cart price > 0)
- [ ] we should include PWYW fields on corresponding product page CTI version
image
- [ ] the quantity selector should be centered
image
- [ ] adding a product to cart should give some feedback, maybe by disabling the "add to cart" CTA for a few seconds, or displaying a "✅Product added" label for a few seconds
- [ ] the quantity selector can go more than 10 and then it has some buggy behavior
  - [ ] we should sync with standalone / max qty per order checks also
- [ ] booking management interface on CTI product page
- [ ] there's 2 "enable CTI" on admin/CTI, I'm pretty sure the second should be "Enable customer login screen at begin of the order journey"
image

- [x] header should be center (especially the shop logo)
- [ ] we should add an option to have a CTI specific logo, maybe ? to considerate ⚖️ 
- [ ] to evaluate : on each payment method, in case of failure, you either should be able to 1/ renew payment attempt 2/ choose another payment method (currently impossible as it converts the cart as order and the "back" will lead to an empty cart) 3/ pay at the counter with an order number (and a QR code display for URL pick) as a fallback
- [ ] on credit card payment (sum up, stripe), only display QR Code (remove the clickable SVG that leads to the payment form, users should not fill their credit card information on the CTI device)
- [ ] remove clickable action on payment QR codes (on-chain, lightning, credit card) because it'll try to either open a CB form (see earlier) or open a bitcoin / lightning wallet on the CTI device (that is not expected / acceptable)
- [ ] we should add in admin/cti a list of enabled payments, similar to the one in admin / settings / generals, but specifically for CTI
- [ ] we also need to check if payment method exclusion on product is taken in account for CTI journey
- [ ] allow to collect the receipt as a plain QR code with the order URL through a new CTA (UI to do 🚧)
- [ ] "Ask for help" should send a notification through either email or nostr to an employee account (with its account contact address setted in ARM").
  - [ ] The message should be "_There's an issue with the CTI logged as {CTI employee session alias}, the customer is calling for help, please go to check what happens_"
  - [ ] We need a referent association interface. For now, we can imagine an unique one. Here's the UI in admin/cti (displayed option = employee alias (role alias) : 
image

- [x] on email notification form (after payment), the CTA should be in another style (currently the CTA seems disabled)
- [ ] Through CTI, "Imprimer" currently lead on a web receipt PDF, it should generate the ticket similar to the one for thermal impression (ticket-format)
- [ ] we should add an option on /admin/CTI to hide the day/light theme selector
- [ ] depending of the browser / OS, the language selector doesn't display the flag emojis, we need to rethink that
- [ ] at the top left we currently display "HH:mm DD/MM/AAAA", we should display a "DD/MM/AAAA`<br>`HH:MM"
- [ ] standalone products cannot be removed from cart as we only do that through selector : we need to add a "delete product line icon"
image

Tirodem:
# Nostr connexion

We have a cameraless device that we want to use as a pilot for CTI.

Nostr connexion is out of question without short address management (you cannot type your npub on a public device because it's too long / complicated -trust someone who had to type an on-chain address on a BTC ATM after 3 pints of beer in order to be refunded 🤡)

So, here's the plan here :
- on checkout, be-BOP will display a QR Code of its nostr-bot address and a unique temporary code
- customer will scan the QR code to access the be-BOP nostr-bot
- customer will DM the nostr-bot with the temporary code
- be-BOP will process the message and fill the session with the customer npub as contact target

## 1/ Change notification / identification journey

image

This screen should not be used for notification, but for customer "login" (false login, as the session keeps the employee / PoS credential) in the future, when we'll introduce customer account & discount.

We cannot collect email so early because contact information is not related to a cart but related to a session and / or to an order, and at this point, no cart neither order exists.

So, let's add this option in /admin/cti :
image

Default : disabled
- when disabled : /pos/customer-touch/start is bypassed, so /pos/customer-touch/welcome goes directly to /pos/customer-touch/list/home
- when enabled : current behavior

We'll improve the customer "login" behavior in the near future.

## 2/ Add a notification address screen on checkout

Thing here is that, contrary to a web or shop order, we don't want to send an email to the customer for pending payment.
So, right after this page ( /pos/customer-touch/order/{id} ), let's display this when "Enable customer login screen at begin of the order journey" is disabled (if enabled, in the future, we'll take the same information as the customer "login") :
image

Click on "Envoyer" :
- should be displayed as long as "Enable customer login screen at begin of the order journey" is disabled
- should lead to this page : /pos/customer-touch/order/{id}/notify

image

- email should redirect to this :
image

After email is sent, send back to this :
image

- nostr should redirect to this :
image

On QR Code, content is `nostr:npub`
After code is received on nostr DM and nostr message is sent back, send back to this :
image

- "Pas de reçu, merci" should send back to /pos/customer-touch/welcome (fully processed order, next customer's round)
Once validated

## 3/ nostr code control

On Nostr side, here's the experience :

- with a valid code :
image

- with an invalid code : no answer

The code should be available for 60 secondes, and used only once.
It's obviously unique for each order.

Tirodem:
Also, we can add an option in /admin/cti :
[ ] Replace print receipt by e-receipt option

So, when enabled, replace the "Imprimer" CTA after payment by "Get receipt on phone"
That'll generate a QR code with the order receipt URL.

Maybe in the future we'll allow the two options at the same time.

roosemberth:
@Tirodem: As discussed, the branch has been updated with the changes to enable users to receive a link to their order via Nostr.

I've also rebased the branch atop main.

Tirodem:
Nostr notification without camera is working ✅ 

Small improvements :
- [x] timeout after sent receipt should be 10-15 seconds
- [x] when notification is "Envoyé ✅" ou "Imprimé ✅ " :
  - [x] updated CTA should not be clickable
  - [x] "Non merci" CTA should not be displayed

Regression :
- [x] When adding a product to cart, the process to cart CTA is not updated in real time anymore (we need to refresh to see the cart amount) :x:

PR #1984, opened 2025-06-23
🦋 Add Tap-to-pay for automatic reconcilliation of POS external payments

### LINKED ISSUE #1134: 🦋 Reconciliate POS payment through tap-to-pay / TPE app automatically with pending payment / order through API
# Admin
Image
- Set the CB provider used for tay-to-pay (ex : Stripe) or external TPE use (ex : Sum Up) with a menu
  - only include setted providers (admin > payment settings)
  - include a "Not used" option (default)
- Allow to fill mobile application URL (it'll allow to open directly the app from be-BOP order page)
  - optional, if not filled, do not put any link on the CTA

# Order

The idea here is to allow to open external payment app from be-BOP, then, once the payment is done, make an automatic sync between the be-BOP pending order and the paid transaction on the said payment provider.

The action is possible through a new CTA that'll replace the input form for PoS payment.

Rules are :
- only one PoS be-BOP pending order at a time can provide the option
- to avoid confusion and mumbo-jumbo with simultaneous orders, a 2nd PoS pending be-BOP order will have to wait for the first one to be synced with success or canceled
  - here we speak in fact about pending payment (in case of order with deposit)
  - when the 1st pending order / payment is synced OK, it's possible to use the CTA on the 2nd one, then the 3rd, etc
- the pending be-BOP payment will be synced as OK once we found on payment provider a transaction :
  - status is OK
  - amount is **exactly** equal to the expected amount of the pending payment
  - transaction time is > to pending payment creation time
- If the payment is cancelled, we do not try to sync anymore

## 1st pending PoS order
Image
When clicking on "Tap to pay", the page becomes "1st pending PoS order - sync in progress".
If a mobile application URL was filled, clicking on the CTA also calls the corresponding resource (an open the billing app, either Paynow for Stripe or Sum Up).

## 1st pending PoS order - sync in progress
Image

## Sync done
Image

## 2nd pending PoS order - autosync disabled
Image
Once the 1st sync is done, the "Tap to pay" CTA is displayed, and the warning disappears.
Note : the earlier rule "*transaction time is > to pending payment creation time*" should be lightly changed to replace "*pending payment creation time*" by "*sync begin time for pending payment*", that would be equal to the time we click on the CTA "Tap to pay".

roosemberth:
See commit message.

Closes #1134.


roosemberth:

I'd rather we don't squash all commits (instead use "Rebase and merge").

src/lib/server/locks/order-lock.ts
```diff
@@ -12,16 +12,56 @@ import { refreshPromise, runtimeConfig, runtimeConfigUpdatedAt } from '../runtim
...
 
+async function findStripePaymentMatch(
+	order: Order,
+	paymentId: ObjectId
+): Promise<string | undefined> {
+	const payment = order.payments.find((p) => p._id === paymentId);
+	if (!payment) {
+		throw new Error(`Payment ${paymentId} or order ${order._id} not found`);
+	} else if (!payment.posTapToPay) {
+		return; // No tap-to-pay for this payment
```
roosemberth: The function name does not predicate a Tap-to-pay is required in the order.

Since the check is already done before calling this function, these lines can simply be removed.

src/lib/server/locks/order-lock.ts
```diff
@@ -409,9 +449,54 @@ async function maintainOrders() {
 						} catch (err) {
 							console.error(inspect(err, { depth: 10 }));
 						}
+					case 'point-of-sale':
+						try {
+							if (payment.posTapToPay && payment.posTapToPay.expiresAt > new Date()) {
+								switch (payment.processor) {
+									case 'stripe':
+										if (!runtimeConfig.stripe.secretKey) {
```
roosemberth: Use `isStripeEnabled` instead.

Tirodem:
Works with Stripe ✅ 
I create a new ticket for Sum Up in the future

PR #1986, opened 2025-06-24
🦋 Enable admin to add missing discounts to existing subscriptions

### LINKED ISSUE #1942: 🦋 When you create a discount on a subscription, add a option to make it retroactive on active subscriptions done before discount
Context :
While doing a demo, I took a subscription, then create a discount : the discount (next gen) wasn't applied. It's because it's not retroactive on the subscriptions.paid.freeProductById collection array.

Rules :
If I have a discount next gen related to a subscription (1 to n) and valid dates
When I click on "Sync subscriptions" CTA
Then user having active matching subscription in subscriptions.paid and having no entrie in freeProductsById should be synced in order to benefit from free products

We use a "Sync subscriptions" CTA and we don't use "Save" button because the operation can be -I think- quite long on be-BOP with a lot of subscribers so it feels better to me to split discount save process and retroactive alimentation of freeProductsById process.

Note :
We can also do that with new free products.
For example, Discount next gen is this :
- Product A, 3 free
- Product B, 4 free
- Product C, 8 free

Customer subscriptions.paid.freeProductsById is this :
- Product A, 3 free
- Product B, 4 free
- Product C, 5 free (3 used)

If the be-BOP owner add a new product on the discount for it to become like this :
- Product A, 3 free
- Product B, 4 free
- Product C, 8 free
- Product D, 2 free

Then customer subscriptions.paid.freeProductsById should become this :
Customer subscriptions.paid.freeProductsById is this :
- Product A, 3 free
- Product B, 4 free
- Product C, 5 free (3 used)
- Product D, 2 free
=> as Product D entry does not exist, we create it.

Note that if we change the free amount of an existing product, we should not update the quantity in subscriptions.paid.freeProductById (otherwise, if the customer already use some of his / her free products, it'll give him / her a fresh new batch).
So we only add subscriptions.paid.freeProductById entry if it doesn't exist, or we create subscriptions.paid.freeProductById if it doesn't exist. We don't update amounts.

Design :
Image

roosemberth:
Please see commit messages. Closes #1942

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -9,6 +9,7 @@ import {
 import { ClientSession, ObjectId, type WithId } from 'mongodb';
 import { collections, withTransaction } from './database';
 import {
+	Duration,
 	add,
 	addHours,
 	addMinutes,
@@ -73,6 +74,7 @@ import { isEmptyObject } from '$lib/utils/is-empty-object';
 import { binaryFindAround } from '$lib/utils/binary-find';
 import { toZonedTime } from 'date-fns-tz';
 import { isSwissBitcoinPayConfigured, sbpCreateCheckout } from './swiss-bitcoin-pay';
+import { PaidSubscription } from '$lib/types/PaidSubscription';
 
 async function generateOrderNumber(): Promise<number> {
 	const res = await collections.runtimeConfig.findOneAndUpdate(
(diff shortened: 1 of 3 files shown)


roosemberth:

Tirodem:
Works 👌

Tirodem (approved):

Issue #1994, opened 2025-06-26
🐛 "variationLabels.values.undefined: Expected object, received array" when product form has one empty variation entry


Tirodem:
When creating a product with 1 empty variation entry :

![Image](https://github.com/user-attachments/assets/1b99207c-6189-4ffe-bf4d-9f4bf53535fc)

Changes are not saved because of this error message :

When filling the empty variation entry, it's OK.

2 options here :
- highlight the empty variation input and put the error message here : possible but as we put 2 variations entry on product when checking the option, and as we cannot remove variation entry until product is saved, if won't be a UX-compliant option :x:
- ignore the variation empty entry while saving (discard the line then save) < that's my favorite

roosemberth:
Closed by https://github.com/be-BOP-io-SA/be-BOP/pull/1995.

PR #1995, opened 2025-06-26
🐛 'variationLabels.values.undefined: Expected object, received array'…

ithiame:
… when product form has one empty variation entry #1994

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -125,6 +125,18 @@
 		// Need to load here, or for some reason, some inputs disappear afterwards
 		const formData = new FormData(formElement);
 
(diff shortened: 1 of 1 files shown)


Tirodem:
Functional tests are OK ✅ 
@roosemberth for review, merge if OK
(note that if change are to be done, you'll have to make them directly)

roosemberth (approved):
The patch is good, thank you.

In the future, please separate argument sanitation from data processing.

Specifically, here it's really hard to understand why the form is not the argument of the `checkForm` function and why the form is coming from a completely unrelated place (as opposed to the argument). It is also hard to understand why function `checkForm` modifies the form and finally why function `checkForm` does not check the form, but a projection of the form.

Issue #1518, opened 2024-10-19
🦋 Manage failed status on credit card orders (sum up, stripe)


Tirodem:
Some customers on live pilots complain about the "expired" order statut when, in fact, the transaction is refused by either Stripe or Sum Up (car be the same thing with Paypal).
They believe they went to slow on form, create another order, and 3-4 order are created and go "expired" because 3rd payment party always answers "Fail" for their card.
We need to improve the process to make a difference between a true expired order and an order when the 3rd payment party say "fail" for the payment mean.

Tirodem:
@coyotte508 : ping for estimation (time + complexity) next week please

coyotte508:
I guess 1 day

coyotte508:
Is is this still relevant or is this done?

roosemberth:
Technical implementation:

- Enhance order updated to try and retrieve and persist failure reason;
- Display failure reason (if present) when displaying payment details.

Development:

Feature: 0.5d
(No setup, we'll hash out the last issues with Tirodem)
Integration: 0.5d with QA

PR #2001, opened 2025-07-10
🦋 Display refusal reason when processing card payments

roosemberth:
Closes https://github.com/be-BOP-io-SA/be-BOP/issues/1518.

This MR is draft until support for Sumup and Paypal has been added.


roosemberth:

Please don't forget to do a fast-forward-only with merge-commit merge (as discussed).

Tirodem:
So is this stripe only or also sum up ?
From what I found here ( https://docs.stripe.com/testing ), Stripe processes directly the payment error on their form :
image
Sum Up being Sum Up, I want to be sure it's covered before setting it again.

roosemberth:
> So is this stripe only or also sum up ? From what I found here ( https://docs.stripe.com/testing ), Stripe processes directly the payment error on their form : image Sum Up being Sum Up, I want to be sure it's covered before setting it again.

No, this is just for stripe. This alert simply displays the error when the [user pays](https://github.com/be-BOP-io-SA/be-BOP/blob/0182cfb28cedb3240b6fa4c6fdb04b4a8c2e0dae/src/routes/(app)/order/%5Bid%5D/payment/%5BpaymentId%5D/pay/%2Bpage.svelte#L243), not when we claim the payment. In the case of SumUp, the documentation is too vague on whether the error they provide can be displayed to the user or not [ref](https://developer.sumup.com/online-payments/tools/card-widget#configurations). Maybe we can check that together?

In any case, my understanding is that we cannot detect all errors here: For instance, if the payment fails because the user failed to provide additional information (such as 3d secure), the payment intent will effectively be created, but the failure happens later. Again, the documentation is not clear, so the most conservative approach is implemented.

Tirodem:
Ok.
The initial issue having been created for Sum Up specific bad behaviors in production, I propose we put this MR in Draft status as long as Sum Up is not done (it's not a priority of these upcoming weeks).
We'll check the Sum Up documentation together.

roosemberth:
Sumsub is done with the more conservative approach in the screenshot of the MR description.

roosemberth:
This MR is draft until commit “Debug statements for Christophe” is removed from the source branch.

This should happen once Christophe reports back on the behaviour of sumsub in case of an error.

Tirodem:
As seen, the API return is successfull but the status is FAILED and no error is displayed.
We need a quick peer-testing session for that 👌

PR #2011, opened 2025-07-25
🦋 Easy to use installation script

roosemberth:
The current changes in this PR builds a standalone release artifact suitable for production deployments. This enables production deployments to switch versions without the need to build the project.

This PR is draft until the necessary changes to address #1428 are included.

Closes #1428.

PR #2023, opened 2025-08-17
Pre-build svelte application in docker to accelerate starting time

marsu-dev:
Use dynamic env server side

- replace all $env/static/private by $env/dynamic/private
- keep PUBLIC_VERSION static for svelte layout & pages (available at build time)
- remove ORIGIN on build
- rename setup-env.sh
- run node in the final container

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,7 @@
 node_modules
 *.md
 dist
-.env.local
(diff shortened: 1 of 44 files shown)


marsu-dev:
#2011 will superseed this PR
Using this from downstream waiting for next release.

Issue #2022, opened 2025-08-15
☢️ Delivery fees are not billed and considered as a deposit leftover


Tirodem:
<img width="914" height="419" alt="Image" src="https://github.com/user-attachments/assets/0f4d6014-2eb3-4f05-9343-a5a424dcf486" />

<img width="918" height="511" alt="Image" src="https://github.com/user-attachments/assets/fe1016a7-33fb-4b32-824a-9ae4b307a275" />

<img width="406" height="253" alt="Image" src="https://github.com/user-attachments/assets/940860d3-809e-4a4f-8d79-c01249535dbc" />

I think it's related to https://github.com/be-BOP-io-SA/be-BOP/pull/1985#issuecomment-3081326524

PR #2034, opened 2025-08-27
🐛 Fix delivery fees not billed along with the order

roosemberth:
Previously, the delivery fees would not be included in the amount to pay; thus, paying for the order would be considered a deposit and a second payment (this time for the delivery fees) would be required.

Fixes https://github.com/be-BOP-io-SA/be-BOP/issues/2022.

### DIFF
--- a/src/routes/(app)/checkout/+page.svelte
+++ b/src/routes/(app)/checkout/+page.svelte
@@ -29,8 +29,6 @@
 	let country = defaultShippingCountry;
 
 	let isFreeVat = false;
-	let onLocation = data.hasPosOptions && data.defaultOnLocation;
-	$: offerDeliveryFees = onLocation;
 	let addDiscount = false;
 	let offerOrder = false;
 	let discountAmount = 0;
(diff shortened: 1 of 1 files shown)


roosemberth:
@Tirodem Can you please also check the current behaviour when a customer places an order for the deposit of a product that has delivery fees?

Tirodem:
@roosemberth test on main :
Product with no deposit and delivery :
image
Product with deposit and delivery :
image
image

Tirodem:
With fix :
image
image
Delivery fees appears on cart but not on checkout anymore :x:
image
image
Still, their cost is in the total amount

Tirodem:
Issue still here.
You can check with https://dev-bootik2.pvh-labs.ch/product/physical-product , it'll trigger the issue on checkout

Tirodem:
Last test.
As employee I add a product to my cart (10CHF VAT included).
The delivery fees are offered by default.
Here's my checkout (10CHF, no delivery fees) :
image
image
Here's my order (18CHF, delivery fees are to be paid) :
image
If I uncheck the delivery fees for free default option and I check the option to make them free :
image
image
Here's my cart, and delivery fees are free (that is expected) :
image
My guess is we have to retake the previous behavior :
- if "Il s'agit d'une commande physique sur place entièrement payée" option is checked, "Enlever les frais de port pour cette commande" should be displayed, checked and pre-filled with default text
- if "Il s'agit d'une commande physique sur place entièrement payée" option is unchecked, "Enlever les frais de port pour cette commande" should be displayed and unchecked

Tirodem:
Good for me 👌

PR #2035, opened 2025-08-27
🐛 Fix incorrect S3 client configuration used in release artifacts

roosemberth:


src/lib/server/s3.ts
```diff
@@ -10,32 +9,29 @@ import {
...
-	building || !PUBLIC_S3_ENDPOINT_URL
-		? s3client
-		: new AWS.S3({
-				endpoint: PUBLIC_S3_ENDPOINT_URL,
-				region: S3_REGION,
-				credentials: { accessKeyId: S3_KEY_ID, secretAccessKey: S3_KEY_SECRET },
-				// Handle minio, eg http://minio:9000/bucket instead of http://bucket.minio:9000
-				// Should work with any S3-compatible server, so no harm in leaving it on
-				forcePathStyle: true
-		  });
```
coyotte508, line 37: For info, `publicS3Client` is used in the case of docker compose, where the browser accesses minio through a different URL than the bebop process (which uses internal `minio` host instead of `localhost`)

src/lib/server/s3.ts
```diff
@@ -1,4 +1,3 @@
-import { building } from '$app/environment';
```
coyotte508, line 1: Was used to avoid errors when `pnpm build` on the project. Not sure if still relevant

See https://svelte.dev/docs/kit/building-your-app

src/lib/server/s3.ts
roosemberth: We want to decouple build-time from run-time. Specifically, the environment variables will not be present during build time (e.g. `pnpm run build`), but only once the application is started.

(Similarly, a full checkout of the repository won't be available at runtime.)

roosemberth:
Since 8393f398628896c22b17667a26492ca3223b34a9, this is no-longer needed.

/close

Issue #1978, opened 2025-06-13
🦋 PoS Pool management


Tirodem:
(but can also be called "Tab" instead of "Pool")

# Context

Following talk with Amuse Bar :
- make the pool management system works (PoC)
- make the pool admin interface
- manage non-null pool with unpaid leftover
- when creating / opening a Pool, allow to associate a number of customers (optional)
- keep track of pool creation time / opening time, final payment time / pool closure, and product addition time to pool for reporting purposes

# What is a Pool ?

In a restaurant, when a waiter/waitress take orders, he/she goes to the table, take orders, forward drink requests to the bartender/bartendress and food requests to the kitchen each time customers order new things.

Same for isolated customers at the counter.

At the end, the waiter/waitress fill cash-in either the whole table bill, an equal split of the bill, inequal split of the bill (arbitrary or corresponding to each customer drinks & foods) and will give a receipt for the payment (and eventually for each customer detailled food & drink). 

Same when a customer leave before the rest of the table, while the other will continue to order.

Waiters/waitress, of course, should be able to make parallel management of different pools (I take orders for pool 1, then I take orders for pool 2 and 3, then I take again orders for pool 1, I close pool 2 after payment, I take order again for pool 3...).

Note that a pool can be :
- a table of many customers that'll pay at the end
- a unique identified customer that'll pay at the end
- a random unidentified customer at the counter that'll pay at each order
- other possible usercases

# What is a Pool /tab ? (technical)

While it's tempting to consider that a Pool = a Cart, and the waiter/waitress switch between carts, we have to consider the whole picture and future related features, including :
- https://github.com/be-BOP-io-SA/be-BOP/issues/1968
- https://github.com/be-BOP-io-SA/be-BOP/issues/2039
- https://github.com/be-BOP-io-SA/be-BOP/issues/2038

Currently, on be-BOP, on both PoS and website :
- you add a 1st product
- a cart is created
- once you select a payment method on checkout and valid :
  - order is created from cart
    - invoice is created as pending
    - proforma bill is created
  - cart is deleted
- once a payment is done, receipt is created

Here, for bars and restaurants, here's what will happen :
- a waiter/waitress will have to manage various tables and independants customers, so they should be able to switch from one to another then come back to the first
- from orders of a table, waiter/waitress should make quick extract of newly ordered products to give to bartender/bartendress and cooks
- orders from a table should be paid either :
  - all at once in the end with a unique ticket
[...]

Issue #1968, opened 2025-06-04
🦋 PoS interface : allow split payment by clicking on ordered products


Tirodem:
Restaurant use-case
The idea is to pick products from a global order from a table to create split payments / subcarts.
Exemple :
- I have a ticket with 12 fries, 6 beers, 8 burgers and 3 t-bones
- I choose split payment
- I click on burger 3 times, t-bone 1 time, fries 3 times and beers 4 times
- This will create a payment for 3 burgers, 1 t-bone, 3 fries and 4  beers
- On the order, this will be left : 9 fries (12-3), 2 beers (6-4), 5 burgers (8-3), 2 t-bones (3-1)
This is a must-have for restaurants at a certain size.

Also, each payment should generate a receipt / ticket, for pro accounting purposes (like pro lunch).
This should include items from subcarts only, not the others, so we cannot use the exact same mechanic that we use on the payment with deposit.

In that case, we should considerate a pool not as a cart, but as a pool of products that can generate 1..n carts with its content, so 1..n orders from a pool.

With this new system, when clicking here on "Payer" CTA :
<img width="616" height="259" alt="Image" src="https://github.com/user-attachments/assets/c7552272-cae2-4395-ac7c-2ac0223657ab" />

Go to /pos/touch/split (or /pos/touch/split/{pool_id}) :
<img width="2480" height="3508" alt="Image" src="https://github.com/user-attachments/assets/23d931c3-f696-44f9-9632-c16d9dcddd30" />

- "**Encaisser (tout)**" CTA leads directly to checkout / payment where the whole pool is a cart.
- "**Séparer (parts)**" (to be designed yet) will leads directly to checkout where the whole pool is a cart but with multiple payments option (free amount with default being divided by a certain number in equal parts)
- "**Séparer (détail)**" whill lead to this new interface allowing to split a pool between 2..n carts by selecting products.

Interface consist of the pool on the left, a new cart on the right, and options on each product line :
- ➕ will move 1 product from pool's product line to cart
- ⏩ will move all products from pool's product line to cart

"**Encaisser**" CTA must be disabled as long as the cart is empty.

<img width="2480" height="3508" alt="Image" src="https://github.com/user-attachments/assets/7cc43dff-e23e-49bd-be94-5d5855152285" />

When a product line is created on cart :
- ➖ will move 1 product from cart's product line to pool
- ⏪ will move all products from cart's product line to pool

[...]

PR #2053, opened 2025-09-16
Initial implementation of order tabs

roosemberth:
An order tab holds products and quantities, that will be added into a cart and an order at a later time.

At this time, tabs can only be accessed by users with the POS role. A tab is independent of a user. A tab contains lines, each line contains a product and a quantity.

A collection of lines may be used to create a cart, in which case, the lines learn the cart they're associated to. When the cart is checked-out, the lines also learn the order they're associated to.

The current implementation handles gracefully when a cart previously associated to a tab is removed; likewise when an order associated to a tab is cancelled or expired.

Because tabs are accessible by any POS user, a cart may be associated to a tab, but then abandoned by the user (e.g. employee clocks out); in which case, the associated lines cannot be checked-out. This remains to be specified.

Addresses https://github.com/be-BOP-io-SA/be-BOP/issues/1978.

### DIFF
--- a/src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.server.ts
+++ b/src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.server.ts
@@ -88,10 +88,20 @@ async function getHydratedOrderTab(locale: Locale, tabSlug: string) {
 export const load = async ({ locals, depends, params }) => {
 	const tabSlug = params.orderTabSlug;
 	depends(UrlDependency.orderTab(tabSlug));
-	if (await orderTabNotEmptyAndFullyPaid({ slug: tabSlug })) {
+
+	const [shouldConclude, initialOrderTab] = await Promise.all([
+		orderTabNotEmptyAndFullyPaid({ slug: tabSlug }),
+		getHydratedOrderTab(locals.language, tabSlug)
+	]);
+
+	let orderTab;
+	if (shouldConclude) {
 		await concludeOrderTab({ slug: tabSlug });
+		orderTab = await getHydratedOrderTab(locals.language, tabSlug);
+	} else {
+		orderTab = initialOrderTab;
 	}
-	const orderTab = await getHydratedOrderTab(locals.language, tabSlug);
+
 	return {
 		orderTab: pojo(orderTab),
 		posTabGroups: runtimeConfig.posTabGroups,
@@ -120,8 +130,6 @@ export const actions = {
 		const { note, quantity, tabSlug, tabItemId } = parseUpdateOrderTabItemReq(
 			await request.formData()
 		);
-		const orderTab = await getHydratedOrderTab(locals.language, tabSlug);
-
 		if (!ObjectId.isValid(tabItemId)) {
 			throw error(400, 'The specified tab item is invalid');
 		}
@@ -129,7 +137,7 @@ export const actions = {
 		let res;
 		if (quantity === 0) {
 			res = await collections.orderTabs.updateOne(
-				{ _id: orderTab._id },
+				{ slug: tabSlug },
 				{
 					$pull: {
 						items: { _id: new ObjectId(tabItemId) }
@@ -138,7 +146,7 @@ export const actions = {
 			);
 		} else {
 			res = await collections.orderTabs.updateOne(
-				{ _id: orderTab._id },
+				{ slug: tabSlug },
 				{
 					$set: {
 						'items.$[elem].internalNote': {
--- a/src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
+++ b/src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
@@ -72,7 +72,7 @@
 			if (quantityInput && quantityInput instanceof HTMLInputElement) {
 				quantityInput.value = quantity.toString();
 			}
-			form.submit();
+			form.requestSubmit();
 		}
 		itemToEditIndex = undefined;
 	}
@@ -179,6 +179,14 @@
 								method="post"
 								bind:this={formNotes[i]}
 								action="?/updateOrderTabItem"
+								use:enhance={() => {
+									return async ({ result }) => {
+										if (result.type === 'error') {
+											return await applyAction(result);
+										}
+										await invalidate(UrlDependency.orderTab(tabSlug));
+									};
+								}}
 							>
 								<input type="hidden" name="note" />
 								<input type="hidden" name="tabItemId" value={item.tabItemId} />


roosemberth:
@Tirodem

- What should happen if one employee creates a cart and abandons the site?
- What should happen after we've paid for the entire tab? Should the tab be *deleted*, *archived* or should we replace the “PAYER” button for a different CTA?
- How does an employee remove a product added by mistake?

Tirodem:
> * What should happen if one employee creates a cart and abandons the site?

Through classic interface or through PoS touchscreen (or both) ?
If employee manages a cart on classical interface, it's cart only, not a pool, it'll be saved / associated with session, and voilà.
If employee manage a pool through PoS touchscreen and abandons the site, another employee will be able to select this pool and use it.
We need to manage a cart-only behavior outside the PoS touchscreen.

> * What should happen after we've paid for the entire tab? Should the tab be _deleted_, _archived_ or should we replace the “PAYER” button for a different CTA?

cf #1968 :
image
If the tab was paid through through the "Encaisser (tout)" option, the whole tab becomes a cart, the cart is paid, then the tab is closed and can be reuse for adding products.

If the tab was paid through the "Séparer (parts ou détail)" options, you land on this :
image
When pool left to pay amount is 0, you have the "Close pool" option.

> * How does an employee remove a product added by mistake?

Either by changing quantity ( #1981 ) or by clicking ❎ CTA to delete last entry.

roosemberth:
>         What should happen if one employee creates a cart and abandons the site?
> 
> Through classic interface or through PoS touchscreen (or both) ?

Both. Today, an employee can only create a cart from the tab through the POS interface, but it's much simpler if the behaviour is consistent regardless of the mechanism.

>  If employee manage a pool through PoS touchscreen and abandons the site, another employee will be able to select this pool and use it.

How do we detect when an employee abandons the site?

roosemberth:
@Tirodem I think we're only missing translations and the splitting feature. Should I go ahead and add these too?

Tirodem:
Yes, you can, I'll make a whole test batch end of afternoon.

Tirodem:
This orderTab went 500 after a product addition, I save the datas for diagnosis (then purge datas and continue my test) :
```
{
  "_id": {
    "$oid": "68d95489475d1ff45767591c"
  },
  "createdAt": {
    "$date": "2025-09-28T15:30:17.553Z"
  },
  "items": [
    {
      "_id": {
        "$oid": "68d954a8475d1ff45767591d"
      },
      "productId": "la-cuisine-des-titans",
      "quantity": 1
    },
    {
      "_id": {
        "$oid": "68d954a9475d1ff45767591e"
      },
      "productId": "hypno-toddler",
      "quantity": 2
    },
    {
      "_id": {
        "$oid": "68d954aa475d1ff45767591f"
      },
      "productId": "rostis-chaux-de-fonniers",
      "quantity": 1
    },
    {
      "_id": {
        "$oid": "68d954ab475d1ff457675920"
      },
      "productId": "pourboire",
      "quantity": 1
    }
  ],
  "slug": "tables-0",
  "updatedAt": {
    "$date": "2025-09-28T15:30:17.553Z"
  }
}
```

Tirodem:
Ok, I feared it was about PWYW product but two others are added with minimal amount.
The product provoking 500 error is https://dev-bootik2.pvh-labs.ch/admin-tirodem/product/pourboire
I remove it from my tests but we'll check that later in the week.

Console error when adding this little rascal :
```
start.93a06ab9.js:1  GET https://dev-bootik2.pvh-labs.ch/pos/touch/__data.json?tab=tables-0&filter=pos-favorite&skip=0&x-sveltekit-invalidated=0001 500 (Internal Server Error)
Fe @ start.93a06ab9.js:1
pe @ start.93a06ab9.js:1
xe @ start.93a06ab9.js:1
await in xe
invalidate @ start.93a06ab9.js:1
(anonyme) @ singletons.d9267ed7.js:1
(anonyme) @ 168.2b88ba92.js:1
l @ forms.5144a2c1.js:1
```

Tirodem:
Pool mechanic as standalone seems functional (and performances are acceptables now).

For the 500 on product addition, a new product met the bug :
- we had https://dev-bootik2.pvh-labs.ch/admin-tirodem/product/pourboire
- we also have

Pool switch UI might be a bit tiny on my usual viewport, we should keep an eye on it on real test devices.
image

Pool switching is OK. Maybe a small loading performance issue, but it's acceptable so far. Adding a spinner and displaying pool name more explicitely in the future would be a great addition.
However, while testing, I had a full wipe of my every pools ☢️ to diagnosis

After paying my whole pool, I tried to came back to /pos/touch and meet a 500 error, while I was redirecteed (still with error) on tab=tables-0
Turns out it was an existing pool with error 500 due to product, but it makes me think : may be  when you fully pay Pool-5 (for example) it's better to go back to Pool-5 when you finish a payment 💡 

Unless I missed something, we need some data tracking to connect an order to its initial pool 💡 

And, finally, we'll need to wipe the Pool when it's paid : currently, I fully paid mine and every product is still here. But I imagine it'll be managed on Split Payment issue.

Nice progress.

Tirodem:
Also note that we need to adapt https://github.com/be-BOP-io-SA/be-BOP/pull/2068 for Pool behavior.

roosemberth:
Pool switching performance has been improved. A button to return to the POS screen was added to the order.

The current behaviour is to remove the order tab when not-empty and all items have been paid for. (This behaviour was already implemented, but did not trigger previously in the sandbox due to a bug.)

I've fixed the issue with the `pourboir` product (affecting products with a specific tax profile).

> Unless I missed something, we need some data tracking to connect an order to its initial pool 💡

Both the cart and order now store the related order tab.

I've added an alert for the not-yet-implemented split features. I'll continue with the translations and then rebase while you test the latest changeset.

Tirodem:
- no more 500 on custom VAT product ✅ 
- whole experience seems more stable and fluid ✅ 
- pool is cancelled after transformed into an order (no split payment) ✅ 
- orderTabSlug is added to orders object and cart ✅

Tirodem:
Last expectations :
- translation keys
- switch selector improvement (too small on real device)
- bringing #1981 on Pool behavior

Those will be done in the future :
- #2094
- #2095
- #2096
- #2097

roosemberth (commented):
I've read through my commits. Unfortunately, GitHub won't allow me to approve my own PRs. @Tirodem merge when convenient.

Tirodem:
Oh, funny thing, I don't know if it's related to the merge with main (with advanced HTML edition) but now /pos doesn't have footer / header anymore 😱 
image
Not dramatic but unusual

Tirodem:
Everything is OK except the quantity selector, it works but I have the impression it's a bit slow (at least slower than adding products manually).
WDYT ?

roosemberth:
@Tirodem I've applied some optimizations:

Previously, upon submission of the form, the following tasks were performed.
- Submit the update (122.04ms - 240.659ms)
- Load the data for all the products displayed (412.528ms - 767.63ms)
- Load the data for the products in the tab (159.988ms - 490.578ms)

With the latest patch, the following tasks are performed:
- Submit the update (unchanged)
- (The data for all the products displayed is reused - 0ms)
- Load the data for the products in the tab (83.501ms - 129.716ms)

The only negative impact is after a tab is fully paid: This code-path will take slightly longer (few 100ms), since now optimistically fetch the tab information before knowing whether we need to delete the tab.

(The measurements were taken on my laptop using the dev-bootik2 database over 5G.)

roosemberth:
@Tirodem If necessary, I can optimize this further (for both the quantity selector dialog and adding products manually).

Tirodem (approved):

Issue #2050, opened 2025-09-10
🐛 PoS order with 100% = error ; Free order with 100% subscription discount = error. (both 400)


Tirodem:
1st case :
- log (employee) as PoS
- add paying product
- apply 100% discount on checkout
Result :
<img width="1288" height="134" alt="Image" src="https://github.com/user-attachments/assets/b44be840-250d-4f2f-aa38-6737d8472517" />

2nd case :
- enable a discount of 100% on products associated to a subscription
- log (user) with an active subscription
- add a fully discounted product
- process checkout with cart = 0€
Result :
<img width="1268" height="157" alt="Image" src="https://github.com/user-attachments/assets/82bff521-0299-467c-968a-de41a2c2ee18" />

Note that this exact message rings a bell : at a time it was possible to use "!checkout free" on nostr-bot even on non-free products.
Here's the commit done to fix that :
https://github.com/be-BOP-io-SA/be-BOP/commit/8f201789a431cb36db70f2a6e6aad93369defe5a

PR #2057, opened 2025-09-17
🐛 Checkout: Use discount on frontend price computation

roosemberth:
Incidentally, this fixes the "free" payment method not being available when the discount is for the whole order.

/closes https://github.com/be-BOP-io-SA/be-BOP/issues/2050

### DIFF
--- a/src/lib/components/OrderSummary.svelte
+++ b/src/lib/components/OrderSummary.svelte
@@ -123,26 +123,26 @@
 			</div>
 
 			<div class="flex flex-col ml-auto items-end justify-center">
-				<PriceTag
-					class="text-2xl truncate {item.discountPercentage ? 'line-through' : ''}"
-					amount={orderPriceInfo.perItem[i].unitsToBill *
-						(item.currencySnapshot.main.customPrice?.amount ??
-							item.currencySnapshot.main.price.amount)}
-					currency={item.currencySnapshot.main.customPrice?.currency ??
-						item.currencySnapshot.main.price.currency}
-					short={!!item.discountPercentage}
-				/>
 				{#if item.discountPercentage}
 					<PriceTag
-						class="text-2xl truncate"
+						class="text-2xl truncate line-through"
 						amount={orderPriceInfo.perItem[i].unitsToBill *
 							(item.currencySnapshot.main.customPrice?.amount ??
-								item.currencySnapshot.main.price.amount) *
-							(item.discountPercentage ? (100 - item.discountPercentage) / 100 : 1)}
+								item.currencySnapshot.main.price.amount)}
 						currency={item.currencySnapshot.main.customPrice?.currency ??
 							item.currencySnapshot.main.price.currency}
+						short={!!item.discountPercentage}
 					/>
 				{/if}
+				<PriceTag
+					class="text-2xl truncate"
+					amount={orderPriceInfo.perItem[i].unitsToBill *
+						(item.currencySnapshot.main.customPrice?.amount ??
+							item.currencySnapshot.main.price.amount) *
+						(item.discountPercentage ? (100 - item.discountPercentage) / 100 : 1)}
+					currency={item.currencySnapshot.main.customPrice?.currency ??
+						item.currencySnapshot.main.price.currency}
+				/>
 				{#if item.currencySnapshot.secondary}
 					<PriceTag
 						class="text-2xl truncate"


roosemberth:
@Tirodem can you please double-check the second scenario? After fixing the first one I was unable to reproduce the second one.

Tirodem:
1st is OK, i'm checking the second

Tirodem:
Issue's still here :
image
For memories :
- you have an active subscription
- the subscription is related to an active discount
- the discount gives 100% discount on Product A (Product A should not be a product with physical component / delivery fees)
- you log with either your npub / email via /login
- you add Product A (free in cart)
- you go through checkout

Then you'll have the error, supposedly.

src/lib/server/orders.ts
```diff
@@ -804,6 +810,9 @@ export async function createOrder(
 		.filter((sub) => discountSubIds.has(sub.productId))
 		.map((sub) => sub.productId);
 
+	for (const item of items) {
+		item.discountPercentage ??= discountByProductId.get(item.product._id) ?? wholeDiscount;
```
coyotte508, line 814: Maybe there should be a `Math.max(...)` here to take the biggest discount between whole catalog discount percentage, product discount percentage (and existing percentage discount?)

Tirodem:
image
image

Total cart amount is right but the cart preview in /order is kinda funny.
As it was done on a previous branch I'll make the cart again :

image
image
image

Yup, we have a front issue with price displayed.
It works way better than earlier but we have to fix that.

BTW, the invoice :
image
is good

src/lib/server/orders.ts
roosemberth: No need: In line 802, the code already uses the max between `wholeDiscount` and `d.discountPercent` (with extra steps).

If we want to make an improvement; to make this easier to read, we could simply move both `discountByProductId` and `wholeDiscount` into a function `getDiscountForProduct(productId: string): number | undefined`.

src/lib/server/order.test.ts
```diff
@@ -217,4 +228,60 @@ describe('order', () => {
...
+			productId: TEST_SUBSCRIPTION_PRODUCT._id,
+			paidUntil: addDays(new Date(), 1),
+			createdAt: new Date(),
+			updatedAt: new Date(),
+			number: 1,
+			user: {
+				sessionId: 'test-session-id'
+			},
+			notifications: []
+		});
```
roosemberth: Can you move this effect (and the next one) to a fixture function: `paidSubscriptionExists(...)`. This way, any actually relevant arguments become obvious.

The same comment applies to the next statement inserting a discount.

src/lib/server/orders.ts
coyotte508: yes I realized afterwards, that the aggregation is also already sorted on discountPercent decreasing

coyotte508:
image

Tirodem:
image

🎉 

Behavior is ✅ , waiting for the PR review.

src/lib/components/OrderSummary.svelte
```diff
@@ -123,11 +123,23 @@
...
+						short={!!item.discountPercentage}
+					/>
+				{/if}
 				<PriceTag
 					class="text-2xl truncate"
 					amount={orderPriceInfo.perItem[i].unitsToBill *
 						(item.currencySnapshot.main.customPrice?.amount ??
-							item.currencySnapshot.main.price.amount)}
+							item.currencySnapshot.main.price.amount) *
+						(item.discountPercentage ? (100 - item.discountPercentage) / 100 : 1)}
```
roosemberth, line 142: Why are we recomputing price elements in the frontend? The price computation is the responsibility of `computePriceInfo`. The information should come from `CartPriceInfo` (or added if not present).

coyotte508: here it's after order has been created - price is fixed in stone and has already been through `computePriceInfo` 

it's based on this data:

https://github.com/be-BOP-io-SA/be-BOP/blob/74b080cd8452843b35874f167bb112d6a46ad1f0/src/lib/types/Order.ts#L165-L187

(I guess we could store `amount` and `amountWithoutDiscount` in order data, but would need migration)

roosemberth: > (I guess we could store amount and amountWithoutDiscount in order data, but would need migration)

In principle, we didn't lose any information; it would be more convenient if `computePriceInfo` could simply evaluate the items in an order...

Either way, that's outside the scope of this PR. Let's keep the current code :+1:

Tirodem:
Weirdissimo : https://github.com/be-BOP-io-SA/be-BOP/issues/2076

PR #2060, opened 2025-09-22
Add devcontainer config

coyotte508:
Add devcontainer config, for easy development. Just open the project in VS code / cursor and follow the instructions.

This allows to run the project inside a docker container to not contaminate your main env, but also have hot reload, ... without waiting for docker build times.

Notes:

- added `.nvmrc` with 22.19.0. This is the current LTS of node. We should probably edit the `Dockerfile` to use nvm + nvmrc, or at least switch the base image to node-slim:20
- didn't edit the main docker-compose file. Ideally it would be reused/extended, but didn't want to cause any problem with existing deployments


coyotte508:
closing for now, until it's used by at least one member of the permanent dev team :)

Issue #2014, opened 2025-07-28
🦋 Allow for be-BOP on-chain nodeless to connect to NBExplorer


Tirodem:
TBD
To help with be-BOP x BTC Pay Server integration

PR #2063, opened 2025-09-22
Fix/bitcoin nodeless management 1951

### LINKED ISSUE #1951: 🦋 Improve bitcoin nodeless management

Image

TBC

gntw:
Fix: Improve bitcoin nodeless management (#1951)
   
   Implements three solutions for address reuse prevention:
   - Temporal filtering for transactions
   - Admin warning for used addresses  
   - Skip used addresses during generation
   
   Fixes #1951


Tirodem:
First point seems OK, I'm QA-digging before going to 2nd point.
image

Tirodem:
For the second case :
- I've made an order on main with an existing address with funds : the order was marked as paid withou any payment (as expected) (in fact it was worse than expected, as the order is paid and the payment expires later, with the notification being sent to the customer).
- I did the same on this branch : the payment doesn't go directly as "paid" ✅
  - edit : my bad, in fact, the adress was bypassed (3rd case)
- I still need to make a real payment to check if the "pending" to "paid" on an already used address works.
  - edit : no need as the used address is bypassed. I still need to check on a wallet with a lot of used address to check the performances.
- Then I'll do an order on a full, virgin address, for non-regression purpose.

That might take a bit of time as I don't have testnet wallet anymore 🕐 Stay tuned

Tirodem:
From what I see while testing :
- (1st case) in /admin/bitcoin-nodeless, we display `!already used` if the address carries past transactions, even if the current amount on this address is 0
- (3rd case) if my next address is already used and if its current amount is > order amount (or at least not null), the address will be skipped, and a new one will be used instead for the order

But if my address was already used and doesn't carry any amount anymore :
- it will be marked as `!already used` on /admin/bitcoin-nodeless
- it will be used on the next order

(it acted like that in my last test order)

@gntw do you confirm this hypothesis ?

It not necessarily a very bad thing but, if we keep the 3rd behavior, we might harmonize things a little.

BTW, considering the 3rd behavior is already implemented, it's hard to test the 2nd behavior.

gntw:
@Tirodem sorry, but i cannot confirm the hypothesis you described. On my local setup using Signet, the behavior is different:

My test scenario: 
- Used address `tb1qf8vsq9thsxnprgyrux9wsf6c4edthshtv6dp6s` with past transactions (10k sats)
- Sent all sats back to faucet to make balance = 0
- Waited for 1+ confirmations
- Created a new order

Results:
- Address was showed `! already used` in `/admin/bitcoin-nodeless`
- But when creating an order, this address was skipped 
- Next clean address was presented at Order instead

So in my case, the 3rd behavior (skip used addresses) works,  any address with transaction history will be skipped regardless of current balance

Could you share which network you're testing on and the specific address that wasn't skipped?

Tirodem:
As seen in DM, I have to find back my particular zpub / bc1 to confirm what I saw earlier.

In a more general way, as :
- point 1 is working
- point 2 is not testable because of point 3
- point 3 is working but might have performance issues on heavily used existing wallets (and I don't have test datas for this right now)

So, we have 2 possibilities :
- we remove point 3 but it sad to potentially remove good code
- we keep point 3's code but make it optional through an option admin

As product I have a huge preference for this last option, so, let's add this in admin/bitcoin-nodeless :
image

If unchecked :
- do not apply code from point 3 (that way, point 2 will apply - and be testable more easily)

If checked :
- apply code from point 3
- warn user about potential performance issues (fix my typo with "experimenal", miss a t 😅)
image

src/lib/server/bitcoin-nodeless.ts
```diff
@@ -32,27 +32,30 @@ export function isZPubValid(zpub: string): boolean {
...
-		{ returnDocument: 'before' }
-	);
-
-	if (!res.value) {
-		throw new Error('Failed to increment derivation index');
+	let index = runtimeConfig.bitcoinNodeless.derivationIndex;
+	
+	for (let attempts = 0; attempts < 10; attempts++) {
+		const address = bip84Address(runtimeConfig.bitcoinNodeless.publicKey, index);
+		const isUsed = await isAddressUsed(address).catch(() => false);
```
roosemberth: This catch silences a potentially unrelated issue (e.g. API rate-limit); we should at least log the failure.

src/lib/server/bitcoin-nodeless.ts
```diff
@@ -32,27 +32,30 @@ export function isZPubValid(zpub: string): boolean {
...
+		
+		if (!isUsed) {
+			await collections.runtimeConfig.updateOne(
+				{ _id: 'bitcoinNodeless' },
+				{ $set: { 'data.derivationIndex': index + 1, updatedAt: new Date() } }
+			);
```
roosemberth: The pattern of updating both the database and inmemory object of `runtimeConfig` is pretty common.

We should at least use `persistConfigElement`.

In a separate commit we could improve `persistConfigElement` to return the persisted value **and** update `runtimeConfig`.

Tirodem:
As need in DM, this feature has a flaw (that might explain the weird behavior from https://github.com/be-BOP-io-SA/be-BOP/pull/2063#issuecomment-3318287732 ) : while heavily testing, point 1 (aka checking every address on /admin/bitcoin-nodeless) does trigger mempool API rate limit.

So, if we encouter this situation :
- point 2 is broken
- point 3 is broken
- pending payment check on address is broken

Which is logical : it doesn't have to be fixed on this code, the feature needs some redesign, and be-BOP needs some additional checks.

What we can do :
- [ ] On /admin/bitcoin-nodeless , check if addresses are already used only when updating bitcoin nodeless settings
  - While a standard use might not provoke any issue, just displaying the page without any change should not be API call consuming.
- [ ] Add an option to refresh checks manually
  - image
- #2066 (for later)
- #2067 (for later)

Great comment btw :

> The update button will keep us from getting extra requests, BUT before that, we were hoping that, for example, if the key was stolen, the user would notice the problem because they would see orange markers in the admin panel. like "what?"

It's a good point, but mempool doesn't provide premium plan, so the only alternative here might be https://github.com/be-BOP-io-SA/be-BOP/issues/2014 and possibly https://github.com/be-BOP-io-SA/be-BOP/issues/2019

src/lib/server/bitcoin-nodeless.ts
```diff
@@ -32,27 +32,30 @@ export function isZPubValid(zpub: string): boolean {
...
+			runtimeConfig.bitcoinNodeless.derivationIndex = index + 1;
+			return index;
+		}
+		index++;
 	}
-
-	const val = res.value.data as typeof defaultConfig.bitcoinNodeless;
-
-	return val.derivationIndex;
+	throw new Error('Too many used 	addresses in sequence');
```
roosemberth: If the calling context only has the function name (ie. `generateDerivationIndex`), it's hard to understand what this error is. From the point of view of the user of this function, they don't know what policy is used to generate an address.

How about: _Unable provide the next derivation index because too many intermediate derivations are already used_?

Since we're already computing the address here, why not simply change the semantic of this function to return the next unused address? In this case, the error becomes much simpler: _Unable to provide the next unused address due to many intermediate addresses already used_.

src/lib/server/bitcoin-nodeless.ts
```diff
@@ -32,27 +32,30 @@ export function isZPubValid(zpub: string): boolean {
...
 }
 
 export async function getSatoshiReceivedNodeless(
 	address: string,
-	confirmations: number
+	confirmations: number,
+	orderCreatedAt?: Date,
+	orderExpiresAt?: Date
```
roosemberth: Is there really a case where we don't know the `orderCreatedAt` or `orderExpiresAt`?

It seems to me that this information is always known. If so, let's make these arguments mandatory, to simplify the possible combinations of this code.

src/lib/server/bitcoin-nodeless.ts
```diff
@@ -113,3 +133,15 @@ export async function getSatoshiReceivedNodeless(
 		transactions
 	};
 }
+
+export async function isAddressUsed(address: string): Promise<boolean> {
+	try {
+		const mempoolUrl = trimSuffix(runtimeConfig.bitcoinNodeless.mempoolUrl, '/') + `/api/address/${address}/txs`;
```
roosemberth: Why not simply use https://mempool.space/docs/api/rest#get-address?

We only need `tx_count`.

src/lib/server/locks/order-lock.ts
```diff
@@ -101,38 +101,40 @@ async function maintainOrders() {
...
-										throw new Error('Failed to fetch block height');
-									}
-
-									const blockHeight = z.number().parse(await resp.json());
-
-									if (runtimeConfig.bitcoinBlockHeight !== blockHeight) {
-										console.log('Updating bitcoin block height to', blockHeight);
-										runtimeConfig.bitcoinBlockHeight = blockHeight;
-										runtimeConfigUpdatedAt['bitcoinBlockHeight'] = new Date();
+									try {
```
roosemberth: The contents of this try block could easily be moved into a function; this would make what we're `try`'ing simpler to understand.

src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.server.ts
```diff
@@ -1,25 +1,43 @@
...
+			
+			return { address, isUsed, index };
+		})
+	);
+
 	return {
 		bitcoinNodeless: runtimeConfig.bitcoinNodeless,
-		nextAddresses
+		nextAddresses,
+		hasAlreadyUsedNextAddresses: nextAddresses.some(addr => addr.isUsed)
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.server.ts:43]: Suggestion: `hasAlreadyUsedNextAddresses` could easily be computed in the client.

If you prefer to keep it here, you can resolve this comment with no changes.

src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.svelte
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.svelte:1]: All of these changes pertain changes in you previous commit. It's preferable to squash these changes into the previous commit.

src/lib/server/bitcoin-nodeless.ts
roosemberth, line 1: All of these changes pertain your first commit. It's preferable to squash these changes onto that commit.

src/lib/server/bitcoin-nodeless.ts
```diff
@@ -39,10 +39,15 @@ export async function generateDerivationIndex(): Promise<number> {
...
-			await collections.runtimeConfig.updateOne(
+			const result = await collections.runtimeConfig.updateOne(
 				{ _id: 'bitcoinNodeless' },
 				{ $set: { 'data.derivationIndex': index + 1, updatedAt: new Date() } }
 			);
+			
+			if (result.matchedCount === 0) {
+				throw new Error('Bitcoin nodeless config not found');
+			}
+
```
roosemberth: There are multiple legitimate reasons why this setting may not found.

If we're concerned by this setting not existing, we should upsert it. Thus, this should be done by a specialized function that every code interested in this setting uses to obtain its value.

For example, see [here](https://github.com/be-BOP-io-SA/be-BOP/blob/538f8bcdb60c2f5b2855ec726f4780837e211d1c/src/lib/server/cart.ts#L22).

src/lib/server/bitcoin-nodeless.ts
roosemberth: This comment has not been addressed. The code still uses `/api/address/${address}/txs`.

Tirodem:
Mempool.space seems down this afternoon 😢

Tirodem:
I'm going with mempool.holdings (https://x.com/mempool/status/1970839824132382935)

Tirodem:
Point 3 is functional ✅ 
I'm waiting a testnet3 faucet to check if confirmation check on on-chain transaction is still OK.
I'll also play with an existing order to replace the orders.payment.address with an already used one, in order to test point 2.
@gntw can you confirm to me that the selector to enable point 3 or not is not developped yet ?

Tirodem:
Point 2 works ✅ 
Initially it was a virgin address (so pending), then I replace with an already filled address but no recent transaction (so pending) then I replace with an already filled address with recent transaction (so pending > confirmed).

roosemberth:
@gntw Can you please address the linting issues? Also, please rebase your branch :)

roosemberth:
Superseeded by https://github.com/be-BOP-io-SA/be-BOP/pull/2099.

/close

PR #2068, opened 2025-09-22
Quantity selection on POS screen after clicking ticket item

coyotte508:
Fix https://github.com/be-BOP-io-SA/be-BOP/issues/1932

https://github.com/user-attachments/assets/8a3cab06-c1f5-4704-9dd5-8df5cb6f0517

### DIFF
--- a/src/routes/(app)/cart/[id]/+page.server.ts
+++ b/src/routes/(app)/cart/[id]/+page.server.ts
@@ -126,7 +126,7 @@ export const actions = {
 
 		const formData = await request.formData();
 
-		const max = product.maxQuantityPerOrder || DEFAULT_MAX_QUANTITY_PER_ORDER;
+		const max = product.maxQuantityPerOrder ?? DEFAULT_MAX_QUANTITY_PER_ORDER;
 
 		const { quantity, mode, lineId } = z
 			.object({


Tirodem:
It was globally fine until this :
image
https://dev-bootik2.pvh-labs.ch/product/cassoulet
I have to check why this one and not the others.
Product doesn't have stock, is not standalone, and the error trigger when I go from 1 to 10, 1 to 9...
If I add one then add another and another and another I can go up to 10 (by clicking on product).
My cart at this time :
```
{
  "_id": {
    "$oid": "68d1aed7940bfc546f0644d9"
  },
  "createdAt": {
    "$date": "2025-09-22T20:17:27.914Z"
  },
  "items": [
    {
      "_id": "54dabcfe-1069-4432-86d7-e6ba4e17117d",
      "productId": "hypno-toddler",
      "quantity": 6,
      "reservedUntil": {
        "$date": "2025-09-22T20:21:15.893Z"
      }
    },
    {
      "_id": "287e92b3-c044-4f39-94e8-b8bd0013e403",
      "productId": "blonde-on-tap-5d-l",
      "quantity": 9,
      "reservedUntil": {
        "$date": "2025-09-22T20:20:00.856Z"
      }
    },
    {
      "_id": "550272a8-fc14-480a-b32f-d582a7cf1041",
      "productId": "plat-du-jour",
      "quantity": 8,
      "reservedUntil": {
        "$date": "2025-09-22T20:20:28.906Z"
      }
    },
    {
      "_id": "8da77d1a-0054-45b3-8245-e6a7eb9e1c0c",
      "productId": "colonel",
      "quantity": 3,
      "reservedUntil": {
        "$date": "2025-09-22T20:19:42.445Z"
      },
      "internalNote": {
        "value": "saignan",
        "updatedAt": {
          "$date": "2025-09-22T20:18:33.662Z"
        },
        "updatedById": {
          "$oid": "6526e1ba4232f5ddefcc98c0"
        }
      }
    },
    {
      "_id": "5c8dfe83-9dcf-48a7-a00e-295f003e29d8",
      "productId": "rostis-chaux-de-fonniers",
      "quantity": 6,
      "reservedUntil": {
        "$date": "2025-09-22T20:19:51.099Z"
      },
      "internalNote": {
        "value": "2 bleu\r\n1 sans salade",
        "updatedAt": {
          "$date": "2025-09-22T20:18:45.738Z"
        },
        "updatedById": {
          "$oid": "6526e1ba4232f5ddefcc98c0"
        }
      }
    },
    {
      "_id": "9e48cf50-e7c1-4bde-9618-449f0dc400f3",
      "productId": "steak-frites",
      "quantity": 1,
      "reservedUntil": {
        "$date": "2025-09-22T20:19:32.452Z"
      }
    },
    {
      "_id": "2114ad69-c88b-4487-bac3-8ae11b91721d",
      "productId": "best-ball-duplicate-2",
      "quantity": 1,
      "reservedUntil": {
        "$date": "2025-09-22T20:19:33.177Z"
      }
    },
    {
      "_id": "932a741a-655b-4b6f-bf83-45d0df5829b1",
      "productId": "square-pic",
      "quantity": 10,
      "reservedUntil": {
        "$date": "2025-09-22T20:21:00.883Z"
      }
    },
    {
      "_id": "a6ac5d8a-6e9b-45c9-b1a3-eb61501e8e80",
      "productId": "cassoulet",
      "quantity": 1,
      "reservedUntil": {
        "$date": "2025-09-22T20:23:09.407Z"
      }
    }
  ],
  "updatedAt": {
    "$date": "2025-09-22T20:22:06.479Z"
  },
  "user": {
    "ssoIds": null,
    "userId": {
      "$oid": "6526e1ba4232f5ddefcc98c0"
    },
    "email": null,
    "secondaryEmails": null,
    "npub": null,
    "sessionId": "182a900e319f7c8b44ab7ca916e9201bb6260745080e17a04a3f4d7ff234d8ff",
    "userLogin": "tirodem",
    "userRoleId": "super-admin",
    "userAlias": "CPO3",
    "userHasPosOptions": true
  }
}
```
If I delete one complete product line then I try to update this product quantity : same issue
Datas are from https://dev-bootik2.pvh-labs.ch/
Console error when it happens :
```
start.cbeb5d18.js:1  POST https://dev-bootik2.pvh-labs.ch/cart/best-ball-duplicate-2?/increase 400 (Bad Request)
window.fetch @ start.cbeb5d18.js:1
u @ 167.1a37f91f.js:5
I @ 167.1a37f91f.js:1
```
In Network :
```
window.fetch @ https://dev-bootik2.pvh-labs.ch/_app/immutable/entry/start.cbeb5d18.js:1
u @ https://dev-bootik2.pvh-labs.ch/_app/immutable/nodes/167.1a37f91f.js:5
I @ https://dev-bootik2.pvh-labs.ch/_app/immutable/nodes/167.1a37f91f.js:1
```
Note that we also have this error in classical cart with the same session when I try to increment this product quantity :
image

Tirodem:
I'll deploy `main` to check the behavior on the same datas just in case.

Tirodem:
Ahahah, funny thing with main 🤡 
With main :
- incrementing "Cassoulet" gives error on /cart
- adding a new "Cassoulet" by clicking on product on /pos/touch works !

🤷‍♂️

Maybe the code on this branch is cart-incrementation related, and as there's an error on cart incrementation...
I dunno.

This code might be good and the fix might be located on cart incrementation methods on legacy but we need to investigate this.

coyotte508:
well I can't test on dev-bootik-2: 

> Ce produit n'est pas disponible à la vente par cette plateforme, merci de nous contacter

coyotte508:
added a commit to log error in console (in POS interface), if you can reproduce

Tirodem:
on my way

Tirodem:
I've checked but there's not that much more in console.

In /pos/touch :
image

In /cart :
image

Did I miss something ?

Tirodem:
BTW I've added you as a point of sale user on dev-bootik2 just in case.
I just have to deploy another branch on it for 10-15mn before putting this one back.

src/routes/(app)/cart/[id]/+page.server.ts
```diff
@@ -57,23 +57,25 @@ export const actions = {
...
 					.int()
 					.min(1)
 					.max(max - 1),
 				deposit: z.boolean({ coerce: true }).optional(),
+				by: z
+					.number({ coerce: true })
+					.int()
+					.min(1)
+					.max(max - 1)
+					.default(1),
```
Comment (roosemberth) [inline src/routes/(app)/cart/[id]/+page.server.ts]: Is a default really needed here? Why can't all users of this action specify the argument?

src/routes/(app)/cart/[id]/+page.server.ts
```diff
@@ -57,23 +57,25 @@ export const actions = {
...
 			})
-			.parse({
-				quantity: formData.get('quantity'),
-				deposit: formData.get('deposit'),
-				lineId: formData.get('lineId')
-			});
+			.parse(Object.fromEntries(formData));
 
-		await addToCartInDb(product, quantity + 1, {
+		await addToCartInDb(product, quantity + by, {
```
Comment (roosemberth) [inline src/routes/(app)/cart/[id]/+page.server.ts]: From what I can tell, both quantity and by are only used in this expression.

Why can't we use a single argument instead? Is there a semantic reason why these two arguments should be part of the `increase` function?

src/routes/(app)/cart/[id]/+page.server.ts
```diff
@@ -71,13 +71,14 @@ export const actions = {
 					.min(1)
 					.max(max - 1)
 					.default(1),
+				mode: z.enum(['eshop', 'pos']).optional(),
```
Comment (coyotte508) [inline src/routes/(app)/cart/[id]/+page.server.ts]: no risk here, the `pos` is checked in `addToCartInDb` against the user

src/routes/(app)/pos/touch/+page.svelte
```diff
@@ -34,30 +35,107 @@
...
+		const item = items[index];
+		const currentQuantity = item.quantity;
+
+		if (quantity === currentQuantity) {
+			return;
+		}
+
+		// Update UI optimistically
+		items[index].quantity = quantity;
+		items = [...items];
```
roosemberth: Why is this necessary?

I see multiple instances of this statement.

src/routes/(app)/pos/touch/+page.svelte
```diff
@@ -34,30 +35,107 @@
...
+
+		const formData = new FormData(formNotes[index]);
+		formData.set('note', note);
+
+		try {
+			const response = await fetch(formNotes[index].action, {
+				method: 'POST',
+				body: formData
+			});
+			const result = await response.json();
```
roosemberth, line 61: Are the svelte action responses really JSON?

It's rather unintuitive given that the request is not JSON, but form-data. It seems the only thing we want here is to resolve the promise.

src/routes/(app)/pos/touch/+page.svelte
coyotte508: removed

src/routes/(app)/cart/[id]/+page.server.ts
Comment (coyotte508) [reply src/routes/(app)/cart/[id]/+page.server.ts]: I can change the code if you want to add `by`  everywhere, to me a default value makes sense as it's the default behavior

src/routes/(app)/cart/[id]/+page.server.ts
Comment (coyotte508) [reply src/routes/(app)/cart/[id]/+page.server.ts]: The initial reason is on cart page, both increase & decrease actions share the same form, and work even with JS disabled, this is why `quantity` is present in the form and is the initial value

and then this PR adds `by`, but we could add a boolean instead that "sets" to a target instead of increasing by.

src/routes/(app)/pos/touch/+page.svelte
coyotte508: it depends on the `Accept` header of the request. If the accept headers starts with `text/html` it will render a page, otherwise a JSON that looks like `{"type":"success","status":204,"data":"-1"}` or `{"type":"redirect","status":303,"location":"http://localhost:5508/pos/touch"}` or `{"type":"error","error":{"message":"Product can't be added to basket "}}`

So that when Javascript is disabled it still works, and when JavaScript is enabled the site feels more dynamic - when using `use:enhance` in a `<form>`

Note: here we need javascript for the POS interface always, just explaining how svelte works

roosemberth: Thanks for the explanation :+1:

src/routes/(app)/cart/[id]/+page.server.ts
Comment (roosemberth) [reply src/routes/(app)/cart/[id]/+page.server.ts]: > and then this PR adds by, but we could add a boolean instead that "sets" to a target instead of increasing by.

I think it would be easier to understand if there was simply another action to _set_ the amount. Instead of reusing this one.

If we want to tidy up the code in the actions, we could extract the common behaviour into a function.

src/routes/(app)/cart/[id]/+page.server.ts
Comment (roosemberth) [reply src/routes/(app)/cart/[id]/+page.server.ts]: Let's wait on the result of https://github.com/be-BOP-io-SA/be-BOP/pull/2068/files#r2371562986 before continuing here.

src/routes/(app)/cart/[id]/+page.server.ts
```diff
@@ -111,6 +107,64 @@ export const actions = {
...
+
+		const cart = await getCartFromDb({ user: userIdentifier(locals) });
+
+		if (!cart) {
+			throw error(404, 'No cart found for user');
+		}
+
+		const formData = await request.formData();
+
+		const max = product.maxQuantityPerOrder || DEFAULT_MAX_QUANTITY_PER_ORDER;
```
Comment (roosemberth) [inline src/routes/(app)/cart/[id]/+page.server.ts]: Offline discussion with @Tirodem: This hard limit currently exists to prevent griefing. For somebody unaware of the feature, it's unintuitive.

We discussed it will be addressed in a separate issue, where `DEFAULT_MAX_QUANTITY_PER_ORDER` is replaced by a runtime config item. We won't be addressing this on this PR.

There's no need to make any code changes.

roosemberth:
@Tirodem this PR can be merged when convenient (if no QA digression).

src/routes/(app)/cart/[id]/+page.server.ts
Comment (coyotte508) [reply src/routes/(app)/cart/[id]/+page.server.ts]: your comment made me notice: 

```suggestion
		const max = product.maxQuantityPerOrder ?? DEFAULT_MAX_QUANTITY_PER_ORDER;
```

Tirodem:
QA on its way.
I've created https://github.com/be-BOP-io-SA/be-BOP/issues/2073 and https://github.com/be-BOP-io-SA/be-BOP/issues/2072 following my previous tests, and https://github.com/be-BOP-io-SA/be-BOP/issues/2071 .

PR #2074, opened 2025-09-24
fix: improve bitcoin nodeless management (#1951)

### LINKED ISSUE #1951: 🦋 Improve bitcoin nodeless management

Tirodem:
- Add temporal filter for transactions between order creation and expiration
- Add admin warning for already used next addresses
- Skip used addresses during invoice generation
- Extract persistConfigElement to shared utility
- Improve error messages and API efficiency
- Make temporal filter parameters mandatory
- Extract updateBitcoinBlockHeight function
- Use upsert pattern for robust config handling
- Log address check errors in generateDerivationIndex

Fixes #1951


roosemberth:
Converted to draft, as this PR is simply a mechanism to test.

PR #2077, opened 2025-09-24
Investigate SSO + full discount problem

### LINKED ISSUE #2076: 🐛 100% discount with subscription goes 400 "You can't use free payment method on this order" if subscriber is logged with SSO / Wordpress
Image

Image

Image

If I'm logged through npub or email, it'll be good :

Image

If I'm logged with SSO (Wordpress), it's not ❌ 
Image

Sadly no error in log.

coyotte508:
Fix #2076

Fix SSO matching in createOrder for discounts
    
Also add secondaryEmail mathcing (SSO email)
    
Added a type safeguard so if new keys are added to UserIdentifier, they're forced to be passed to createOrder

### DIFF
--- a/src/routes/(app)/checkout/+page.server.ts
+++ b/src/routes/(app)/checkout/+page.server.ts
@@ -465,16 +465,7 @@ export const actions = {
 				paymentMethod,
 				{
 					locale: locals.language,
-					user: {
-						sessionId: locals.sessionId,
-						userId: locals.user?._id,
-						userLogin: locals.user?.login,
-						userRoleId: locals.user?.roleId,
-						userAlias: locals.user?.alias,
-						npub: locals.npub,
-						email: locals.email,
-						userHasPosOptions: locals.user?.hasPosOptions
-					},
+					user: userIdentifier(locals),
 					notifications: {
 						paymentStatus: {
 							npub: npubAddress,
--- a/src/lib/server/user.ts
+++ b/src/lib/server/user.ts
@@ -57,7 +57,7 @@ export async function createSuperAdminUserInDb(login: string, password: string) 
 export function userIdentifier(locals: App.Locals): UserIdentifier {
 	const secondaryEmails = locals.sso?.flatMap((sso) => (sso.email ? [sso.email] : []));
 	return {
-		ssoIds: locals.sso?.map((sso) => sso.id),
+		ssoIds: locals.sso?.map((sso) => `${sso.provider}:${sso.id}`),
 		userId: locals.user?._id,
 		email: locals.email,
 		secondaryEmails,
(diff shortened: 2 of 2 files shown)


Tirodem:
WIP

Tirodem:
```
free payment method used on an order with a total price > 0 {
  totalSatoshis: 97235,
  items: [
    {
      product: 'Salle Aquarium',
      quantity: 1,
      discountPercentage: undefined
    }
  ],
  paidSubs: [],
  discounts: [],
  discountSubIds: Set(0) {},
  usedSubIds: [],
  priceInfo: {
    currency: 'SAT',
    discount: 0,
    partialPrice: 89949,
    partialPriceWithVat: 97235,
    partialVat: 7286,
    perItem: [
      {
        amount: 80,
        amountWithoutDiscount: 80,
        currency: 'CHF',
        unitsToAccount: 1,
        unitsToBill: 1,
        usedFreeUnits: 0
      }
    ],
    physicalVatAtCustoms: false,
    singleVatCountry: true,
    totalPrice: 89949,
    totalPriceWithVat: 97235,
    totalVat: 7286,
    vat: [
      {
        country: 'CH',
        rate: 8.1,
        price: { amount: 6.48, currency: 'CHF' },
        partialPrice: { amount: 6.48, currency: 'CHF' }
      }
    ],
    vatRates: [ 8.1 ]
  },
  user: {
    sessionId: '77290dc36f9ffcd460fbb3b6d1fba6b7c7e5e84a319c137f98a189dff38b396c',
    userId: undefined,
    userLogin: undefined,
    userRoleId: undefined,
    userAlias: undefined,
    npub: undefined,
    email: undefined,
    userHasPosOptions: undefined
  }
}
```

I've kept the `token{}` logs if needed.

coyotte508:
yea so the issue is that the backend can't find the sub at all

Tirodem:
Here we go :
image

I'll do some more testing after @roosemberth will do the code review, but at least I'll be able to do my working demo.

src/routes/(app)/checkout/+page.server.ts
```diff
@@ -473,8 +474,12 @@ export const actions = {
 						userAlias: locals.user?.alias,
 						npub: locals.npub,
 						email: locals.email,
-						userHasPosOptions: locals.user?.hasPosOptions
-					},
+						secondaryEmails: locals.sso
+							?.map((sso) => sso.email)
+							.filter((email) => email !== undefined),
+						userHasPosOptions: locals.user?.hasPosOptions,
+						ssoIds: locals.sso?.map((sso) => sso.id)
```
roosemberth: I wouldn't add ssoIds. I think it's not useful, since it's an internal identifier to some SSO. If the SSO is later removed (or worse, replaced by a different one) the id will be useless.

If you have a good idea why we should keep it, we could keep it (perhaps namespaced by the sso).

The rest LGTM.

coyotte508: > If the SSO is later removed (or worse, replaced by a different one) the id will be useless.

Yes removing a SSO will remove access for all users that used it exclusively to log in, it should not be done lightly.

Fixed the first commit to just use `userIdentifer`, and good point about the provider, added a second commit to prefix SSO ids with the provider. (there's a temporary potential edge case if the sso ids contain `:` but wholly harmless IMO)

In any case storing sso ids is vital for me, potentially a user can use their sso id as only identifier. They should still be able to find their list of orders, ...

or cc @Tirodem if we don't care about personal identity info / past subscriptions related to SSO (and ongoing carts) we can directly remove the backward compatibility fix. I don't know how widely SSO is used at the moment

PR #2087, opened 2025-09-25
In SSO settings, display correct redirect URL by adding provider slug

coyotte508:
image

Noticed when testing SSO

Add the provider slug between "/oauth" and "/callback"


Tirodem:
Ahah, funny as it works as it is (as it's setted) on our first pilot with generic OAuth.
I'm merging and will check tomorrow on their install.

Issue #2007, opened 2025-07-17
🦋 Allow for the admin to cancel subscription for someone


Tirodem:
Use case : coworking, someone leaving one of the customer company.

coyotte508:
>  for inactive subscription add an option to trigger the renewal for the customer (with link to /subscription/{id} )

can you clarify exactly what you want? just send a message that the subscription expired and give the link to the subscription? (eg the same message as when the subscription expires)

by the way emails are not sent at the moment for subscription expires/reminder, only nostr messages:

https://github.com/be-BOP-io-SA/be-BOP/blob/766d39447a4ee701fc38fed70aca139783516e4e/src/lib/server/locks/subscription-lock.ts#L69-L88

https://github.com/be-BOP-io-SA/be-BOP/blob/766d39447a4ee701fc38fed70aca139783516e4e/src/lib/server/locks/subscription-lock.ts#L136-L155

PR #2089, opened 2025-09-25
fix subscription status in admin UI

coyotte508:
Fix https://github.com/be-BOP-io-SA/be-BOP/issues/2007

The first commit comes from https://github.com/be-BOP-io-SA/be-BOP/pull/2088


Tirodem:
https://github.com/be-BOP-io-SA/be-BOP/issues/2091 opened on the fly (not to do on this issue)
Cancelling = addition of field `canceledAt` in `subscriptions.paid`
Otherwise :
- cancellation works
- expiration notification is not sent if cancellation exists
- expiration notification is sent if cancellation doesn't exist

I wonder what happens when you renew a subscription while an old cancellation exists 👀

Tirodem:
It works ! The Cancellation line is removed when the subscription is renewed.
@roosemberth functional is OK for me, merge if code review is OK

roosemberth (approved):

Issue #1977, opened 2025-06-13
🦋 Allow reporting per tag


Tirodem:
# Context

We'll use tags as product family to filter results in Transactions > Reporting to have specific reports.

For examples :
- restaurants : distinguish drinks vs kitchen incomes
- bars : distinguish softs vs alcoholic brewages
- shop : distinguish clothes vs shoes vs accessories
- pvh-editions : distinguish book collection 1, book collection 2 and goodies (or either distinguish books per author)

# Requirement

Because tags are used for a lot of purposes, let's add this option on admin tag page :

<img width="407" height="535" alt="Image" src="https://github.com/user-attachments/assets/2a906408-1a45-42c0-95cb-05a775035cae" />

# Design

In Transactions > Reporting :

<img width="274" height="288" alt="Image" src="https://github.com/user-attachments/assets/6261b0d3-d834-47d9-9b47-151349f6c9fc" />

<img width="1037" height="378" alt="Image" src="https://github.com/user-attachments/assets/ccc1e827-173f-404c-8b53-3cda761df8ef" />

For filter effect :

## Order detail
List only orders including products with the said tag.

## Product detail
List only products with the said tag.

## Payment detail
List only payments from orders including products with the said tag.

## Order synthesis
List only orders including products with the said tag.
Displayed amounts are from the whole order, whether there's only products with this tag or other products in it.

## Product synthesis
List only products with the said tag.

## Payment synthesis
List only payments from orders including products with the said tag.
Displayed amounts are from the whole payment, whether there's only products with this tag or other products in it.

## VAT synthesis
List and synthetize VAT only for products with the said tag.

## Delivery fees
List and synthetize delivery fees only for products with the said tag.

coyotte508:
do we need the UI on the tag admin, is it better than being able to select any tag?

PR #2090, opened 2025-09-25
add 'available as filter for reporting' in admin UI for tags

coyotte508:
Fix https://github.com/be-BOP-io-SA/be-BOP/issues/1977

Note: the tag needs to be added to the product **before** the order is created. 

It can be done differently but then we need careful thought to handle all, cases (what happens when a product had the tag in the past and not anymore? What about deleted products?)


Tirodem:
A retroactive effect would be great but from what I see we track the product's tags on orders.items.n.product.tagIds so it would probably need a synchronization to update every order with every product tag, and I fear it might be performance-consumming.

We can go with no retrocompatibility for now and we'll address the ticket if merchants ask for it in the future.

BTW, to answer your questions :

> What happens when a product had the tag in the past and not anymore?

With the orders<>products sync mechanism, the product should lose its tag in the order informations and should not be fetched on filtered reporting anymore for this tag.

> What about deleted products ?

I need to check first how's reporting manages deleted products.
🚧 
Answer : product loses its picture on receipt / order but still appears on reporting without issue (better safe than sorry).
I don't know for this point and might need some pilots feedback on this.
Removing products from reporting on tag would be the most logical for me and the most compatible with the previous point, but who knows...

Otherwise, I found a little leak on the feature.
When filtering reporting with a tag, order is correctly filtered, payment also, but :
- product detail & product synthesis should only includes tagged product (now, it's every product from the order)
- same for VAT synthesis and delivery fees, if it's not too complex (otherwise it can be nice-to-have only and outside MVP)

And, for the last, but that one's on me as I forgot it on the initial issue : order synthesis might need a small improvement when filtered :
image
That way the merchant knows both order total with every product from orders including "drink" product, and also order total for every "drink" product only.
Either this, or that one, depending of your preference (both work for me) :
image
The "total" line would not be exported on .csv, and can be displayed on Product synthesis even when reporting is not filtered.

coyotte508:
> * product detail & product synthesis should only includes tagged product (now, it's every product from the order)
> 
> * same for VAT synthesis and delivery fees, if it's not too complex (otherwise it can be nice-to-have only and outside MVP

To clarify:

- [x] product details should include only tagged product (not other products in same orders)
- [x] product synthesis should include only tagged product (not other products in same orders)
- [x] VAT synthesis should include only tagged products
- Delivery fees should include only tagged products <= **can't do**, delivery fees are not itemized
- [x] orders should have two modes, one for order total and one where only product is taken into account

src/lib/server/orders.ts
```diff
@@ -76,42 +76,8 @@ import { binaryFindAround } from '$lib/utils/binary-find';
...
-		discount: undefined,
-		freeProductUnits,
-		userCountry: undefined,
-		vatExempted: false,
-		vatNullOutsideSellerCountry: runtimeConfig.vatNullOutsideSellerCountry,
-		vatProfiles,
-		vatSingleCountry: runtimeConfig.vatSingleCountry
-	});
-}
-
```
coyotte508, line 114: Removed this and instead created functions in `types/Order.ts` to compute info needed. They can be called in the backend or frontend but always with the same data, the data at the moment of the order, and the logic is centralized now.

Took into account free units too. And added unit tests

This will avoid issues if params change, eg exchange rates, VAT profiles, discounts, subscriptions, etc.

Needed for per-product VAT reporting in the admin, as per-item logic is now needed there.

src/app.css
```diff
@@ -178,9 +178,15 @@
...
+	fieldset:disabled label:has(.form-checkbox) {
+		@apply cursor-not-allowed opacity-50;
+	}
+
+	/* Fallback for browsers that don't support :has - but do not apply opacity-50 to avoid stacking with opacity-50 from :has */
 	.form-checkbox:disabled,
 	fieldset:disabled .form-checkbox {
-		@apply cursor-not-allowed opacity-50;
+		@apply cursor-not-allowed;
 	}
```
coyotte508, line 190: So that if a checkbox is disabled, their label is greyed out too

coyotte508:
image

also improved UI for export as CSV & download as PDF

src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.svelte
```diff
@@ -444,59 +495,54 @@
...
-								</td>
-								<td class="border border-gray-300 px-4 py-2">{data.currencies.main}</td>
-								<td class="border border-gray-300 px-4 py-2"
-									>{(toCurrency(
-										data.currencies.main,
-										(item.customPrice?.amount ?? item.product.price.amount) *
-											(item.discountPercentage ? (100 - item.discountPercentage) / 100 : 1),
-										item.customPrice?.currency ?? item.product.price.currency
-									) *
-										(item.product.deposit?.percentage ?? 100) *
```
Comment (coyotte508) [inline src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.svelte:480]: Note @Tirodem : I removed the prorata of product price * deposit in "Product details" - it doesn't take into account if the product was actually paid with a deposit or not, or even if it was, if the rest of the order was paid later on.

it's very easy to add back (just add back this line)

Tirodem:
I think there's an issue with order synthesis filtered.
https://dev-bootik.pvh-labs.ch/order/02eaa461-21c0-425e-bb3b-73c6baa39552
image
If I change tag :
image
image
The order total / average cart seems stuck at "43", with is good for my first tag, but not the second.

I continue tests on other points.

Tirodem:
Nice CSV/PDF new UI btw

Tirodem:
The rest seems good 👌 so we only have this filtered order synthesis issue to double-check.

coyotte508:
should be good

Tirodem:
Noice.
It's good as it is, but in the future we need to do a ticket to display either VAT incl or excl (if I'm not mistaked, filtered by product result on order synthesis is displayed VAT excl while total order synthesis is VAT incl).
Thanks coyotte, see you soon 🔥

Tirodem:
@roosemberth for code review, tests finished

and I tracked https://github.com/be-BOP-io-SA/be-BOP/issues/2093 for future needs

roosemberth (approved):
Brilliant! Thank you!

roosemberth:
I did not independently check the accuracy of the computation, but I trust the functional review.

Issue #2067, opened 2025-09-22
🛠️ On checkout, if mempool API rate limit is reached, don't propose BTC on-chain as a valid payment method


Tirodem:
Following #2063 tests.
TBC

Tirodem:
To check with https://github.com/be-BOP-io-SA/be-BOP/issues/1493

PR #2099, opened 2025-10-01
feat: improve bitcoin nodeless management (#1951)

### LINKED ISSUE #1951: 🦋 Improve bitcoin nodeless management

gntw:
- Add temporal filter for transactions between order creation and expiration
- Add admin warning for already used next addresses
- Skip used addresses during invoice generation
- Add skipUsedAddresses checkbox in admin UI
- Extract persistConfigElement to shared utility
- Improve error messages and API efficiency
- Make temporal filter parameters mandatory
- Extract updateBitcoinBlockHeight function
- Use upsert pattern for robust config handling

Fixes #1951


roosemberth:
Ping @Tirodem for QA.

Tirodem:
Points 2-3 selector is functional ✅ 
Point 1 is functional ✅ 
Point 2 / 3 selector is functional ✅ 
However, I've hit the API rate limit so hard (by displaying the nodeless admin page, then change the zpub, then refresh it) that I fear it might break on-chain payment check too quick.

I have two last requests to mitigate this :
- [ ] instead of displaying & checking 10 addresses on Payment Settings > Bitcoin-nodeless, let's only display and check 5 adresses
- [ ] I don't know how much it might be complicated, but for /admin/bitcoin-nodeless, if an address was checked in the last 10 minutes, don't check it again (in theory derivation index will do the job, if you use your address with our wallet outside be-BOP, it won't happen in less than 10 minutes) (btw that should apply only on /admin/bitcoin-nodeless, not on on-chain nodeless payment)

If it's done, I think it could be merged to main, then we'll go soon after with #2067 and #2068 (or a new version of 2068).

@roosemberth do we have any information about mempool API rate limit and how long API access would be denied ?

Tirodem:
As seen, let's to the 10 > 5 addresses on this PR.
The other point will be adressed on a new issue : #2104

src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.server.ts
```diff
@@ -1,31 +1,57 @@
...
+			bitcoinNodeless: runtimeConfig.bitcoinNodeless,
+			nextAddresses: [],
+			hasAlreadyUsedNextAddresses: false
+		};
+	}
+
+	const ADDRESS_CHECK_TIMEOUT_MS = 3000;
+
+	const nextAddresses = await Promise.all(
+		Array.from({ length: 10 }, async (_, i) => {
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.server.ts]:
```suggestion
		Array.from({ length: 5 }, async (_, i) => {
```

Per Christophe's request.

roosemberth (approved):

PR #2100, opened 2025-10-01
🐛 Fix schedule event checkboxes not working on existing events #2054

### LINKED ISSUE #2054: 🐛 Some Schedule event options aren't usable on existing event
- "Add RSVP option" : can be enabled only on creation. On an existing event with option disabled, checking the option doesn't work (no explicit console message, just no result)

Image

Image

- "Event has custom color on calendar" : same

Image

Image

- "Make event unavailable (postponed, cancelled, sold out)", which is not proposed on event creation and only on existing event, cannot be checked on existing events like the two other options

Image

I've checked from here difference between checkboxes but I don't see the difference.
Tests made on Chrome Version 140.0.7339.80 (Build officiel) (64 bits) on Windows 10.

gntw:
Problem: Checkboxes for RSVP, calendar color, and unavailability weren't working when editing existing schedule events. The checkboxes appeared to not respond to clicks.

Root cause: Commit cd09df8 changed `let` declarations to reactive statements (`$:`), which breaks Svelte's two-way binding for checkboxes. Reactive statements recreate arrays on every dependency change, losing user input.

Solution:
- Revert to `let` declarations for checkbox state arrays
- Add reactive sync when data.schedule.events length changes
- This preserves checkbox bindings while handling picture uploads

Fixes #2054

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/schedule/[id]/+page.svelte
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/schedule/[id]/+page.svelte
@@ -14,16 +14,32 @@
 	let slug = data.schedule._id;
 	let displayPastEvents = data.schedule.displayPastEvents;
 	let eventLines = data.schedule.events.length || 1;
-	$: eventAvailable = data.schedule.events.map((eve) => ({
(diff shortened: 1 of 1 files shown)


Tirodem:
Works for me ✅

PR #2128, opened 2025-10-11
🐛 Fix incorrect project root path detection on production builds

roosemberth:
This bug resulted in the IP location databases not being loaded.

Additionally be-BOP will now also display a warning message if it's unable to load the IP location databases.


Tirodem:
image

Issue #2141, opened 2025-10-16
🦋 allow user to choose payment method on subscription renewal


CypherJack:
TBD

Tirodem:
Duplicate from #2114

PR #2110, opened 2025-10-05
🐛 fix error when renewing a subscription

roosemberth:
The error would previously trigger when a user tries to renew a subscription.

Closes https://github.com/be-BOP-io-SA/be-BOP/issues/2046.

PR #2145, opened 2025-10-16
🦋Add admin interface for managing product subscriptions:

### LINKED ISSUE #2115: 🦋 Allow to create subscriptions for users manually in back-office
Some of our pilots use the subscription system on be-BOP but migration is painful as they have available ongoing subscriptions on their current system and have to report it to be-BOP.

But right now unless you create subscription orders with 100% discount or manually create items in `subscriptions.paid`, you can't do it.

So, here's what we need :

Image

When adding a subscriber :

Image

When importing CSV :

Image
For now :
- column 1 : `subscriptions.paid.user.email`
- column 2 : `subscriptions.paid.user.npub`
- column 3 : `subscriptions.paid.paidUntil`

gntw:
- Manual subscriber addition (multi-line form, up to 100 at once)
- CSV bulk import/export with validation (10K row limit)
- Individual and bulk delete/cancel operations
- Optimized performance (single DB call for bulk number generation)
- Zod validation, functional patterns, all-or-nothing import

Closes #2115

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts
@@ -6,7 +6,6 @@ import type { UserIdentifier } from '$lib/types/UserIdentifier';
 import type { Actions } from './$types';
 import { z } from 'zod';
 
-// @ts-expect-error - SvelteKit auto-generates types for params
 export const load = async ({ params }) => {
 	const product = await collections.products.findOne({ _id: params.id });
 	const subscriptions = await collections.paidSubscriptions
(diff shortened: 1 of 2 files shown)


Tirodem:
Seems to work nicely.
I have to go a bit deeper, I'll test again in 2 days.

Is it me or do I miss the "Cancel" CTA on individual line ?
It was this one previously :
image

However, I though about something wednesday : this feature should need another feature, maybe 2, to be really usable.

Firstly, we still have https://github.com/be-BOP-io-SA/be-BOP/pull/2110 here for renewal. We still have an issue with free (discounted) subscriptions renewal.

Secondly, while testing a subscription created on #2145 then renewed with #2110 , I see that a manually created subscription ended up in error, so I guessed we need this :

- #2147 (my guess is that, when we renew, we fetch orders corresponding to this subscription, then we clone the order -payment method included- then we create and push the new order to the customer ; so, when a subscription is created manually with #2145 , no payment method is associated, so a manually created subscription leads to an error when renewal is attempted

OR (or, why not, and, in term of comfort, but it can replace #2141 as a first step)

- #2141 (as in the future it could be good to pay 6 months of subscription with lightning, 7th and 8th month in paypal, then Lightning again, by giving the user an option to change the payment method ; if no payment method is setted on subscription renewal, it'll become a "choose your payment method" user journey and can answer our current progblem)

Put this ticket aside for a few days, we'll check that soon after a discussion with Roos (and feel free to give your opinion !).

roosemberth:
@Tirodem not merging since you mentioned you'd like to re-test the branch later.

Merge when convenient.

Issue #2130, opened 2025-10-12
💡 Allow phoenixd configuration through envvar ?


Tirodem:
Following #1546 but also following the first release of the wizard, a little thought of mine.
Currently, phoenixd password is setted through UI in /admin/phoenixd
This password is either fetch manually on VPS or given by the wizard at the end of the install procedure.
If one day "we" (us or a partner) provide a full non-SSH setup with wizard through a procedure where the user doesn't use SSH, this user won't be able to fetch his phoenixd password.
So, here's my thought : why not allow, in addition to current implementation, to preset phoenixd password through envvar (and so we'll set it with wizard directly from ~/.phoenix/phoenixd.conf ) ?
Phoenixd would still need the "Detect" CTA interaction in /admin/phoenixd but the password would be autofilled from envvar if phoenixd is up.

roosemberth:
Closed by #2146.

/close

PR #2146, opened 2025-10-16
💡 Enable phoenixd configuration using environment variables

roosemberth:
The _phoenixd_ configuration can now be set through environment variables.

### DIFF
--- a/.env
+++ b/.env
@@ -46,6 +46,9 @@ NO_LOCK=
 NOSTR_PRIVATE_KEY=
 # The url of your app
 ORIGIN=
+# The phoenixd endpoint URL and password; this will be used to autoconfigure phoenixd, if specified.
+PHOENIXD_ENDPOINT_URL=
+PHOENIXD_HTTP_PASSWORD=
 # Set to "true" if the S3 server is located very close to the bebop server (same datacenter, server)
 # This will proxy S3 downloads through the bebop server, which allows better caching.
 #
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -44,6 +44,7 @@ import { renewSessionId } from '$lib/server/user';
 import { typedInclude } from '$lib/utils/typedIncludes';
 import { rateLimit } from '$lib/server/rateLimit';
 import { toIPv4Maybe } from '$lib/server/utils/toIPv4Maybe';
+import { attemptAutoconfigurePhoenixd } from '$lib/server/phoenixd';
 
 const SSO_COOKIE = 'next-auth.session-token';
 
@@ -509,3 +510,6 @@ const handleSSO = authProviders
 export const handle = handleSSO
 	? sequence(addSecurityHeaders, handleGlobal, handleSSO, handleSsoCookie)
 	: sequence(addSecurityHeaders, handleGlobal);
+
+// Kick-off autoconfiguration of phoenixd
+attemptAutoconfigurePhoenixd();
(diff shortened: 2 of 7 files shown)


roosemberth:
Closes https://github.com/be-BOP-io-SA/be-BOP/issues/2130.

roosemberth:
If this feature is used, both the “Reset” button and the password field in the admin UI become useless since the phoenixd URL will be replaced at the next start.

@Tirodem what should we do here? Should we disable the UI elements? Should we simply not display it?

image

Tirodem:
> If this feature is used, both the “Reset” button and the password field in the admin UI become useless since the phoenixd URL will be replaced at the next start.
> 
> @Tirodem what should we do here? Should we disable the UI elements? Should we simply not display it?

We should remove the input, the "save" and the "reset" CTA and display a message like "Your phoenixd settings are managed by environnement variables and cannot be edited or displayed on be-BOP backoffice."

roosemberth:
> We should remove the input, the "save" and the "reset" CTA and display a message like "Your phoenixd settings are managed by environnement variables and cannot be edited or displayed on be-BOP backoffice."

Applied suggestion.

Tirodem:
Before merging to use with wizard I wanted to try it on an already deployed pilot, so I disabled the phoenixd current configuration then update the .env.local then rebotted :
```
PHOENIX_ENDPOINT_URL=http://host.docker.internal:9740
PHOENIX_HTTP_PASSWORD=hiddenpasswordhere
```
(yup, it's a docker compose pilot)
When I go to /admin/phoenixd , I have the legacy UI (with reset button and all).
After reset, if I detect phoenixd :
image
image
So values in .env.local don't seem to be called.
https://www.b-onlyfans.com/.well-known/version.txt = a47f38db86215e9433ce07214f6976bd2209a5b1
Did I miss something ?

roosemberth:
> PHOENIX_ENDPOINT_URL=http://host.docker.internal:9740
> PHOENIX_HTTP_PASSWORD=hiddenpasswordhere

The names of the environment variables seem incorrect. It should be PHOENIXD_ENDPOINT_URL and PHOENIXD_HTTP_PASSWORD.

Tirodem:
Ah, yes, the missing D.
@CypherJack , note that for your onboarding, this issue was about the will of the D.

Tirodem:
`In the presence of a non-empty PHOENIXD_ENDPOINT_URL environment variable. I attempted to reach phoenixd at http://host.docker.internaz:9740, but failed. Please check the value of the PHOENIXD_ENDPOINT_URL variable`

and

```
Unable to conclude phoenixd configuration: HttpError {
  status: 500,
  body: {
    message: 'Could not get lnaddress: 401 Invalid authentication (use basic auth with the http password set in phoenix.conf)'
  }
}
```

I wanted to check that.
If OK, it works, if absent, it works, if wrong, error managed.
If URL is good but HTTP password is bad in .env.local, you have to clean your .env.local :
image
I know it's bad to have a bad value in .env.local but my feeling is that, if we have a 500 with phoenixd credentials (good url, bad password), we should not lock the UI and we should allow to put correct credentials here (so, ignore .env.local).

BTW, I'm merging as it is, I'll put a draft issue on the burner for that to discuss in a few days, so I'll be able to test this branch coupled with the wizard.

As long as we don't have regression, it's good.

Tirodem (approved):

Tirodem:
https://github.com/be-BOP-io-SA/be-BOP/issues/2152 for follow-up

PR #2148, opened 2025-10-17
🦋Enable subscription renewal reminder through email

### LINKED ISSUE #2045: 🦋 Enable subscription renewal reminder through email
- https://github.com/be-BOP-io-SA/be-BOP/blob/5ee3e9375e6fcf777bee55618b7d89b7a4988ca7/src/lib/server/locks/subscription-lock.ts#L81

- https://github.com/be-BOP-io-SA/be-BOP/blob/5ee3e9375e6fcf777bee55618b7d89b7a4988ca7/src/lib/server/locks/subscription-lock.ts#L87

After a quick analysis for a pilot launch, it seems that we only then notifications through Nostr :
Image

But not through email (reminders being stored as `"medium": "none"`) :
Image

We need to update be-BOP/src/lib/server/locks/subscription-lock.ts and create a new email template called `subscription.reminder` and (and `subscription.ended` if we warn of subscription ending when not renewed).
For memories, templates are here : https://github.com/be-BOP-io-SA/be-BOP/blob/5ee3e9375e6fcf777bee55618b7d89b7a4988ca7/src/lib/server/runtime-config.ts#L249 and shall be integrated in /admin/template/emails

gntw:
Added email notifications for subscription lifecycle events:
- Reminder email sent 24h before expiration
- Expiration notification when subscription ends
- Renewal confirmation email after successful payment

Updated notification medium type to support 'email' alongside 'nostr'. Fixed subscription renewal flow with proper user data handling.

Closes #2045

### DIFF
--- a/src/lib/server/locks/email-notifications.ts
+++ b/src/lib/server/locks/email-notifications.ts
@@ -1,10 +1,12 @@
-import type { ChangeStream, ChangeStreamDocument } from 'mongodb';
+import { MongoServerError, Timestamp, type ChangeStream, type ChangeStreamDocument } from 'mongodb';
 import { Lock } from '../lock';
 import { collections } from '../database';
 import type { EmailNotification } from '$lib/types/EmailNotification';
 import { isEmailConfigured, sendEmail } from '../email';
 import { building } from '$app/environment';
 import { rateLimit } from '../rateLimit';
+import { getUnixTime, subHours } from 'date-fns';
+import { typedInclude } from '$lib/utils/typedIncludes';
 
 const lock = new Lock('notifications.email');
 
@@ -66,24 +68,51 @@ async function handleEmailNotification(email: EmailNotification): Promise<void> 
 	}
 }
 
-let changeStream: ChangeStream<EmailNotification>;
+let changeStream: ChangeStream<EmailNotification> | null = null;
 
-function watch() {
+async function watch(opts?: { requestChangesSince?: Timestamp }) {
 	try {
 		rateLimit('0.0.0.0', 'changeStream.email-notifications', 10, { minutes: 5 });
 	} catch {
 		console.error("Too many change streams errors for 'email-notifications', exiting");
 		process.exit(1);
 	}
-	changeStream = collections.emailNotifications.watch([{ $match: { operationType: 'insert' } }], {
-		fullDocument: 'updateLookup'
-	});
-	changeStream.on('change', (ev) => handleChanges(ev).catch(console.error));
-	changeStream.on('error', (err) => {
-		console.error('email-notifications', err);
-		changeStream.close().catch(console.error);
-		watch();
-	});
+
+	try {
+		changeStream = collections.emailNotifications
+			.watch([{ $match: { operationType: 'insert' } }], {
+				fullDocument: 'updateLookup',
+				...(opts?.requestChangesSince && {
+					startAtOperationTime: opts.requestChangesSince
+				})
+			})
+			.on('change', (ev) => handleChanges(ev).catch(console.error))
+			.once('error', async (err) => {
+				console.error('change stream error', err);
+				changeStream?.close().catch(console.error);
+				changeStream = null;
+
+				if (
+					err instanceof MongoServerError &&
+					typedInclude(['ChangeStreamHistoryLost', 'ChangeStreamFatalError'], err.codeName)
+				) {
+					// Restart from 1 hour ago if history was lost
+					return watch({
+						requestChangesSince: Timestamp.fromBits(0, getUnixTime(subHours(new Date(), 1)))
+					});
+				} else {
+					return watch();
+				}
+			});
+
+		return changeStream;
+	} catch (err) {
+		if (err instanceof MongoServerError && err.codeName === 'ChangeStreamHistoryLost') {
+			console.warn('Oplog time out of range when starting change stream, falling back to now', err);
+			return watch();
+		}
+		throw err;
+	}
 }
 
 async function processEmailNotifications() {
--- a/src/lib/server/locks/subscription-lock.ts
+++ b/src/lib/server/locks/subscription-lock.ts
@@ -4,11 +4,239 @@ import { Lock } from '../lock';
 import { processClosed } from '../process';
 import { setTimeout } from 'node:timers/promises';
 import { refreshPromise, runtimeConfig } from '../runtime-config';
-import { ObjectId } from 'mongodb';
+import { type FindCursor, ObjectId } from 'mongodb';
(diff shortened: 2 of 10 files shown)


src/lib/server/locks/email-notifications.ts
```diff
@@ -62,24 +64,40 @@ async function handleEmailNotification(email: EmailNotification): Promise<void>
...
-		changeStream.close().catch(console.error);
-		watch();
-	});
+	changeStream = collections.emailNotifications
+		.watch([{ $match: { operationType: 'insert' } }], {
+			fullDocument: 'updateLookup',
+			...(opts?.operationTime && {
+				startAtOperationTime: opts.operationTime
+			})
+		})
```
roosemberth: This will fail if the time is outside the range of the oplog. This can be handled gracefully by making the start time advisory and defaulting to “now”; in this case, the argument should be renamed to something like `requestChangesSince`.

src/lib/server/locks/subscription-lock.ts
```diff
@@ -20,139 +21,198 @@ async function maintainLock() {
...
+				},
+				cancelledAt: { $exists: false },
+				'user.email': { $exists: true },
+				notifications: { $not: { $elemMatch: { type: 'reminder', medium: 'email' } } }
+			});
+
+			for await (const subscription of subscriptionsToRemindEmail) {
+				const userEmail = subscription.user.email;
+				if (!userEmail) {
+					continue;
```
roosemberth: We should at least log that we can't send the email notification. This way, if the customer complains they didn't receive their email, we'll be able to find evidence about this in the logs.

This comment also applies to end of subscription notifications.

src/lib/server/locks/subscription-lock.ts
```diff
@@ -20,139 +21,198 @@ async function maintainLock() {
...
-							{
-								$push: {
-									notifications: {
-										createdAt: new Date(),
-										_id: notifId,
-										type: 'reminder',
-										medium: 'nostr'
-									}
+					const notifId = new ObjectId();
+					await collections.nostrNotifications.insertOne(
```
roosemberth: Please move each of:

- Notify user subscription about to end via Nostr,
- Notify user subscription about to end via Email,
- Notify user subscription ended via Nostr,
- Notify user subscription ended via Email

Each to a separate function.

src/lib/server/orders.ts
```diff
@@ -2103,11 +2103,12 @@ async function applyOrderSubscriptionsDiscounts(order: Order, session: ClientSes
...
 			}
+			const newPaidUntil = add(max([existing.paidUntil, new Date()]), subscriptionDuration);
 			const result = await collections.paidSubscriptions.updateOne(
 				{ _id: existing._id },
 				{
 					$set: {
-						paidUntil: add(max([existing.paidUntil, new Date()]), subscriptionDuration),
+						paidUntil: newPaidUntil,
 						updatedAt: new Date(),
 						notifications: [],
```
roosemberth, line 2155: I think we should not forget the previous notifications we sent.

The code fetching what subscriptions to notify should also be updated.

src/lib/server/locks/subscription-lock.ts
```diff
@@ -20,139 +21,198 @@ async function maintainLock() {
 		}
 
 		try {
-			const subscriptionsToRemind = collections.paidSubscriptions.find({
+			const now = new Date();
+			const reminderWindow = addSeconds(now, runtimeConfig.subscriptionReminderSeconds);
+
+			const subscriptionsToRemindNostr = collections.paidSubscriptions.find({
```
roosemberth: Please move each of the queries generating:

- The subscriptions to remind via Nostr,
- The subscriptions to remind via email,
- The subscriptions to notify end via Nostr,
- The subscriptions to notify end via email

To separate functions. Please also add a unit test checking that:

- After purchasing a subscription, right before the subscription expires, it's returned by the “subscriptions to remind” function and when it expires it's returned by the subscriptions to nofity end.
- Same, but after renewing the subscription as it approaches the expiration of the renewal.

There's no need to test that the email was sent. Only that they were considered.

src/routes/(app)/order/[id]/+page.svelte
```diff
@@ -177,8 +177,8 @@
 								class="break-words {payment.status === 'paid'
 									? 'text-green-500'
 									: 'body-secondaryText'} "
-								amount={payment.price.amount}
-								currency={payment.price.currency}
+								amount={payment.price?.amount ?? payment.currencySnapshot?.main.price.amount}
+								currency={payment.price?.currency ?? payment.currencySnapshot?.main.price.currency}
```
Comment (roosemberth) [inline src/routes/(app)/order/[id]/+page.svelte]: Why is this fallback required?

src/lib/server/locks/email-notifications.ts
```diff
@@ -62,24 +64,51 @@ async function handleEmailNotification(email: EmailNotification): Promise<void>
...
+				if (
+					err instanceof MongoServerError &&
+					typedInclude(['ChangeStreamHistoryLost', 'ChangeStreamFatalError'], err.codeName)
+				) {
+					// Restart from 1 hour ago if history was lost
+					watch({
+						requestChangesSince: Timestamp.fromBits(0, getUnixTime(subHours(new Date(), 1)))
+					});
+				} else {
+					watch();
```
roosemberth: `return watch(...`

Else the closure will kickoff but lose the new promise; thus the promise returned to the calling function will resolve to undefined even if the underlying process has not completed.

Same in the other branch of the if expression.

src/lib/server/locks/subscription-lock.ts
```diff
@@ -4,12 +4,224 @@ import { Lock } from '../lock';
...
+}
+
+async function notifySubscriptionEndedViaNostr(subscription: PaidSubscription) {
+	await withTransaction(async (session) => {
+		const notifId = new ObjectId();
+		await collections.nostrNotifications.insertOne(
+			{
+				_id: notifId,
+				kind: Kind.EncryptedDirectMessage,
+				dest: subscription.user.npub,
```
roosemberth: We should add a sanity check that the user npub actually exists (like we do for emails) and log otherwise.

Tirodem: Is npub existence check is done elsewhere ? If not, do not.

Tirodem: (Btw it's a reminder so an order was already made with this npub)

roosemberth: If `subscription.user.npub` is undefined where is the relay supposed to send the message?

Here we simply push the task to the DB: if the address value is invalid, the statement will not fail here; instead, the error will appear later at the worker.

Issue #2124, opened 2025-10-09
🐛 Issue with VAT & price rounding


Tirodem:
<img width="1425" height="1033" alt="Image" src="https://github.com/user-attachments/assets/30f96a86-75f7-401f-8488-d037aeff3e3f" />
The product is 5 but with the 8.1 VAT we either have 4.63 which give 5.01 or 4.62 which gives 4.99.
I've tried to fill 4.6253 in database (products.price.amount) but it didn't change.
We should fix the issue if we're able to manage VAT price with 3 decimals.

PR #2156, opened 2025-10-21
🐛  Fix VAT calculation rounding on invoices and receipts

gntw:
Merchants can now set precise prices that result in round numbers after VAT is applied.

  Previously, setting a price like CHF 4.63 with 8.1% VAT would display incorrectly as CHF 5.01 instead of CHF 5.00 on invoices and receipts. This happened because the system rounded prices too early in
  calculations.

  Now you can set exact prices (like CHF 4.6253) that calculate to perfect round numbers with VAT (CHF 5.00), making invoices cleaner and more professional.

  Issue #2124

### DIFF
--- a/src/lib/components/PriceTag.svelte
+++ b/src/lib/components/PriceTag.svelte
@@ -31,46 +31,48 @@
 	const mainCurrency = $currencies.main;
 	const secondaryCurrency = $currencies.secondary;
 
-	$: actualCurrency = main ? mainCurrency : secondary ? secondaryCurrency : convertedTo ?? currency;
+	$: actualCurrency = (
+		main ? mainCurrency : secondary ? secondaryCurrency : convertedTo ?? currency
+	) as Currency | null;
 	$: actualAmount = actualCurrency === null ? 0 : toCurrency(actualCurrency, amount, currency);
 
-	$: displayedAmount = !force
-		? actualCurrency === 'BTC' && actualAmount < 0.01
-			? actualAmount * SATOSHIS_PER_BTC
-			: actualCurrency === 'SAT' && actualAmount >= 1_000_000
-			? actualAmount / SATOSHIS_PER_BTC
-			: actualAmount
-		: actualAmount;
-	$: displayedCurrency = !force
-		? actualCurrency === 'BTC' && actualAmount < 0.01
-			? 'SAT'
-			: actualCurrency === 'SAT' && actualAmount >= 1_000_000
-			? 'BTC'
-			: actualCurrency || 'BTC'
-		: actualCurrency || 'BTC';
+	// Auto-conversion thresholds: small BTC amounts show as SAT, large SAT amounts show as BTC
+	$: [displayedAmount, displayedCurrency] = (() => {
+		if (!force && actualCurrency === 'BTC' && actualAmount < 0.01) {
+			return [actualAmount * SATOSHIS_PER_BTC, 'SAT'] as const;
+		}
+		if (!force && actualCurrency === 'SAT' && actualAmount >= 1_000_000) {
+			return [actualAmount / SATOSHIS_PER_BTC, 'BTC'] as const;
+		}
+		return [actualAmount, actualCurrency || 'BTC'] as const;
+	})() as [number, Currency];
 
-	$: displayed =
-		displayedCurrency !== 'BTC' && actualAmount > 0 && displayedAmount < 0.01
-			? '< ' +
-			  Number(0.01).toLocaleString('en', {
-					style: displayedCurrency === 'SAT' ? undefined : 'currency',
-					currency: displayedCurrency === 'SAT' ? undefined : displayedCurrency,
-					maximumFractionDigits: 2,
-					minimumFractionDigits: 0
-			  })
-			: displayedAmount.toLocaleString('en', {
-					style:
-						displayedCurrency === 'SAT' || displayedCurrency === 'BTC' ? undefined : 'currency',
-					currency:
-						displayedCurrency === 'SAT' || displayedCurrency === 'BTC'
-							? undefined
-							: displayedCurrency,
-					maximumFractionDigits: displayedCurrency === 'BTC' ? 8 : 2,
-					minimumFractionDigits:
-						!Number.isInteger(displayedAmount) && displayedCurrency !== 'BTC'
-							? FRACTION_DIGITS_PER_CURRENCY[displayedCurrency]
-							: 0
-			  }) + (displayedCurrency === 'SAT' && !short ? ' SAT' : '');
+	// Helper: check if currency is crypto (SAT or BTC)
+	$: isCrypto = displayedCurrency === 'SAT' || displayedCurrency === 'BTC';
+
+	// Determine decimal places based on currency
+	$: maximumFractionDigits = FRACTION_DIGITS_PER_CURRENCY[displayedCurrency];
+
+	$: minimumFractionDigits =
+		!Number.isInteger(displayedAmount) && !isCrypto ? maximumFractionDigits : 0;
+
+	// Edge case: very small fiat amounts (> 0 but < 0.01) show as "< 0.01"
+	$: isSmallFiatAmount = !isCrypto && actualAmount > 0 && displayedAmount < 0.01;
+
+	// Build small formatting config - crypto gets no currency symbol, fiat gets it
+	$: formatConfig = {
+		maximumFractionDigits,
+		minimumFractionDigits,
+		...(isCrypto ? {} : { style: 'currency' as const, currency: displayedCurrency })
+	} satisfies Intl.NumberFormatOptions;
+
+	$: formattedAmount = isSmallFiatAmount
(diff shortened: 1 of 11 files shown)


Tirodem:
Test :
image
image
image
image
image
image
image
image
- :x: we need to edit proforma invoice ( https://dev-bootik2.pvh-labs.ch/order/8a3fa304-1a76-4959-87f2-9da00d7ac1d7/payment/68f7f63da40318372adce526/ticket and https://dev-bootik2.pvh-labs.ch/order/8a3fa304-1a76-4959-87f2-9da00d7ac1d7/payment/68f7f63da40318372adce526/receipt )

BTW, a new addition to the cart after an oder gave this :
image

I already had a 4.99 € situation with this VAT excluded price, and I had to change it back to have the correct price.

Here are some examples :
image
image

If I play a bit with the VAT excluded price without the VAT calculator :
image
image
image
image

And at the same time with 4,6253 and 8.1% I can have :
image
image

So, my guess is :
- there's some random shit 🤷‍♂️ 
- we might fix this by updating the VAT calculator and making something with the rounding (still the exact value is 4.62534690102 so 4.6253 is nearer than 4.6254)
- we need to fix receipts and tickets

Tirodem:
Tests OK ✅

src/lib/types/Currency.ts
```diff
@@ -41,6 +41,32 @@ export const FRACTION_DIGITS_PER_CURRENCY = Object.freeze({
...
 }) satisfies Record<Currency, number>;
 
+/**
+ * Maximum number of decimal places for storing prices WITHOUT VAT.
+ * This allows merchants to set precise base prices that result in round
+ * numbers after VAT calculation (e.g., 4.6253 CHF → 5.00 CHF with 8.1% VAT).
+ *
+ * Display precision remains FRACTION_DIGITS_PER_CURRENCY (2 for most currencies).
+ */
+export const STORAGE_FRACTION_DIGITS_PER_CURRENCY = Object.freeze({
```
roosemberth: This is a representation detail and must be persisted alongside with the price.

If we change it in the future, the interpretation of whatever we stored in the database should not depend on a code constant; else we'll have to implement heuristics to guess what we meant to store or break backwards compatibility.

We need to store the price precision along with the amount currency and value.

It would be useful to define a "currency amount" type that contains the currency, value and precision.

Symbol `STORAGE_FRACTION_DIGITS_PER_CURRENCY` should not be exported, export util functions to project and reconstruct amounts.

src/lib/types/Currency.ts
```diff
@@ -61,17 +88,43 @@ export const CURRENCY_UNIT = Object.freeze({
...
+
+	return {
+		amount: rounded,
+		currency,
+		...(useStoragePrecision && { precision })
+	};
+}
+
+//Get the storage precision (number of decimal places) for a currency.
+export function getStoragePrecision(currency: Currency): number {
```
roosemberth: This function basically exports why array. Why is the precision needed elsewhere?

To keep encapsulation, we should use functions like `createPrice` that keep the representation details abstracted away.

src/lib/components/ProductForm.svelte
```diff
@@ -111,7 +111,7 @@
 	let vatRate = 0;
 	$: product.price.amount = Number(
 		((100 * priceAmountVATIncluded) / (100 + vatRate)).toFixed(
-			FRACTION_DIGITS_PER_CURRENCY[product.price.currency]
+			getStoragePrecision(product.price.currency)
```
roosemberth: This is a price computation, we should use the createPrice function.

src/lib/components/PriceTag.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import {
 		FRACTION_DIGITS_PER_CURRENCY,
+		getStoragePrecision,
```
roosemberth: Why does the price tag display performs price computation?

This is a display component, it shouldn't care about how the price is stored.

gntw:
OMG, false start. I switched between branches many times and forgot to apply my last stash before pushing :(
Of course, createPrice should be used for price computations , fixed now

Tirodem:
@roosemberth can you confirm a last code review ? We need to deploy it on pilots tomorrow, I'd like to merge this afternoon if OK.

roosemberth:
(Rebased branch to remove accidental merge commit therein.)

src/lib/components/PriceTag.svelte
```diff
@@ -21,6 +21,11 @@
 	 * Convert to the secondary currency
 	 */
 	export let secondary = false;
+	/**
+	 * Optional precision for displaying the amount. If provided, uses this precision
+	 * instead of the default FRACTION_DIGITS_PER_CURRENCY.
+	 */
+	export let precision: number | undefined = undefined;
```
roosemberth: As far as I can tell, this is used nowhere.

Why do we support `precision` to be undefined? Wouldn't it be simpler to support a single form of price? It would certainly eliminate this issue.

roosemberth: I suspect, there should be a function to interpret the price from the database (be it order or otherwise) that defaults to the “storage price” if the price is not set.

src/lib/types/Currency.ts
```diff
@@ -61,17 +88,38 @@ export const CURRENCY_UNIT = Object.freeze({
...
 	}
 
 	return priceAmount;
 }
 
+// Returns Price object with amount rounded to opt. specified precision
+export function createPrice(
+	amount: string | number,
+	currency: Currency,
+	useStoragePrecision = false
```
roosemberth: Why is this argument optional?

As far as I can tell it's always set to `true`.

I suspect the intention was to call this function to freeze the price for storage purposes. In it's current form, this function fullfils two roles: To compute “some” price for storage purposes and to compute “some” price for display purposes.

Perhaps it would be simpler to use two separate functions? That way the problem where we “forget” to specify what we want cannot exist and reading the calling code would be simpler (e.g. `computePriceForStorage`, `computePriceForDisplay`)?

src/lib/types/Order.ts
```diff
@@ -23,6 +23,7 @@ export type DiscountType = 'fiat' | 'percentage';
 export type Price = {
 	amount: number;
 	currency: Currency;
+	precision?: number;
```
roosemberth, line 46: We should add a note here, that this property is optional for backwards compatibility reasons). Right now it is not clear what the precision is when this value is undefined. We should also add a comment about this too.

src/lib/utils/toStorageCurrency.ts
```diff
@@ -0,0 +1,34 @@
...
+import { FRACTION_DIGITS_PER_CURRENCY, createPrice, type Currency } from '$lib/types/Currency';
+import { get } from 'svelte/store';
+
+/**
+ * Convert amount to target currency with storage precision (4 decimals for fiat, 8 for BTC).
+ * Use this when storing prices in database to preserve full precision for VAT calculations.
+ *
+ * For display purposes, use toCurrency() which rounds to FRACTION_DIGITS_PER_CURRENCY (2 decimals for fiat).
+ */
+export function toStorageCurrency(
```
roosemberth: The documentation reads “Convert amount to target currency with storage precision”.

Why is this function not called `convertAmountToCurrencyForStorage`?

What's the relationship with function `createPrice`? Why are there two interpretations of “for storage”?

Conceptually, there's a single projection “to storage” precision; for instance when we store some price (e.g. after order, after configuring a product, ...). Afterwards, we __read__ the price __from__ storage (using some defaulting rules for backwards compatibility), but there's no “conversion” or “creation”.

The current naming makes it hard to understand what function satisfies what responsibility.

src/lib/types/Order.ts
roosemberth: Perhaps it would be easier to read if there was a function `readStoredPrice`? This function would then contain the definition of the defaulting rules within and preserve encapsulation.

src/lib/utils/toStorageCurrency.ts
```diff
@@ -0,0 +1,34 @@
...
+	targetCurrency: Currency,
+	amount: number,
+	fromCurrency: Currency
+): number {
```
roosemberth: It's really unexpected for this function to return a scalar number: The number requires more context to be understood (at least currency). Perhaps it would be best for this to return a `Price` (with currency and precision)?

src/lib/utils/toStorageCurrency.ts
```diff
@@ -0,0 +1,34 @@
...
+	// If currencies are the same, round to storage precision
+	if (fromCurrency === targetCurrency) {
+		return createPrice(amount, targetCurrency, true).amount;
```
roosemberth: We're forgetting information here (ie. precision).

roosemberth (changes requested):

Tirodem:
So far so good, I'll test again in 20 minutes to see if there's variations.

Tirodem:
Tests are OK (and were quite heavy) ✅ 
Sorry @roosemberth , I have to merge this one also, it's required for tomorrow for two pilots.

Issue #2155, opened 2025-10-21
Application CSS is included in pages with “advanced HTML” mode


roosemberth:
Various different elements of the application CSS are included in pages with the “advanced HTML” edition mode. This causes non-trivial style interactions between the styles specified by the “advanced HTML” content and those included in all other be-BOP pages.

There are probably other non-trivial interactions between other `<head>` elements (such as scripts).

As a user, if I write a full page in the “advanced HTML” model, this page should not be affected by implementation details of be-BOP.

Mind that widgets (e.g. `[Product=...]`) should be affected by the be-BOP styles and scripts, but not the rest of the site. We can probably address that in a separate issue though.

PR #2157, opened 2025-10-21
Support for full-HTML CMS pages

roosemberth:
You can now publish full HTML documents using “advanced HTML” CMS pages.

To enable this, the page content **must** begin with the `<!doctype html>` tag. When enabled, these pages are served exactly¹ as authored — without any additional layout, styling, or scripts from be-BOP.

This mode is ideal for:

- Embedding completely custom pages created in your CMS, without mixing in be-BOP’s styles;
- Serving standalone documents that require full control of the HTML structure, such as pages implementing [the Open Graph protocol](https://ogp.me/);
- Integrating externally designed landing pages or microsites.

All other pages continue to render through the standard application layout as before, so your existing content and styles remain unaffected. This also includes CMS pages already using the “advanced HTML” mode that are not full HTML documents.

¹: Because the content is still parsed by Svelte, the served HTML is not byte-for-byte identical to the CMS source, but it is guaranteed to be *semantically* equivalent.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -1,25 +1,9 @@
 <!doctype html>
 <html>
 	<head>
-		<meta charset="utf-8" />
-		<script>
-			const theme = localStorage.getItem('theme') ?? document.documentElement.dataset.theme;
-			if (theme === 'dark') {
-				document.documentElement.classList.add('dark');
-			} else if (theme === 'light') {
-				document.documentElement.classList.remove('dark');
-			} else {
-				if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
-					document.documentElement.classList.add('dark');
-				} else {
-					document.documentElement.classList.remove('dark');
-				}
-			}
-		</script>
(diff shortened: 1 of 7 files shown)


roosemberth:
Closes https://github.com/be-BOP-io-SA/be-BOP/issues/2155.

roosemberth:
Moving to draft since incompatible with the existing infrastructure for be-BOP.io.

We reasses after Lugano.

Tirodem:
Yup 💯 
For memories (a reminder for future testings) : with this kind of deploy, style does not apply on non-advance HTML CMS pages, nor in /admin, and probably /pos
If fixed, it should be tested on :
- nodl-ed be-BOP
- Kiveclair (pm2 with external S3 and mongo)
- old-script dockerized be-BOP
- wizarded be-BOP

roosemberth:
Closing since we're very likely to change the approach.

Issue #2109, opened 2025-10-02
🦋 Manage discount for a particular type of product on PoS only


Tirodem:
Following https://www.passeport-gourmand.ch/neuchatel-jura-jura-bernois

Waiters/waitresses need to be able a discount on a certain type of product (here : food).
We won't go as far as managing automatic discount like those.
But we need to give them a quick way to manage manual discount (in amount mode) by displaying them the total price for a certain type of product, VAT included.

We can base this US on https://github.com/be-BOP-io-SA/be-BOP/issues/1977 for checkout.

What we need might look like this :

<img width="499" height="1635" alt="Image" src="https://github.com/user-attachments/assets/d6366494-d1a1-4106-b59c-46713a7bddda" />

We should obviously do the same with https://github.com/be-BOP-io-SA/be-BOP/issues/2051

But, by doing this that way, waiter/waitress can now how much this or that tag represents in the order, and a 20 or 50% on this, then apply the discount in currency :

<img width="777" height="904" alt="Image" src="https://github.com/user-attachments/assets/2d8ef89f-eefa-4ef3-a5f3-7deb346d0f1b" />

A direct calcul helper would be great, but let's already do that then the PoS touchscreen checkout then we'll collect feedback on this.

Tags displayed here should be tags with the option "available for reporting" for now. Maybe in the future we'll dedicate a specific new option for them.

Note : this option should be PoS account only, obviously.

PR #2158, opened 2025-10-23
 🦋 Add tag-based price breakdown to POS checkout (#2109)

gntw:
Display cart totals grouped by product tags (with VAT) for POS users. Helps
  waiters calculate manual discounts for programs like Passeport Gourmand.

  - Load reportingTags and add tagIds to cart
  - Calculate tag totals
  - Show breakdown above "Payer" button (POS only)


Tirodem:
image
QA OK, @roosemberth for code review please 👌

Issue #2038, opened 2025-09-04
🦋 Allow to generate mid-order tickets to forward orders to bartenders and cooks through product tagging on PoS Touchscreen


Tirodem:
Idea here is, when a waiter/waitress add products to a Pool tab, he/she'll be able to print a ticket to give to bartender or cook to prepare either drinks or food.

On PoS touchscreen :
- add a "Print ticket option"

<img width="137" height="526" alt="Image" src="https://github.com/user-attachments/assets/85679b15-c000-44ec-a1da-8783d33e8ca9" />

On "Print ticket option" :
- Allow to print the whole Pool ticket (including previously ordered products)
  - This'll require to reset the "Pool" object (or at least its content) once it's fully paid otherwise it'll display the products previously ordered by other customers on the same table 3 days ago)
- Allow to print only newly ordered products (and once they're printed, they go from "Pending" to "Acknowledged")
  - this should be the default option

In either case, allow to do it with a filter, to either print drinks, or food, or whatever.

For this, we'll use the existing Tags.
We need to add a new option on `/admin/tags/{id}` :
<img width="328" height="347" alt="Image" src="https://github.com/user-attachments/assets/126a8ff4-c5f7-46e3-b6b8-d1885d1f893d" />

When printing ticket, I can choose :
- "All"
- Tag 1 (eg: drinks)
- Tag 2 (eg: food)
- ...
- Tag N

Output tickets should be something like this :
- Initial ticket on POS touchscreen :
<img width="155" height="694" alt="Image" src="https://github.com/user-attachments/assets/ffe00117-a4cc-4c39-b179-3551ec6e9fe2" />

- Printed ticket for "All" option :
<img width="153" height="321" alt="Image" src="https://github.com/user-attachments/assets/eab6ec35-2399-4829-a603-fb6f5001023f" />

- Printed ticket for "Newly ordered" option :
<img width="155" height="154" alt="Image" src="https://github.com/user-attachments/assets/c2a09e24-9d5d-45e3-8d9d-79f2a9d7e2cd" />

- Printed ticket for "Drinks" option :
<img width="153" height="117" alt="Image" src="https://github.com/user-attachments/assets/7717ed6d-891a-4225-a821-734249b17993" />

- Printed ticket for "Food" option :
<img width="157" height="158" alt="Image" src="https://github.com/user-attachments/assets/67841942-ef6c-4170-bd7c-dd18586b555a" />

[...]

PR #2163, opened 2025-10-27
🦋 Add mid-order ticket printing for PoS touchscreen

gntw:
- Print modes: "All items" and "Newly ordered"
  - Tag-based filtering (drinks, food, etc.)
  - Track printedQuantity for reorder handling
  - Group by productId with variation counts
  - 80mm thermal format without prices

issue #2038

### DIFF
--- a/src/lib/components/PrintTicketModal.svelte
+++ b/src/lib/components/PrintTicketModal.svelte
@@ -1,0 +1,205 @@
+<script lang="ts">
+	import type { Tag } from '$lib/types/Tag';
+	import type { PrintTicketOptions } from '$lib/types/PrintTicketOptions';
+	import type { PrintHistoryEntry } from '$lib/types/PrintHistoryEntry';
+	import { format } from 'date-fns';
+
+	export let isOpen: boolean;
+	export let availableTags: Array<Pick<Tag, '_id' | 'name'>>;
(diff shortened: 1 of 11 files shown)


Tirodem:
I know it's not finished but beware, tags on /pos/touch defined in /admin/pos are replaced with tags available for ticket reporting :
image
image

Tirodem:
As said earlier, mid-order ticket should display products per tag if this kind of setting is enabled.

Also, greenlight for your three propositions 👌

Tirodem:
On tickets :

image

On "print ticket" menu :

image

On ticket, you seem to use "every" tag.
You should only use either tags "BOISSONS" or "NOURRITURE" displayed in the modal on the ticket, not the others.

For reminder, here's the differencing option :
image
image
It's `Use tag for filter printed receipts`

In my config, I only expect to have one ticket part for "BOISSONS" and another ticket part for "NOURRITURE"

Beside that, good progress

Tirodem:
Why some "nourriture" and "boissons" here are not grouped ?
image

Tirodem:
As seen, let's use this last conflict to change CTAs to this target :
image

gntw:
@Tirodem conflicts are done, and CTAs also changed

PR #2183, opened 2025-11-04
Runtime Configuration for SMTP, S3, and Nostr

roosemberth:
SMTP, S3, and Nostr settings now use the runtime configuration system instead of fixed environment variables. Administrators can update these settings directly from the backoffice without redeploying. Environment variables still take precedence, and validation/testing has been added for S3 connections.

The environment variables ADMIN_LOGIN and ADMIN_PASSWORD were removed from the environment file and are no-longer supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 be-BOP is a free and open-source, peer-to-peer monetization platform built for
 communities and creators. It combines e-commerce, point-of-sale (PoS),
 subscriptions, crowdfunding/peerfunding, ticketing, donations, and
-pay-what-you-want models — all under one roof. 
+pay-what-you-want models — all under one roof.
 
 ## Requirements
 


src/routes/.well-known/nostr.json/+server.ts
```diff
@@ -13,6 +13,10 @@ export const OPTIONS = () => {
 };
 
 export const GET = () => {
+	if (!isNostrConfigured()) {
+		throw error(404);
+	}
+	const { publicKeyHex: nostrPublicKeyHex } = getNostrKeys();
```
gntw: should be { pubKeyHex: nostrPublicKeyHex }

roosemberth: Funny the CI didn't catch that.

Applied suggestion.

roosemberth:
@Tirodem for feedback.

Tirodem:
For existing .env.local configuration, we should disable "Remove SMTP settings" or think in another way, because it's enabling fake mode, and after that we cannot set smpt configuration not remove smtp configuration so we're kinda stuck.

image

Note that the config will go back after redeploying be-BOP (kinda logical).

I'm checking with non-setted configuration.

Tirodem:
Fun fact : with this, it seems that S3 settings in .env.local are useless for docker-compose existing installation 👀 

Otherwise, OK for SMTP and Nostr.

I'm testing a last thing before merging, then I'll test again on wizard pilot.

Tirodem:
Non-regression is OK when var are setted, I'm merging for test convenience.

Tirodem:
It's OK for SMTP and Nostr;
I'll test S3 later.

PR #2186, opened 2025-11-08
🦋 Add opt-in telemetry beacon for be-BOP.io statistics

gntw:
Shop owners can now voluntarily share their shop's existence with be-BOP.io for platform statistics and promotion. A banner appears on the admin homepage asking for consent with three options: accept (sends periodic beacons every 6-8 days), decline for 30 days, or decline permanently.

Key features:
- Privacy-first: opt-in only, minimal data (shop URL + version)
- Encrypted transmission via Nostr protocol
- Manageable from SEO settings page
- Countdown animations for better UX
- i18n support (en, fr)

### DIFF
--- a/src/lib/components/TelemetryBanner.svelte
+++ b/src/lib/components/TelemetryBanner.svelte
@@ -1,0 +1,224 @@
+<script lang="ts">
+	import { onDestroy } from 'svelte';
+	import { useI18n } from '$lib/i18n';
+	import { enhance } from '$app/forms';
+	import DEFAULT_LOGO from '$lib/assets/bebop-light.svg';
+
+	export let adminPrefix: string;
+
+	const { t } = useI18n();
+
+	type ViewState = 'initial' | 'accepted' | 'declined' | 'hidden' | 'minimal-accepted';
+
+	let currentView: ViewState = 'initial';
+	let countdownSeconds = 0;
+	let countdownInterval: ReturnType<typeof setInterval> | null = null;
+
+	function updateCountdown() {
+		countdownSeconds--;
+		if (countdownSeconds <= 0) {
+			if (countdownInterval) {
+				clearInterval(countdownInterval);
+			}
+			currentView = currentView === 'accepted' ? 'minimal-accepted' : 'hidden';
+		}
+	}
+
+	onDestroy(() => {
+		if (countdownInterval) {
+			clearInterval(countdownInterval);
+		}
+	});
+</script>
+
+{#if currentView !== 'hidden'}
+	<h1 class="text-xl mb-0 flex items-center gap-2">
+		<span>{t('telemetry.banner.aWordFrom')}</span><img
+			src={DEFAULT_LOGO}
+			alt="be-BOP"
+			class="mb-3 h-10 inline-block align-middle"
+		/><span>:</span>
+	</h1>
+
+	{#if currentView === 'minimal-accepted'}
+		<p class="text-sm mt-2 mb-6">
+			{t('telemetry.banner.accepted.thankYou')}
+		</p>
+	{:else}
+		<div class="p-6 mb-6 border border-gray-200 rounded-lg bg-white shadow-sm">
+			<div class="flex items-start gap-4 mb-4">
+				<div class="flex-shrink-0 w-[80px] ml-3">
+					<svg
+						class="h-16 w-16"
+						viewBox="0 0 100 100"
+						fill="none"
+						xmlns="http://www.w3.org/2000/svg"
+					>
+						<circle cx="50" cy="50" r="8" fill="#31A5DC" />
+						<path
+							d="M 35 35 Q 30 40, 35 45"
+							stroke="#1683f7"
+							stroke-width="3"
+							fill="none"
+							stroke-linecap="round"
+						/>
+						<path
+							d="M 65 35 Q 70 40, 65 45"
+							stroke="#31A5DC"
+							stroke-width="3"
+							fill="none"
+							stroke-linecap="round"
+						/>
+						<path
+							d="M 25 25 Q 18 40, 25 55"
+							stroke="#31A5DC"
+							stroke-width="3"
+							fill="none"
+							stroke-linecap="round"
+						/>
+						<path
+							d="M 75 25 Q 82 40, 75 55"
+							stroke="#31A5DC"
+							stroke-width="3"
+							fill="none"
+							stroke-linecap="round"
+						/>
+						<path
+							d="M 15 15 Q 5 40, 15 65"
+							stroke="#31A5DC"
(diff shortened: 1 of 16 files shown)


src/lib/components/TelemetryBanner.svelte
```diff
@@ -0,0 +1,166 @@
...
+			if (countdown <= 0 && intervalId) {
+				clearInterval(intervalId);
+				intervalId = null;
+				currentView = nextView;
+			}
+		}, 1000);
+	}
+
+	function handleAccept() {
+		dispatch('accept');
```
roosemberth: Why use the dispatch mechanism instead of a simple form? (Like we do in [ProductWidgetPOS](https://github.com/be-BOP-io-SA/be-BOP/blob/1c2e827e90ea7a192fefeb2f17900bf7b636b742/src/lib/components/ProductWidget/ProductWidgetPOS.svelte#L21).)

The actions in this component (accept, decline30, declineForever) or the context surrounding them are for sure not reusable for anything else than telemetry consent.

It seems it would be much easier to read for the form and actionUrls to be hardcoded in this component; then, the parent component would simply instantiate the widget.

src/routes/(app)/admin[[hash=admin_hash]]/seo/+page.svelte
```diff
@@ -17,5 +21,24 @@
...
 	</label>
+
+	<div class="flex flex-col gap-1">
+		<label class="checkbox-label">
+			<input
+				type="checkbox"
+				name="enableBeaconSignal"
+				class="form-checkbox"
+				checked={data.enableBeaconSignal}
+				disabled={data.telemetryState !== 'accepted'}
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/seo/+page.svelte]: If an installation initially refuses telemetry, how do they change their mind and enable telemetry afterwards?

src/lib/components/TelemetryBanner.svelte
```diff
@@ -0,0 +1,224 @@
...
+	onDestroy(() => {
+		if (countdownInterval) {
+			clearInterval(countdownInterval);
+		}
+	});
+</script>
+
+{#if currentView !== 'hidden'}
+	<h1 class="text-xl mb-0 flex items-center gap-2">
+		<span>A word from</span><img
```
roosemberth: Missing translation.

src/lib/components/TelemetryBanner.svelte
```diff
@@ -0,0 +1,224 @@
...
+						<p class="font-semibold">{t('telemetry.banner.question')}</p>
+					</div>
+				</div>
+			</div>
+
+			{#if currentView === 'initial'}
+				<div class="flex gap-4">
+					<form
+						method="post"
+						action="?/telemetry"
```
roosemberth: This should be an absolute path: If we change where the widget is mounted, the action should post to the right endpoint.

(Alternatively, we could turn the handler into a factory and mount the 'telemetry' action on the page, but that feels overkill...)

src/routes/(app)/admin[[hash=admin_hash]]/+page.server.ts
```diff
@@ -38,13 +39,43 @@ export async function load({ url, locals }) {
...
 	} catch (err) {
 		console.error(`Error reading files from docs/${result.lang}`, err);
 		throw error(404, `Unable to load documentation files for "${result.lang}"`);
 	}
 }
+
+export const actions = {
+	telemetry: async ({ request }) => {
+		const formData = await request.formData();
+		const choice = formData.get('choice') as string;
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/+page.server.ts]: Let's parse this as an enum instead and turn the if-else cascade bellow into a switch.

The switch can be made exhaustive with the statement `choice satisfies never` in the default branch.

roosemberth (changes requested):

roosemberth (approved):
Thank you!

I suggest we add the translations for the missing languages (these can be AI-generated) and use `persistConfigElement`, but that's optional and the PR can be merged as-is.

roosemberth:

Tirodem:
Translations are missing outside fr and en.
It leads to crash of the admin welcome page ☢️ 
We should add those translations.
We should also review translation files to check if last commits had a similar issue and fix it.

gntw:
> Translations are missing outside fr and en. It leads to crash of the admin welcome page ☢️ 

i added translations for 5 more langs

Tirodem:
make notifications.nostr.kind 9 instead of 4 (which is for manual message righ now)
please add a comment in code to remind that 4 is manual message, 9 is telemetry message, 0 is automatic sent message (order, and I think so far we don't distinguish rsvp notification, schedule notification, order notification & nostr-bot messages)

Tirodem:
Otherwise it's good for me 👌

src/lib/components/TelemetryBanner.svelte
```diff
@@ -0,0 +1,226 @@
...
+		if (countdownSeconds <= 0) {
+			if (countdownInterval) {
+				clearInterval(countdownInterval);
+			}
+			currentView = currentView === 'accepted' ? 'minimal-accepted' : 'hidden';
+		}
+	}
+
+	onMount(() => {});
+
```
roosemberth: Does nothing, remove.

src/lib/server/sendNotification.ts
```diff
@@ -105,3 +107,31 @@ ${runtimeConfig.brandName} team`;
 		});
 	}
 }
+
+const BEBOP_IO_NOSTR_BOT_NPUB = 'npub16l9pnrkhhagkucjhxvvztz2czv9ex8s5u7yg80ghw9ccjp4j25pqaku4ha';
```
roosemberth: Discussion with @Tirodem on Zulip: We'll use a different identity.

roosemberth:
> make notifications.nostr.kind 9 instead of 4 (which is for manual message righ now)
> please add a comment in code to remind that 4 is manual message, 9 is telemetry message, [...]

@tirodem [Nostr kind 9](https://github.com/nostr-protocol/nips/blob/master/C7.md) is “Chat message”, not Telemetry.

On a side-note, Nostr kind 4 is deprecated, but we're still using it. We should probably update it everywhere in the code.

It the current behavior is okay, feel free to merge. Else, please ask gntw to update the Kind and then merge without any additional review.

Tirodem:
Ok, I've learned about https://nostr.dev/ with kind nomenclature, for me it was an internal reference
I have doubt about historical implementation (cf some messages being 4 and some being 0)
So we need in a future ticket to make a better use of notifications.nostr.kind, but we also need to track a notifications.nostr.use/type for internal usage only, to distinguish manual nostr dm from be-BOP, telemetry dm, orders dm, schedule dm, etc.
Let's use 4 on this, so, and let's address the two topics later.

PR #2193, opened 2025-11-11
Fix 422 error when changing payment method to non-POS types

gntw:
Allow null posSubtype for payment methods without subtypes


Tirodem:
@roosemberth same, sorry, merge with no code review, I don't want to be for too long with a compromised main and the diff seems light & safe

roosemberth:
Change looks good 👍

PR #2205, opened 2025-11-20
🦋 Add external product URL support in CMS widgets

### LINKED ISSUE #1820: 🦋 Allow to use Product widget with an external be-BOP product url
tl;dr : I'm on a be-BOP, I take a product URL from another be-BOP and use is with the Product widget, as `[Product=https://be-bop.url/productslug?display=img-5]`.
That'll display the product on a widget on my CMS page.
Instead of having a Add to cart / Order product CTA, we'll have a "Get to product" CTA (hyperlink, new tab).

gntw:
CMS Product widgets now support external be-BOP product URLs with display variants:

- [Product=https://other-shop.com/product/slug]
- [Product=https://other-shop.com/product/slug display=img-5]
- [Product=https://other-shop.com/product/slug?display=img-5]

External products display with "View Product" button that opens in a new tab instead of "Add to cart". Product data is fetched from the external shop's /.well-known/product/{slug} endpoint.

Closes #1820

### DIFF
--- a/src/lib/components/CmsPage.svelte
+++ b/src/lib/components/CmsPage.svelte
@@ -31,6 +31,12 @@
 		}[];
 	};
 	export let products: CmsProduct[];
+	export let externalProducts: Array<
+		import('$lib/components/ProductWidget/ProductWidgetProduct').ProductWidgetProduct & {
+			externalUrl: string;
+			pictures: import('$lib/types/Picture').Picture[];
+		}
+	> = [];
 	export let pictures: CmsPicture[];
 	export let challenges: CmsChallenge[];
 	export let tokens: CmsTokens;
@@ -67,6 +73,7 @@
 {#if cmsPage.fullScreen}
 	<CmsDesign
 		{products}
+		{externalProducts}
 		{pictures}
 		{challenges}
 		{tokens}
@@ -95,6 +102,7 @@
 		{/if}
 		<CmsDesign
 			{products}
+			{externalProducts}
 			{pictures}
 			{challenges}
 			{tokens}
--- a/src/lib/server/cms.ts
+++ b/src/lib/server/cms.ts
@@ -1,6 +1,8 @@
 import type { Challenge } from '$lib/types/Challenge';
 import type { DigitalFile } from '$lib/types/DigitalFile';
 import type { Product } from '$lib/types/Product';
+import type { Picture } from '$lib/types/Picture';
+import type { Currency } from '$lib/types/Currency';
 import { trimPrefix } from '$lib/utils/trimPrefix';
 import { trimSuffix } from '$lib/utils/trimSuffix';
 import { JSDOM } from 'jsdom';
@@ -17,6 +19,12 @@ import type { Leaderboard } from '$lib/types/Leaderboard';
 import type { ScheduleEventBooked } from '$lib/types/Schedule';
 import { groupBy } from '$lib/utils/group-by';
 import { subMinutes } from 'date-fns';
+import { z } from 'zod';
+import type { ProductWidgetProduct } from '$lib/components/ProductWidget/ProductWidgetProduct';
+export type ExternalProductData = ProductWidgetProduct & {
+	externalUrl: string;
+	pictures: Picture[];
+};
 
 const window = new JSDOM('').window;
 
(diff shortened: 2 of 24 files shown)


Tirodem:
Looks great.

BTW, `[Product=https://other-shop.com/product/slug?display=img-5]` syntax doesn't seem to work (the product is displayed but with img-1 pic)

```
[Product=https://dev-bootik.pvh-labs.ch/product/snack]

[Product=https://dev-bootik.pvh-labs.ch/product/snack display=img-1]

[Product=https://dev-bootik.pvh-labs.ch/product/snack display=img-2]

[Product=https://dev-bootik.pvh-labs.ch/product/snack display=img-3]

[Product=https://dev-bootik.pvh-labs.ch/product/snack display=img-4]

[Product=https://dev-bootik.pvh-labs.ch/product/snack display=img-5]

[Product=https://dev-bootik.pvh-labs.ch/product/snack display=img-6]

[Product=https://dev-bootik.pvh-labs.ch/product/snack?display=img-1]

[Product=https://dev-bootik.pvh-labs.ch/product/snack?display=img-2]

[Product=https://dev-bootik.pvh-labs.ch/product/snack?display=img-3]

[Product=https://dev-bootik.pvh-labs.ch/product/snack?display=img-4]

[Product=https://dev-bootik.pvh-labs.ch/product/snack?display=img-5]

[Product=https://dev-bootik.pvh-labs.ch/product/snack?display=img-6]
```

https://dev-bootik2.pvh-labs.ch/home

src/lib/server/cms.ts
```diff
@@ -29,6 +37,90 @@ purify.addHook('afterSanitizeAttributes', function (node) {
...
+		const urlObj = new URL(url);
+		const pathMatch = urlObj.pathname.match(/\/product\/([^\/]+)/);
+
+		if (!pathMatch) {
+			console.error('Invalid product URL format:', url);
+			return null;
+		}
+
+		const slug = pathMatch[1];
+		const apiUrl = `${urlObj.origin}/.well-known/product/${slug}`;
```
roosemberth: Why is this on `.well-known` and not `/api`?

[RFC 8615](https://datatracker.ietf.org/doc/html/rfc8615) introduces Well-Known URIs as:

> the discovery of information about an origin

Here, we're not doing discovery nor interoperability with other “product-bearing” implementations. Furthermore, the same specifications clearly states:

> Applications that wish to mint new well-known URIs MUST register them, following the procedures in [Section 5.1](https://datatracker.ietf.org/doc/html/rfc8615#section-5.1) [...].

src/lib/server/cms.ts
```diff
@@ -29,6 +37,90 @@ purify.addHook('afterSanitizeAttributes', function (node) {
...
+
+		const response = await fetch(apiUrl, {
+			headers: { Accept: 'application/json' },
+			signal: AbortSignal.timeout(5000) // 5s timeout
+		});
+
+		if (!response.ok) {
+			console.error(`Failed to fetch external product: ${response.status}`);
+			return null;
+		}
```
roosemberth, line 102: Returning `null` will simply ignore the widget exists. It would be more useful to display to the user that a remote product was supposed to be displayed here, but could not be fetch for _reason_.

src/lib/server/cms.ts
```diff
@@ -40,6 +132,12 @@ type TokenObject =
 			display: string | undefined;
 			raw: string;
 	  }
+	| {
+			type: 'externalProductWidget';
```
roosemberth: Why is the `externalProductWidget` not the same as the regular product widget (with a URL that just so happens to be remote)? Why do we care about the fact it is remote?

src/lib/server/cms.ts
```diff
@@ -29,6 +37,90 @@ purify.addHook('afterSanitizeAttributes', function (node) {
 		node.setAttribute('rel', 'noopener');
 	}
 });
+
+// Zod schema for external product data validation
+const externalProductSchema = z.object({
+	_id: z.string(),
```
roosemberth: Why do we care about the `_id`? Shouldn't we simply use the URL of the resource?

src/lib/server/cms.ts
```diff
@@ -29,6 +37,90 @@ purify.addHook('afterSanitizeAttributes', function (node) {
...
+	name: z.string(),
+	shortDescription: z.string(),
+	price: z.object({
+		amount: z.number(),
+		currency: z.string()
+	}),
+	pictures: z.array(z.any()).optional()
```
roosemberth: `z.any()` seems problematic here. It would be best to tighten the type.

gntw:
> BTW, `[Product=https://other-shop.com/product/slug?display=img-5]` syntax doesn't seem to work (the product is displayed but with img-1 pic)
> 

fixed ! @Tirodem

Tirodem:
It's fixed, indeed.
However, I found an issue I didn't think about last time.
If product origin be-BOP has its product like this :
image
The product is properly displayed on the bebop through the [Product=url] widget, and if I click on it, I'm redirected on home as product is not available.
I think we should not expose the product on the .well-known if product is setted to be not displayed on the web ❌

Tirodem:
Works like a charm today ✅ 
@roosemberth you can merge if code review OK

Tirodem:
Note that if you want to display products from a be-BOP B on a be-BOP A, obviously, both need the new version for this feature to work.
We might wait a bit after merge for communication, otherwise it'll lead to support as long as existing be-BOP don't have fresh install / new version.

src/lib/components/CmsPage.svelte
```diff
@@ -31,6 +31,12 @@
 		}[];
 	};
 	export let products: CmsProduct[];
+	export let externalProducts: Array<
+		import('$lib/components/ProductWidget/ProductWidgetProduct').ProductWidgetProduct & {
```
roosemberth, line 35: @gntw Please mind next time to use the standard `import type ...` stanza instead of inline import statements that most standard tools cannot parse.

Issue #2051, opened 2025-09-11
🦋 PoS touchscreen checkout adaptation


Tirodem:
Currenty the PoS touchscreen ( initial issue : https://github.com/be-BOP-io-SA/be-BOP/issues/1291 ) is connected to the standard checkout.

Following what we did with split payment ( https://github.com/be-BOP-io-SA/be-BOP/issues/2051 ), we directly go from /pos/touch pool to /order

After pool discount management ( https://github.com/be-BOP-io-SA/be-BOP/issues/2247 ), we won't need to go to /checkout to manage discount from /pos/touch

The last thing we'll need will be an adaptation of /order page.

Current /order page on mobile for employee :

<img width="341" height="1269" alt="Image" src="https://github.com/user-attachments/assets/1ba3c06a-2f7d-489e-8bfb-8c0a09dbf024" />

<img width="296" height="1269" alt="Image" src="https://github.com/user-attachments/assets/eed98f22-a8a8-4ac3-bfcf-25b7b392ffd6" />

Target /order page on mobile for employee coming from /pos/touch : 

<img width="765" height="1283" alt="Image" src="https://github.com/user-attachments/assets/4f0858e1-0065-4435-b26e-ef9697a44ead" />

That includes :
- remove padding-top and padding-bottom of main class="py-10"
- remove header
- remove footer
- remove "add label" (+) option
- remove "Thanks for your order" message
- remove previous split payment
- on payment, remove ln / onchain invoice link and sum up / paypal / stripe form link
- on payment, remove expiration time
- on payment, remove "Payez pour finaliser la commande"
- change "Annuler" / "Remplacer le mode de paiement" for a better display with col-2
- remove billindg date
- remove contact address
- remove vendor name
- remove order content (products)

PR #2207, opened 2025-11-21
🦋 PoS split payment by items & shares

gntw:
Restaurant POS now supports splitting shared table bills (pools) into partial payments using two flexible modes:

- **"Split by items"**: Select specific products from pool (e.g., "3 burgers, 2 beers") → creates separate order with individual receipt
- **"Split by shares"**: Divide total into N equal parts or custom amounts (e.g., "table of 7, each pays 1/7th") → all payments accumulate on one order
- Choose payment method directly in PoS split interface. Visual selector with icons for cards, cash, etc. Works in both split modes
- Real-time visual feedback as items move between pool and cart
- Payment history tracking shows what's paid and what remains
- Flexible workflow: can start with Split by items and split remaining part with shares

Fixes issue #1968

### DIFF
--- a/src/lib/components/PosPaymentsList.svelte
+++ b/src/lib/components/PosPaymentsList.svelte
@@ -1,0 +1,76 @@
+<script lang="ts">
+	import PriceTag from '$lib/components/PriceTag.svelte';
+	import type { Currency } from '$lib/types/Currency';
+	import { useI18n } from '$lib/i18n';
+
+	const { t } = useI18n();
+
+	export let payments: Array<{
+		number: number;
+		amount: number;
+		currency: Currency;
+		isPaid: boolean;
+		status: string;
+		method: string;
+		createdAt: Date;
+		paidAt?: Date;
+		orderId?: string;
+	}>;
+	export let title: string;
+	export let bgClass = 'bg-gray-100';
+	export let returnTo: string | undefined = undefined;
+
+	function getPaymentLink(payment: (typeof payments)[0]): string | undefined {
+		if (!payment.orderId) {
+			return undefined;
+		}
+		const base = `/order/${payment.orderId}`;
+		return returnTo ? `${base}?returnTo=${encodeURIComponent(returnTo)}` : base;
+	}
+</script>
+
+{#if payments.length > 0}
+	<div class="{bgClass} rounded-lg p-4">
+		<h3 class="font-semibold text-3xl mb-3">{title}</h3>
+		<div class="space-y-2">
+			{#each payments as payment}
+				{@const link = getPaymentLink(payment)}
+				<div class="flex items-center justify-between">
+					<div class="flex-1">
+						<div class="flex items-center gap-3 text-2xl">
+							{#if link}
+								<a href={link} class="font-semibold text-blue-600 underline">
+									Payment {payment.number}:
+								</a>
+							{:else}
+								<span class="font-semibold">Payment {payment.number}:</span>
+							{/if}
+							<span class="font-bold">
+								<PriceTag amount={payment.amount} currency={payment.currency} />
+							</span>
+							{#if payment.isPaid}
+								<span class="text-green-600 text-3xl">✅</span>
+							{:else if payment.status === 'pending'}
+								<span class="text-orange-600 text-3xl">⏳</span>
+							{:else}
+								<span class="text-gray-400 text-3xl">⏸</span>
+							{/if}
+						</div>
+						<div class="text-lg text-gray-600 mt-1">
+							{t(`checkout.paymentMethod.${payment.method}`)} •
+							{new Date(payment.paidAt || payment.createdAt).toLocaleTimeString('en-GB', {
+								hour: '2-digit',
+								minute: '2-digit'
+							})}
+						</div>
+					</div>
+					{#if link}
+						<a href={link} class="bg-blue-600 text-white font-bold px-4 py-2 rounded text-xl">
+							View
+						</a>
+					{/if}
+				</div>
+			{/each}
+		</div>
+	</div>
+{/if}
--- a/src/lib/components/PosSplitItemRow.svelte
+++ b/src/lib/components/PosSplitItemRow.svelte
@@ -1,0 +1,92 @@
+<script lang="ts">
+	import { useI18n } from '$lib/i18n';
+	import PriceTag from '$lib/components/PriceTag.svelte';
+	import type { Currency } from '$lib/types/Currency';
+
+	export let item: {
+		product: {
+			name: string;
+			price: { currency: Currency };
+		};
+		internalNote?: {
+			value: string;
+		};
+	};
+	export let quantity: number;
+	export let priceInfo: { amount: number; currency: Currency };
+	export let vatRate: number;
+
+	// Controls
+	export let showInternalNote = true;
+	export let controls: 'none' | 'move-to-cart' | 'return-to-pool' = 'none';
+	export let isComplete = false;
+
+	// Callbacks
+	export let onMoveOne: (() => void) | undefined = undefined;
+	export let onMoveAll: (() => void) | undefined = undefined;
+	export let onReturnOne: (() => void) | undefined = undefined;
+	export let onReturnAll: (() => void) | undefined = undefined;
+
+	const { t } = useI18n();
+</script>
+
+<div class="py-3 border-b border-gray-300">
+	<div class="flex items-center gap-2 mb-2">
+		{#if controls === 'return-to-pool'}
+			<div class="flex gap-2">
+				<button
+					class="text-orange-600 hover:text-orange-700 font-bold text-3xl leading-none"
+					on:click={onReturnAll}
+				>
+					⏪
+				</button>
+				<button
+					class="border-4 border-orange-600 hover:border-orange-700 text-orange-600 hover:text-orange-700 font-bold text-2xl leading-none px-3 py-1 rounded"
+					on:click={onReturnOne}
+				>
+					➖
+				</button>
+			</div>
+		{/if}
+
+		<div class="flex-1 text-2xl font-bold">
+			{quantity} X {item.product.name.toUpperCase()}
+		</div>
+
+		{#if controls === 'move-to-cart'}
+			<div class="flex gap-2">
+				{#if isComplete}
+					<div class="text-3xl text-green-600">✅</div>
+				{:else}
+					<button
+						class="border-4 border-blue-600 hover:border-blue-700 text-blue-600 hover:text-blue-700 font-bold text-2xl leading-none px-3 py-1 rounded"
+						on:click={onMoveOne}
+					>
+						➕
+					</button>
+					<button
+						class="text-blue-600 hover:text-blue-700 font-bold text-3xl leading-none"
+						on:click={onMoveAll}
+					>
+						⏩
+					</button>
+				{/if}
+			</div>
+		{/if}
+
+		{#if isComplete && controls === 'none'}
+			<div class="text-3xl text-green-600">✅</div>
+		{/if}
+	</div>
+
+	{#if showInternalNote && item.internalNote?.value}
+		<div class="text-xl text-gray-600 mb-2">+{item.internalNote.value}</div>
+	{/if}
+
+	<div class="grid grid-cols-4 w-full items-center justify-around text-xl">
+		<div>{t('pos.split.exclVat')}</div>
+		<PriceTag amount={priceInfo.amount} currency={priceInfo.currency} main />
+		<div>{t('pos.split.vatRate', { rate: vatRate })}</div>
+		<PriceTag amount={(priceInfo.amount * vatRate) / 100} currency={priceInfo.currency} main />
+	</div>
+</div>
--- a/src/lib/server/orderTab.ts
+++ b/src/lib/server/orderTab.ts
@@ -1,9 +1,9 @@
-import { OrderTab } from '$lib/types/OrderTab';
-import { ObjectId } from 'mongodb';
+import { OrderTab, type CheckoutOrderTabParams } from '$lib/types/OrderTab';
+import { ObjectId, type ClientSession } from 'mongodb';
 import { collections } from './database';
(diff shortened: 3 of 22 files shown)


Tirodem:
I have various issues on this one.
I have to retest all of it to confirm the behavior, but from what I saw, there's at least :
- adding products to the pool, then split pay by product (not everything), then adding new products, then paying leads to inconsistencies in global ticket ❌ 
- adding product to the pool, then split pay by product (not everything) then paying everyting doesn't empty the pool ❌

Tirodem:
1st case : error when combining split per product then global payment
- on global payment (leftover), we have cart line at qty 0
- as seen, after payment, the pool is not emptied

Demo's here : https://youtu.be/Y1TjmTjwBpA

Tirodem:
2nd case : spit paying then adding product then paying the rest doesn't generate a good total
- I have 1 product A, 2 products B, 3 products C and 51 products D
- I split pay per product 6 products D (left to pay : 45)
- I add 6 more products D (left to pay : 51)
- I pay everything
- on my ticket, there's a total of 51 products D, it should be 57 (45 + 6 + 6)

https://youtube.com/shorts/7AUSxg-aeIM

Tirodem:
Also :
- split payment per product send to /order directly with a define payment mode, it should send to /order so far, and keep the mobile view
- there's an issue between split payment (divison) and global payment, they have not the same journey, and global payment should be split payment (division per 1)
- missing @@translations

Tirodem:
In order to make a manageable split by part system, we need to associate an icon (emoji) to PoS payment subtype.
Cash will be 💵 per default (but can be changed).
Other generic PoS payment method will be customized by the user.
That way the payment will be selectable on split interface as a CTA (template is coming).

image

src/lib/components/PosPaymentsList.svelte
```diff
@@ -0,0 +1,76 @@
+<script lang="ts">
+	import PriceTag from '$lib/components/PriceTag.svelte';
+	import type { Currency } from '$lib/types/Currency';
+	import { useI18n } from '$lib/i18n';
+
+	const { t } = useI18n();
+
+	export let payments: Array<{
+		number: number;
```
roosemberth: I think paymentId is more useful than a sequential number. This is already a sorted list and we can use the createdAt as as natural sorting key.

src/lib/components/PosSplitItemRow.svelte
```diff
@@ -0,0 +1,93 @@
+<script lang="ts">
+	import { useI18n } from '$lib/i18n';
+	import PriceTag from '$lib/components/PriceTag.svelte';
+	import type { Currency } from '$lib/types/Currency';
+
+	// Props для item
```
roosemberth: Please write comments in English.

src/lib/server/orders.ts
```diff
@@ -2272,6 +2284,11 @@ async function applyOrderSubscriptionsDiscounts(order: Order, session: ClientSes
 }
 
 export async function updateAfterOrderPaid(order: Order, session: ClientSession) {
+	if (order.orderTabSlug) {
+		const { handleOrderTabAfterPayment } = await import('./orderTab');
```
roosemberth: Why is this a dynamic import? Dynamic imports often get in the way of tree-shaking.

src/routes/(app)/order/[id]/+page.svelte
```diff
@@ -886,6 +886,14 @@
 								>{t('order.note.seeText')}</a
 							>
 							{#if data.order.orderTabSlug}
+								{#if data.splitMode}
+									<a
+										href="/pos/touch/tab/{data.order.orderTabSlug}/split?mode={data.splitMode}"
+										class="btn lg:w-auto w-full btn-black self-end"
+									>
+										@@Continue split ({data.splitMode})
```
Comment (roosemberth) [inline src/routes/(app)/order/[id]/+page.svelte]: Untranslated string.

src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts
```diff
@@ -65,12 +82,224 @@ async function getHydratedOrderTab(locale: Locale, tabSlug: string) {
...
+				order = await collections.orders.findOne({ _id: orderId });
+				if (!order) {
+					throw new Error('Failed to create order');
+				}
+			}
+
+			const remainingToPay = orderRemainingToPay(order);
+			const amountToPay =
+				paymentParams.data.mode === 'equal'
+					? remainingToPay / paymentParams.data.shares
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts:283]: What happens if `remainingToPay` is indivisible by the number of shares? Do we round? If so how?

Tirodem:
I have poor performances while wanting to @@continue split + these entries in Network console. In fact, on this deploy, most of the time, this CTA doesn't take me back on split payment, I have to refresh and retry.
image

Tirodem:
If I want to pay a share with Lightning but I'm wrong and I wanted PoS / cash, if I substitue the payment method, I end up with pending / cancelled payment requests that can be confusing for the waiter, we need to improve display :
image
image
It's not so bad, but we need to think of small UI changes.

src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts
Comment (Tirodem) [reply src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts:283]: When we split 6 in fact we do not create 6 invoices. We do 1/6th invoice, then 1/5th invoice, then 1/4th, 1/3rd, 1/2nd and finally leftover invoice. So, between the 1st and the 6th final payment, we might have a 1ct difference, but in the end everything is paid. It does not shock me too much, it's some kind of usual behavior in restaurants when splitting. From what I see, it's rounded to the nearest.

Tirodem:
My pool was 244.44, I paid some, then I added new products.
When I go back to split per part interface :
image
162,96 (already paid) + 81,46 (left to pay) = 244.44 (previous amount), it should be my current total : 286,24).
So, "remaining to pay" should be dynamic and be updated if pool content is changing.
If I update my pool again :
image
If I split by two the remaining, I have 40,74, so half of 81,46 (wrong, non-updated remaning to pay).
If I now try to paid the remaining through "Pay remaning" CTA : same amount, and so, in the end, here's my order status :
image
image
And here's my pool :
image
We have a "0 x CASSOULET" on the right side when we have "6 x CASSOULET" on the left side, but we invoiced only one, in fact (the first before ticket edition) :
image
Still, on the global ticket, we display 6 x CASSOULET :
image
I understand why we have that : when we use split per part, we use the multiple payment on order, and so, as we create the order, the pool is like the cart, and any update on the product content won't be taken in account, as the order cannot be (and should not be, in a more general way of managing orders, for law and accounting compliance) updated.
Still, in a restaurant, I can eat with you and Roos, decide to leave earlier, pay my third of the bill, then you can dessert and coffee, then you split the rest by half : you should pay half of the pool leftover including desserts and coffees, not half of the pool of when I paid & left.
So far, the quickest way of managing that, in my opinion, is to forbid pool edition when we begin to pay part of a pool, then we should allow update later. But if we want to allow pool edition after a first payment, while we might be good with the split per product, we should meet more issues and have to refacto the functional of multiple payment on the same order, or use a new method, to split per part.

Tirodem:
Possibility for the future : invoicing with a generic PWYW product (for ex : 50€), then, once paid, add a "Payment : -50€" product on the pool, to manage the split & total ?

Tirodem:
One weird case (it might be just display) I have to investigate on monday :
image
image
It's confusing to have 1 qty on /pos/touch and 2 on the other interface. Paying one of the two products should not lead to displaying 1 on /pos/touch I think.

Tirodem:
Split per product button is disabled when there's only one product to pay. But it might be a bit confusing to have this display once everything else if paid :
image
image

Especially since the button is enabled when we want to pay an order of only one product :
image

BTW it doesn't seem to be only when there's only 1 product left : I have the impression that I can split product only one time, then I have to split per part instead, which is highly problematic ☢️
here's a product ordered 3 times, and I split per product and paid 1 :
image

@gntw we need to check this monday. It's even demo-killing if we cannot split per product more than one time.

Tirodem:
@@Continue to split payment seems slow especially from an order splitted per part (this morning it seems more reactive from an order splitted per product).

src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts
Comment (roosemberth) [reply src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts:283]: > When we split 6 in fact we do not create 6 invoices. We do 1/6th invoice [..]

This is fine.

What I'm asking is about how we compute the amount (e.g. 1/6th). For instance:

```
> String(10/6)
'1.6666666666666667'
```

This is obviously, non-transferable; thus I wonder how (where?) this is rounded to a transferable amount.

gntw:
@Tirodem,  i made a separate commit to optimize speed. Please check if the lags you observed are gone

Tirodem:
Most issues are fixed.

What we still have :

- we cannot go back from a split per part payment to the pool split payment interface (each time I have to go back manually to /pos/touch) (note that it doesn't concern a backtrack from payment split per product)

- with payment split per part, if I choose to pay 50%, I don't pay (pending payment), then I pay the remaining, my (for example) 90CHF product will have a 45CHF pending payment, a 90CHF done payment, a payment leftover of -45CHF and still the order will be complete

Otherwise, seems OK.

Tirodem:
L'Amuse Bar spotlighted me an issue with the new user journey through /pos/touch : as when paying all-in-one or splitting per part, you go directly on /order instead of /checkout, you won't have this /checkout display
image
For memories those are used to calculate discount on food for Passeport Gourmand.
We need to discuss it tomorrow, we might need to add a way to make discount through the /pos/touch UI, I'll sleep on this and bring a suggestion tomorrow.

Tirodem:
Split per product direct to /order is quite decent 👌 what's missing, in my opinion, is the "print ticket / close pool" interface when you pay the last products and click on "back to split payment".
image
Otherwise : great.
Edit : I'm Christophe from the future and on another test, when I went back, I've had the print ticket / close pool interface 🤔 So I don't know how it happens sometimes, and not some other times.

About split part, if I want to pay half, go back, then the whole, I still have the issue with a negative payment leftover (the solution being to cancel pending payment related to this pool when you do a new split) :
image
However, I think we can to like that, as long as we track the improvement.

Add product, then split & pay, then add other products works 👌 the ticket amount is OK.

Concerning discount on checkout, it'll be covered by #2247 so it's OK for this ticket.

Finally, going back from order/{id} to split payment works like a charm.

@gntw can you have a quick look at the 1st and 2nd point and give your feedback ?

I'm think after that we'll be good & done (still I'll plan some additional QA tomorrow).

src/lib/server/orderTab.ts
```diff
@@ -93,46 +97,94 @@ export async function concludeOrderTab({ slug }: { slug: string }) {
...
+		: orderTab.items.map((line) => ({
+				_id: line._id.toString(),
+				productId: line.productId,
+				quantity: line.quantity,
+				internalNote: line.internalNote,
+				chosenVariations: line.chosenVariations
+		  }));
+
+	// Determine final splitMode
+	const finalSplitMode = itemQuantities ? 'items' : splitMode;
```
roosemberth: It's very counter-intuitive that `splitMode` will be ignored if `itemQuantities` is specified.

Can you improve the function typing to make this obvious?

src/lib/server/orderTab.ts
```diff
@@ -93,46 +97,94 @@ export async function concludeOrderTab({ slug }: { slug: string }) {
...
+		splitMode: 'shares',
+		payments: { $elemMatch: { status: 'paid' } }
+	});
+	if (sharesOrder) {
+		return { success: false, error: 'sharesPaymentStarted' };
+	}
+
+	const poolStartedPayments = orderTab.items.some((i) => i.originalQuantity !== undefined);
+	const existingItem = orderTab.items.find(
+		(i) => !i.cartId && i.productId === params.productId && (!i.orderId || i.quantity === 0)
```
roosemberth: Hard to read.

The adjacent conditions are not trivial. Break this down into a stream, filter the adjacent conditions and then find what you're looking for.

For example:

```
... = orderTab.items
  // Exclude items already in a cart since [...]
  .filter((i) => !i.cartId)
  // Exclude items already in an order unless [...]
  .filter((i) => !i.orderId || i.quantity == 0)
  .find((i) => i.productId === params.productId)
```

src/lib/server/orderTab.ts
gntw: added separate type CheckoutOrderTabParams , now its more obvious

Tirodem:
@gntw considering all the good work we did on split per part and split per product, "Encaisser tout" which brings to /order with no payment selection seem kinda poor in term of experience.

I propose a quickwin : when you click on "Encaisser tout", the user is redirected to /split?mode=shares with "1" part pre-selected.

That'll allow to provide the new payment selector, which is, in my opinion, way better.

We also one last issue : when going from /pos/touch split payment to /order, we switch from mobile view to desktop view. It's really bothersome, but I'm preeeeeeeeeeeeetty sure it can be covered in a later ticket with #2051 . If you fix the first request, you can ignore this second point and we'll be able to merge (we should just not deploy it to Coffice as long as 2051 is not done).

And, so, in 2051, /order/{id} should always be mobile view.

gntw:
> I propose a quickwin : when you click on "Encaisser tout", the user is redirected to /split?mode=shares with "1" part pre-selected.

Done ! 

@Tirodem  i suggest to merge this issue to main, and next we will cover it by 2051

Tirodem:
It seems to work lovely.
My only feedback : once all of the pool is pay, we have this :
image
It seems specific to split per product.
And, so, as the pool is paid as whole, we should not display the payment methods anymore.

gntw:
> It seems specific to split per product.
> And, so, as the pool is paid as whole, we should not display the payment methods anymore.

@Tirodem DONE !

Tirodem:
image
GG @gntw despite this issue making a face-heel turn and becoming an archnemesis in term of complexity and functional in-and-out changes and impacts.

Tirodem:
image

PR #2218, opened 2025-11-27
🦋 Hide WYSIWYG editor for mobile/employee content in advanced HTML mode

### LINKED ISSUE #2194: 🦋 Enable advanced HTML editor for CMS alternative mobile content
self-explaining

Image

if advanced html is used, we should not use WYSIWYG editor for mobile alternative content

gntw:
When "Use advanced HTML edition" is enabled, mobile and employee alternative
content now show only the raw HTML textarea (consistent with main content).

Closes #2194


Tirodem:
Works for me ✅

PR #2222, opened 2025-11-27
🐛 fix: update schedule and booking day logic

### LINKED ISSUE #2221: 🐛 Booking Error due to Day Availability Logic
Image

When trying to book for a Monday while on Thursday, you'll receive the following error message:
Error 400 Message: "Product Book a Slot booking time range is not available on thursday"

Image

Interestingly, when if we make slots available on Thursday, we can book for a Monday without issue. This suggests that the current day’s availability affects the ability to book for future dates.

https://github.com/be-BOP-io-SA/be-BOP/blob/444f5d84dc4c59609e243c5812cabc70c9f4a014/src/lib/server/orders.ts#L1058

It appears that there is a logical flaw in how the daySpec is handled when attempting to book on a future date. If there are no available slots on the current day (Thursday), it should not prevent bookings for another day (Monday).

gntw:
Fixes #2221 - booking slots minor bug fixed

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -1037,9 +1037,9 @@ export async function createOrder(
 						);
 					}
 				}
-				const startDay = startOfDay(startTime.getDay() + 6);
+				const startDay = startOfDay(startTime);
 				const endTime = toZonedTime(time.end, bookingSpec.schedule.timezone);
-				const endDay = endOfDay(subSeconds(endTime, 1).getDay() + 6); // Sub-seconds to allow booking until midnight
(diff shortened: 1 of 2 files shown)


Tirodem:
@roosemberth I'm needing it for new marketplace with home booking. Do you think you can review it before tonight ? If no, no prob, but I'll take it without more review.

roosemberth (approved):
I don't understand why the previous code “worked”. I'm not gonna spend more time trying to understand the replaced code.

@Tirodem If QA is fine, please go ahead and merge.

PR #2252, opened 2025-12-09
🦋 Products can now be grouped by tags in POS touch UI

### LINKED ISSUE #2241: 🦋 Create Group of Tags for multiple levels navigation in /pos/touch
Should be able to create Groups that includes multiple Tags and use those Groups in /pos/touch for quicker navigation. This makes categories of categories of Products that could be use like this:

#### Tags:
- Beers
- Cocktails
- Soft drink
##### Products
- Many
- Different
- Ones
- Because
- You
- Can

#### Tags:
- Pizza
- Meat
- Rice
- Pasta
##### Products
- Many
- Different
- Ones
- Because
- You
- Can

And so on...

gntw:
POS touch interface now supports tag grouping for better product navigation.
Admins can organize tags into named groups (e.g., "Drinks", "Food", "Snacks").

Two display modes:
- Button mode: collapsible groups with expand/collapse
- Dropdown mode: two-level navigation (groups → tags)

Requires migration that converts existing flat tag config into default group.

Closes #2241

### DIFF
--- a/src/lib/components/CategorySelect.svelte
+++ b/src/lib/components/CategorySelect.svelte
@@ -1,39 +1,85 @@
 <script lang="ts">
 	import { useI18n } from '$lib/i18n';
 	import type { Tag } from '$lib/types/Tag';
+	import type { TagGroup } from '$lib/types/TagGroup';
 
 	export let tags: Pick<Tag, '_id' | 'name'>[];
+	export let tagGroups: Pick<TagGroup, '_id' | 'name' | 'tagIds'>[] = [];
 	export let currentFilter: string;
 	export let onSelect: (filterId: string) => void;
 
 	const { t } = useI18n();
 
 	let isOpen = false;
+	let selectedLevel1Idx: number | null = null;
 
 	interface Option {
 		id: string;
 		label: string;
+		type: 'special' | 'group' | 'tag';
 	}
 
-	$: options = [
-		{ id: 'pos-favorite', label: t('pos.touch.favorites') },
-		...tags.map((tag) => ({ id: tag._id, label: tag.name })),
-		{ id: 'all', label: t('pos.touch.allProducts') }
-	] as Option[];
+	$: shouldShowGroups = tagGroups.length >= 2;
 
-	$: currentLabel =
-		options.find((opt) => opt.id === currentFilter)?.label || t('pos.touch.favorites');
+	// Level 1: Groups or flat tags
+	$: level1Options = shouldShowGroups
+		? [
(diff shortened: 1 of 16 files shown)


Tirodem:
image

```
/admin-tirodem:1 Refused to apply style from 'https://dev-bootik2.pvh-labs.ch/style/variables.css?v=179' because its MIME type ('application/json') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
/admin-tirodem:1 Refused to apply style from 'https://dev-bootik2.pvh-labs.ch/style/variables.css?v=179' because its MIME type ('application/json') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
get.c7e1c610.js:1 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'addGroup')
    at get.c7e1c610.js:1:154
    at Array.reduce (<anonymous>)
    at i (get.c7e1c610.js:1:139)
    at Array.u (i18n.4e68db90.js:8:300)
    at bl (105.b244933a.js:14:411)
    at jt (index.ae9021dc.js:4:2972)
    at new El (105.b244933a.js:34:6416)
    at re (scheduler.0f025d56.js:1:8942)
    at Object.p (app.865945bf.js:1:5216)
    at Object.p (app.865945bf.js:1:8523)
pos:1 Refused to apply style from 'https://dev-bootik2.pvh-labs.ch/style/variables.css?v=179' because its MIME type ('application/json') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
pos:1 Unchecked runtime.lastError: A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received
```

I cannot access to / display /admin/pos anymore ❌

Tirodem:
It was also translation.
image
image
image
image

src/lib/components/CategorySelect.svelte
```diff
@@ -1,40 +1,86 @@
...
+	}
+
+	function goBack() {
+		currentGroupId = null;
 	}
 
 	function closeDropdown() {
 		isOpen = false;
+		currentGroupId = null;
 	}
```
roosemberth: There are two collections and two indices.

Why not phrase this in terms of the first and second indices: e.g `level1Idx` and `level2Idx`?
If `level2Idx` is not useful (e.g. because the modal closed), that's fine.

Currently we mix many terms: “options” “groupId”, “back”. Should probably be `l1Idx`, `l2Idx` and `clearSelection`.

What do you think?

Tirodem:
Tests OK following changes ✅

PR #2262, opened 2025-12-13
🐛 Add 30s throttle to nodeless on-chain mempool API calls

### LINKED ISSUE #2261: ☢️ Mempool check on nodeless on-chain payment is done too often
Seem on market.cryptoxr.fr, check is done every 2 seconds : it breaks the mempool API and overloads be-BOP logs.
We should do that way less often.

Tirodem:
Fixes rate limiting issues (429 errors) when using the public mempool.space API by caching results per address for 30 seconds.

Closes #2261

Experimental, not for direct merge

### DIFF
--- a/src/lib/server/bitcoin-nodeless.ts
+++ b/src/lib/server/bitcoin-nodeless.ts
@@ -6,6 +6,23 @@ import { sum } from '$lib/utils/sum';
 import { trimSuffix } from '$lib/utils/trimSuffix';
 import { persistConfigElement } from './utils/persistConfig';
 
+// Cache for throttling mempool API calls (30 seconds per address)
+const MEMPOOL_THROTTLE_MS = 30_000;
+const mempoolCache = new Map<
+	string,
+	{
+		lastFetchedAt: number;
+		result: {
+			satReceived: number;
+			transactions: Array<{
+				currency: 'SAT';
+				amount: number;
+				id: string;
+			}>;
+		};
+	}
+>();
+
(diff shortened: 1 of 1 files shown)


src/lib/server/bitcoin-nodeless.ts
```diff
@@ -145,10 +168,18 @@ export async function getSatoshiReceivedNodeless(
...
 		satReceived: total,
 		transactions
 	};
+
+	// Update cache
+	mempoolCache.set(address, {
+		lastFetchedAt: Date.now(),
+		result
+	});
+
```
roosemberth, line 181: We need either a mechanism for evicting caches (e.g. LRU) or to kickoff something to evict the cache line (e.g. a promise).

Probably a LRU-eviction is the smartest. Else, the cache will continuously grow with record that will never be used after the transaction confirms a payment.

Tirodem: Can it works for a few days before cache grows too much ?

roosemberth: If we leave it as-is, we'll forget about it. In the future, somebody will notice this, mistake it for a memory leak and waste a lot of time debugging code at the wrong place. It makes no sense for a process to allocate ephemeral resources and never release them, thus it's a very unintuivitet issue to nail down.

Fixing it is the more time-efficient. You can probably use the same tool you used to generate the patch to amend the commit and fix it.

The easiest would be to have a 30-entry LRU. 30 is an arbitrary cache size. Any size will do.

PR #2265, opened 2025-12-15
🦋 Add pool-level discount to POS Touch

### LINKED ISSUE #2247: 🆕 Add "Discount" attribute & UI on /pos/touch Pool
We need to be able to make discount on a pool without going to /checkout, as split payment don't go to /checkout

Our first use case if Passeport Gourmand management, which is %discount, and not discount amount, so we'll stick to it.

Product tags exposed here are the same as defined for "Use tag for filter printed receipts".

Only thing here not defined is the modal to fill text justification for discount.

Discount should be carried by the pool, but not used for price display while in the pool. When paying, the discount will be automatically applyed for payment.

Once the order is partially paid, we should disable the discount edition.

When a pool is fully paid / closed / emptied, the discount is also resetted to 0% / no discount.

Image

Image

Image

Image

gntw:
- POS touch pools now support percentage-based discounts (e.g., Passeport Gourmand).
 - Staff can apply discounts via new DISCOUNT button, optionally filtering by tag (Food/Drinks).
 - Discount is locked after first payment and automatically applied at checkout.
 
 Closes #2247

### DIFF
--- a/src/lib/components/OrderSummary.svelte
+++ b/src/lib/components/OrderSummary.svelte
@@ -29,6 +29,7 @@
 				| 'depositPercentage'
 				| 'discountPercentage'
 				| 'freeQuantity'
+				| 'vatRate'
 			> & {
 				digitalFiles: Array<{ _id: string }>;
 				picture?: Picture;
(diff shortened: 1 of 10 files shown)


Tirodem:
It works well !

Only thing : when I apply discount as a PoS manually, here's what I have :
image

When it's done through /pos/touch discount for pool, here's what I have :
image

❌ Discount should be displayed on cart resume on /order/{id}.
Also it should trigger the existing alert for discount :
image

Tirodem:
Also, I think we should disable for now the 100% discount, and forbid a manual discount of 100%, because it should be paid with "free" payment method and we choose a payment method through the interface.
It ends like this :
image
And the order appears as pending but cannot be paid anymore with any payment method.
If we enable full discount, we should automatically set "free" payment method, but for me it not supposed to happen in restaurant and bar

Tirodem:
image
image
We have two things here :
- First, we seems to have a double-discount application. The product is displayed as 14.97, striked, displayed at 7.48 (vat excluded), which is 50%. Then, I have a 4.04 discount, and I have to pay 4.04. Which is 50% again. We should apply discount only once ❌ Maybe we need to remove the strike mechanic in the end if it's sources the dual discount.
- Second, we said "Total already paid : 16.18€, All is paid", which is wrong. All is paid but 16.18€ were not already paid, if was 4.04€. We should display here the real paid price (with discount) ❌

gntw:
> We have two things here

both are fixed ! @Tirodem

Tirodem:
Wunderbar 🎉 
image

The hardest is done ✅ 

We have one last little thing.

Here's a 50% discount via /pos/touch :

image

And here's a legacy 50% discount :

image

❌ The discount displayed here should be from the total VAT included. On /pos/touch we see that we display the discount on total VAT excluded price.

After that, I think we'll be good 👌

Tirodem:
Now we're talking 😃 
Seems all good for me.
@roosemberth for review. Just to let you know, I'll put this on a temporary branch to deploy tomorrow to requesting pilot.

PR #2266, opened 2025-12-15
🦋 Adapt order page for POS touchscreen interface

### LINKED ISSUE #2051: 🦋 PoS touchscreen checkout adaptation

gntw:
Order page now displays a clean, touch-optimized interface for POS users: hides unnecessary details (order labels, creation date, status messages), shows only the active payment, and uses compact button labels in a 2-column grid.

Regular staff and admin users see the unchanged detailed interface with all information visible.

Closes #2051

### DIFF
--- a/src/lib/components/Order/BankTransferPayment.svelte
+++ b/src/lib/components/Order/BankTransferPayment.svelte
@@ -1,0 +1,49 @@
+<script lang="ts">
+	import type { SerializedOrderPayment } from '$lib/types/Order';
+	import type { SellerIdentity } from '$lib/types/SellerIdentity';
+	import { useI18n } from '$lib/i18n';
+
+	const { t } = useI18n();
+
+	export let payment: SerializedOrderPayment;
+	export let sellerIdentity: SellerIdentity | undefined = undefined;
+
+	$: bank = sellerIdentity?.bank;
+</script>
+
+{#if payment.status === 'pending'}
+	<ul class="payment-details-list">
+		{#if bank}
+			<li>
+				<strong>{t('order.paymentAccountHolder')}</strong>
+				<p class="body-secondaryText">{bank.accountHolder}</p>
+			</li>
+
+			{#if bank.accountHolderAddress}
+				<li>
+					<strong>{t('order.paymentAccountHolderAddress')}</strong>
+					<p class="body-secondaryText whitespace-pre-line">{bank.accountHolderAddress}</p>
+				</li>
+			{/if}
+
+			<li>
+				<strong>{t('order.paymentIban')}</strong>
+				<code class="break-all">
+					{bank.iban?.replace(/(.{4})/g, '$1 ')}
+				</code>
+			</li>
+
+			<li>
+				<strong>{t('order.paymentBic')}</strong>
+				<code>{bank.bic}</code>
+			</li>
+		{/if}
+	</ul>
+
+	<!-- Email seller button -->
+	{#if sellerIdentity?.contact.email}
+		<a href="mailto:{sellerIdentity.contact.email}" class="btn btn-black self-start">
+			{t('order.informSeller')}
+		</a>
+	{/if}
+{/if}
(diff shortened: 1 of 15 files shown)


Tirodem:
First access to order :
image

Refresh :
image

So the "true" mobile display is only done after refresh.

image

However, it's fully usable with your updates as it is 👌 

I'll do some more non-regression testing

Tirodem:
What are the conditions to display the new order page ?
From what I see it depend of the employee account, especially the "hasPos" option, whatever the origin is.
But it can't be just like that, as we can have non-bar non-restaurant sellers with a computer selling onsite with hasPos option (to make discount & tva exemption), and they use a desktop / website experience ❌ If we merge the current behavior, it'll impact negatively some of our sellers.
So, for them, we need to maintain classical /order/{id} display (with header and footer).
Also, a restaurant employee coming on an order from /admin/orders or /admin/reporting should have the classical display

What should we have is :
- non-employee : classical /order/{id}
- employee with hasPos option :
  - coming from cart / checkout : classical /order/{id}
  - coming from from direct URL or link to a processed order : classical /order/{id}
  - coming from /pos/touch split interface (supposedly from a mobile device, but not mandatory) : new /order/{id} with now heade / footer

Tirodem:
Now, as a PoS employee making an order from /cart, I don't have header / footer on /order/{id}, but I've also lost them on /checkout 🤔 

image

❌ It should be the opposite, I should keep the classical /checkout AND the classical /order/{id} as I'm not landing from /pos/touch.

Tirodem:
I have to check with the last-last commit, but :

- where's the CTA to print invoice when the payment is done on the simplified view ?

- when I use the classical view, printing ticket leads to this :

image

it's a printing of the current /order/{id}, not the receipt or the ticket 🤔 

I'll have a 🚬 then check again with the last changes

Otherwise the logic to display the classical or simplified /order/{id} seems better, and /checkout is fixed ✅

Tirodem:
Off-github discussion & decision :
- for non-employee, use the "Download invoice" button and lead to /receipt
- for employee, whatever it's from /pos/touch, classical checkout, desktop or mobile :
  - provide a "Print receipt (ticket)" button leading to /ticket
  - provide a "Print receipt (A4)" button leading to /receipt

That way, it'll be simplier and cover all cases.

gntw:
> Off-github discussion & decision :
> 
> * for non-employee, use the "Download invoice" button and lead to /receipt
> * for employee, whatever it's from /pos/touch, classical checkout, desktop or mobile :
>   
>   * provide a "Print receipt (ticket)" button leading to /ticket
>   * provide a "Print receipt (A4)" button leading to /receipt
> 
> That way, it'll be simplier and cover all cases.

@Tirodem , @CypherJack done, lets test it a bit

Tirodem:
That's good 🎉 
@roosemberth for code review

Issue #2236, opened 2025-12-03
🆕 Register the number of people eating at a table on a pool screen in /pos/touch


CypherJack:
Waiter using /pos/touch should be able when he is editing a pool ticket (adding Products to it) he should have a new value that represent people. By default it should be equal to one and being edited at anytime until ticket is done. Something like:🍴1 P

It also should appear on the ticket when the waiter prints all products from a pool.

This is value should be saved somewhere and exported for admin to make his stats like average price per person, per month, etc..

Tirodem:
<img width="747" height="1244" alt="Image" src="https://github.com/user-attachments/assets/b832f936-d45b-429b-bb5b-c46ae872b9f2" />

<img width="728" height="1230" alt="Image" src="https://github.com/user-attachments/assets/5a65e0ea-fb3d-490d-9b42-54d553f23b73" />

CypherJack:
I think the 🧑‍🧑‍🧒‍🧒 emoji shouldn't being used on the selector's option while still being present the selector it self. This bloat the interface and is duplicated. Only number should be used in options.


I am also questioning that particular emoji because both Apple and Google who have the biggest tablet marketshare have replaced family emoji with monochrome androgenic Silhouette to simplify inclusivity [see here](https://emojipedia.org/fr/famille-homme-femme-fille-gar%C3%A7on#designs).

[Pourquoi les émojis famille sont remplacés par des silhouettes neutres sur iPhone ? - numerama.com](https://www.numerama.com/politique/1640170-pourquoi-les-emojis-famille-sont-remplaces-par-des-silhouettes-neutres-sur-iphone.html)

This makes it less recognizable for our use case in my opinion.

What about:
- 👤 [Unique silhouette ](https://emojipedia.org/fr/silhouette-de-buste#designs)
- 👥 [Two simpler silhouettes](https://emojipedia.org/fr/silhouettes-de-bustes#designs)
- 🍴[Fork and Knife](https://emojipedia.org/fr/fourchette-et-couteau#designs)
- 🍽️ [Fork, Plate and Knife](https://emojipedia.org/fr/fourchette-et-couteau-avec-assiette#designs)

PR #2268, opened 2025-12-16
🦋 Add pool-level people count tracker to POS Touch

gntw:
POS staff can now track the number of people at each table/pool directly in the Touch interface. A new dropdown (0-20 people) appears in the bottom-right footer next to the PRINT TICKETS and DISCOUNT buttons, marked with a 👨‍👩‍👧‍👦 icon.

The selected people count:
- Persists when switching between pools
- Automatically saves to the final Order for historical tracking
- Works with both split-by-items and split-by-shares payment modes

This helps staff track table occupancy and provides valuable analytics data in order history (stored as `peopleCountFromPosUi` field).

Fixes issue #2236

### DIFF
--- a/src/lib/translations/de.json
+++ b/src/lib/translations/de.json
@@ -569,7 +569,8 @@
 			"totalInclVat": "Gesamtbetrag inkl. MwSt.",
 			"totalVat": "Gesamt-MwSt.",
 			"unitPrice": "Stückpreis",
-			"vatFreeReason": "MwSt.-befreit: {reason}"
+			"vatFreeReason": "MwSt.-befreit: {reason}",
+			"peopleCount": "Anzahl der Personen"
 		},
 		"receiptFullyPaid": "Quittung erstellen (vollständig bezahlt)",
 		"receiptPending": "Quittung erstellen (ausstehende Zahlung)",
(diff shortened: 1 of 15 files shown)


Tirodem:
❌ Again with translation

❌ "It also should appear on the ticket when the waiter prints all products from a pool." > I didn't find a way to have the information on print as requested.

What we can do is to use the field `orders.receiptNote` on order generation : it'll automatically appear on the receipt like this :
image

Otherwise, good UI / good experience ✅

CypherJack:
If number of people = 0, the information shouldn't be printed at all nor registered anywhere.

Tirodem:
It's good to me ✅ 
@roosemberth for final review
@gntw for merge in deploy-S51-amusebar is not already done

src/lib/translations/de.json
```diff
@@ -617,6 +618,20 @@
 			"resendPaymentMethod": "Zahlungsmethode erneut senden",
 			"updatePaymentInfo": "Zahlungsinformationen aktualisieren"
 		},
+		"discount": {
```
roosemberth: Why are we adding these unrelated translations in this commit?

Please make a separate commit if these changes are necessary (the commit can be included in this PR, no problem there).

(However, I believe these changes have already been merged on main.)

gntw: removed from PR, u r right, its in main already 

done, @Tirodem

PR #2279, opened 2025-12-17
🦋 Fix PWYW product prices to display with VAT when enabled

### LINKED ISSUE #2240: 🦋 If be-BOP display price VAT included PWYW Products price value entered by customer should also be VAT included
Currently when a customer choose a price for a PWYW items on a VAT included be-BOP it doesn't include VAT. It should.

gntw:
- PWYW products now display minimum, recommended, and maximum prices **with VAT included** when the setting is enabled - matching the behavior of regular products.

Closes #2240

### DIFF
--- a/src/lib/utils/vat.ts
+++ b/src/lib/utils/vat.ts
@@ -1,0 +1,54 @@
+import type { CountryAlpha2 } from '$lib/types/Country';
+import { vatRate } from '$lib/types/Country';
+import type { ObjectId } from 'mongodb';
+
+/**
+ * Computes the VAT rate for a product considering custom VAT profiles
+ */
+export function computeVatRate(params: {
+	productVatProfileId: string | ObjectId | undefined;
+	vatProfiles: Array<{
+		_id: string | ObjectId;
+		rates: Partial<Record<CountryAlpha2, number>>;
+	}>;
+	bebopCountry: CountryAlpha2 | undefined;
+	userCountry: CountryAlpha2 | undefined;
+	vatSingleCountry: boolean;
+}): number {
+	const country = params.vatSingleCountry
+		? params.bebopCountry
+		: params.userCountry ?? params.bebopCountry;
+
+	if (!country) {
+		return 0;
+	}
+
+	const vatProfile = params.productVatProfileId
+		? params.vatProfiles.find(
+				(profile) => profile._id.toString() === params.productVatProfileId?.toString()
+		  )
+		: undefined;
+
+	return vatProfile?.rates[country] ?? vatRate(country);
+}
+
+/**
+ * Returns VAT multiplier: (1 + rate/100)
+ */
+export function vatMultiplier(rate: number): number {
+	return 1 + rate / 100;
+}
+
+/**
+ * Applies VAT to an amount
+ */
+export function applyVat(amount: number, rate: number): number {
+	return amount * vatMultiplier(rate);
+}
+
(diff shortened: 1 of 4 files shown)


Tirodem:
If I have to resume the issue, is : when "Display VAT included estimated price in product page" is enable, when I fill the price of a PWYW product, the stocked price should be the filled value minus product VAT

It's about the price chosen by customer, not the minimum / recommended / maximum price chosen by the shop's owner

So, if I set a price of 5€ of a 20% VAT product as a customer :
- if "Display VAT included estimated price in product page" is disabled :
  - VAT excluded price = 5€
  - VAT excluded price = 6€
- if "Display VAT included estimated price in product page" is enabled :
  - VAT excluded price = 4,1667€
  - VAT excluded price = 5€

gntw:
@Tirodem its done ! ready for QA

Tirodem:
Works like a charm 👌 
@roosemberth for review

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -85,21 +89,40 @@
...
-	const PWYWMinimum = toCurrency(
-		PWYWCurrency,
-		data.product.price.amount,
-		data.product.price.currency
-	);
+	const PWYWMinimum = computePriceForDisplay(
+		toCurrency(PWYWCurrency, data.product.price.amount, data.product.price.currency) *
+			vatMultiplier(rate, data.displayVatIncludedInProduct),
+		PWYWCurrency
+	).amount;
```
Comment (roosemberth) [inline src/routes/(app)/product/[id]/+page.svelte]: Why are we binding the amount value instead of keeping the price?
I'm not a fan of keeping amount values without currency (or units) around. It makes computation bugs easy to miss.

The same remark applies to other parts of code.

src/lib/utils/vat.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	return vatProfile?.rates[country] ?? vatRate(country);
+}
+
+/**
+ * Returns VAT multiplier: 1.0 if VAT should not be applied, or (1 + rate/100) if it should
+ */
+export function vatMultiplier(rate: number, shouldApplyVat: boolean): number {
+	return shouldApplyVat ? 1 + rate / 100 : 1;
+}
```
roosemberth, line 40: The `vat` is always defined, regardless of whether we should apply it in a calculation.

I find problematic to have a function with an argument that effectively disables the function. I think it should be the responsibility of the caller to know whether to apply vat or (like it already is, through the argument) and call the correct function accordingly.

If we want to make this code lighter, we can change the semantic of this to something like “price with or without vat to display” and move the logic checking the runtime config here.

The same comment applies to other functions with an argument to bypass the function in this file.

src/routes/(app)/product/[id]/+page.svelte
```diff
@@ -85,21 +89,40 @@
 		});
 	let deposit = 'partial';
 
+	const rate = computeVatRate({
```
Comment (roosemberth) [inline src/routes/(app)/product/[id]/+page.svelte]: `vatRate`.

roosemberth (changes requested):

roosemberth:
@gntw changes look good, thanks!

I'm piggybacking this PR to check automatic generation of artifacts links with https://www.artifact.ci/.

Branch has not been re-QA'd by @Tirodem, but changes look simple enough, so I'm merging anyways. I'll pay him a couple beers if this breaks main.

PR #2291, opened 2026-01-05
🦋 POS Touch: Handle free orders and 100% discounts

### LINKED ISSUE #2288: 🐛 Manage free orders on /pos/touch
While focusing on one of our pilot with /pos/touch, we never considered a possibly free order on bar and restaurant, we even disabled the 100% discount.

But, another of our pilot might need the 100% discount.

Also, this pilot sells free coffee (price = 0), as they need to make a follow-up of this free product stock.

But, currently, if we try to split payment / pay the whole tab if total amount = 0 (pool = only free products), here goes nothing, as we don't cover free payments with the new /pos/touch interface.

We need to :
- enable 100% discount again
- allow "payment" of 0 price pool
  - in that case, when accessing to split payment method selector, if amount = 0, we should provice "Free payment" only
  - we should avoid / check that completely paid orders don't have "Free payment" method on payment method selector
  - we also be aware of this use case : I have 4 paying products on my pool and 3 free products, split per product is done for 4 paying products, what happens for the rest of the pool ?

gntw:
POS touch interface now properly handles free orders:

- 100% discount option added to preset buttons
- Free payment method shown automatically for $0 totals
- Orders with free items only - complete instantly without pending state

Closes #2288

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -2133,9 +2133,13 @@ export async function addOrderPayment(
 	const expiresAt =
 		opts?.expiresAt !== undefined ? opts.expiresAt : paymentMethodExpiration(paymentMethod);
 
+	const isFreePayment = paymentMethod === 'free';
+	const paidAt = isFreePayment ? new Date() : undefined;
+
 	const payment: OrderPayment = {
 		_id: paymentId,
-		status: 'pending',
+		status: isFreePayment ? 'paid' : 'pending',
+		...(paidAt && { paidAt }),
 		method: paymentMethod,
 		price: paymentPrice(paymentMethod, priceToPay),
 		...(paymentMethod === 'point-of-sale' && opts?.posSubtype && { posSubtype: opts.posSubtype }),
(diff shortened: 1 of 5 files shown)


Tirodem:
While testing, I've spotted this :

https://github.com/be-BOP-io-SA/be-BOP/issues/2314

image

By double-triple clicking, I was able to create 3 invoices for the same pool split.

In term of UX, it did this :
- some kind of freeze on split payment interface
- then landing on a 500 error
- then landing on a paid invoice
- then going back to the split, having 1 paid free invoice and 2 pending free invoice

While double-clic protection must be on another ticket, what we need to secure here is the possibility to have a free, pending order.

Free orders are supposed to be automatically created as "paid".

Especially since there's no way to validate them manually as PoS employee :

image

I know it's not supposed to happen anymore once this comment will be fixed but it might be, just in case, a good idea to display the manual validation CTA for a pending free payment. That way, if the case still happen 0.01% of the case, the merchant won't be stuck like this.

gntw:
@Tirodem fixed!

Tirodem:
Ok for me.
@roosemberth for code review please.

Issue #2012, opened 2025-07-25
🐛 When updating nostr relay list on admin, after be-BOP reboot, default relay list is back


Tirodem:
Self-explaining

Tirodem:
It's still present, relays are back after a github action to deploy, I didn't test it again with wizard or docker compose but it must be similar.
My guess is about the seeding / migration script.

PR #2302, opened 2026-01-08
Fix Nostr relay list reset after reboot (#2012)

Tirodem:
Fixes https://github.com/be-BOP-io-SA/be-BOP/issues/2012
  Previously, custom Nostr relay configurations set in Settings > Node Management > Nostr would reset to default values after restarting be-BOP.

  Now relay list changes are properly saved and persist across restarts.


Tirodem:
QA OK ✅

roosemberth:
@Tirodem Can you fix the PR description so that it reads per our usual release notes?
You can take some inspiration from https://github.com/be-BOP-io-SA/be-BOP/releases.

roosemberth:
@Tirodem same in this PR, I really think the previous commit message was much better.

PR #2311, opened 2026-01-09
🐛🤖 Fix documentation encoding issues (#2307)

### LINKED ISSUE #2307: 🐛 / 🦋 Built-in documentation is not encoded correctly resulting in weird characters rendering displayed in raw text
In the Admin Dashboard welcome screen /admin we have at the bottom some WiP documentation that simply open .md files in a new browser tabs. It seems like the encoding isn't correct for rendering in a web browser having weird characters when they are not supported like theses:

Image

Image

---

Also the pages rendered raw text and those not use the MarkDown format for well formatting resulting in intimidating hard to read documentation for people unfamiliar with .md files. Tech-savy users will simply download the file and open it in the software they use.

Image

Tirodem:
Documentation markdown files served from the back-office (`/docs/[lang]/[filename]`) were displaying encoding issues with special characters (é, è, ç, etc.) because the `Content-Type`  header was missing the charset declaration.

Added `charset=utf-8` to the response header so browsers correctly interpret the UTF-8 encoded content.

Partially fixes #2307

### DIFF
--- a/src/routes/(app)/docs/[lang]/[filename]/+server.ts
+++ b/src/routes/(app)/docs/[lang]/[filename]/+server.ts
@@ -34,6 +34,6 @@ export async function GET({ params }) {
 
 	return new Response(content, {
 		status: 200,
-		headers: { 'Content-Type': 'text/plain' }
+		headers: { 'Content-Type': 'text/plain; charset=utf-8' }
 	});
 }


Tirodem:
QA OK ✅ 
image

Issue #2308, opened 2026-01-09
🦋 Improve UI of Login Screen by making "Recovery" button more different and make "Login" more dominant


CypherJack:
Some users reported that they often click on Recovery instead of Login. We could improve the UI to prioritize the most desired action (login) and making the second action less flashy button.

Quick suggestion (not necessarly final UI)

<img width="490" height="366" alt="Image" src="https://github.com/user-attachments/assets/f61d250f-f603-492f-be0c-2096824917d1" />

PR #2312, opened 2026-01-09
🦋🤖 Stack admin login buttons vertically with full width

Tirodem:
Login and Recovery buttons on the admin login page were displayed side by side, which was inconsistent with the form layout.

Changed buttons to stack vertically with full width, matching the login/password input fields for a cleaner, more consistent appearance.

Closes https://github.com/be-BOP-io-SA/be-BOP/issues/2308


Tirodem:
QA OK ✅ 
image

PR #2315, opened 2026-01-12
🦋 Global Ticket printing for thermal printers

### LINKED ISSUE #2280: 🦋 Add a Button to Print the Pool's full bill before splitting or paying on /pos/touch
## Overview
When people finished eating they often ask for the bill before deciding how to split. Waiters should be able to print it before it's paid. The idea is to simply add a Print button on the PAY menu of /pos/touch. Not on the main pool screen where they can print for the cooks and bar.

It should have all the items listed, the discount if any, the total VAT included and the other informations we print on a ticket.

The screen we should add the button (old screenshot)

Image

gntw:
POS Touch UI now supports printing receipts on thermal printers

- **Customer receipts**: Full ticket with prices, VAT breakdown, company logo
- **Ticket logo**: Configurable separately from main site logo in admin

Closes #2280

### DIFF
--- a/src/lib/components/PrintableTicket.svelte
+++ b/src/lib/components/PrintableTicket.svelte
@@ -196,7 +196,6 @@
 	/**
 	 * Wraps text to fit within given width
 	 */
-	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	function wordWrap(text: string, width: number = ticketWidth): string[] {
 		const lines: string[] = [];
 		let currentLine = '';
@@ -244,6 +243,7 @@
 	 * Applies invert formatting to text - white on black (ESC/POS only)
 	 * In plain mode, wraps text with brackets for visibility
 	 */
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	function invert(text: string): string {
 		if (!useEscPos) {
 			return `[${text.trim()}]`;
@@ -289,6 +289,32 @@
 			}))
 		);
 	})();
+
+	// Helper: render variations and notes for an item (shared between global and kitchen tickets)
+	type ItemWithExtras = {
+		variations: Array<{ text: string; count: number }>;
+		notes: string[];
+	};
+
+	const renderVariationsAndNotes = (item: ItemWithExtras, lines: string[], indent = '  ') => {
+		item.variations.forEach((variation) => {
+			lines.push(
+				tableRow([
+					{ text: '', width: 0.15 },
+					{ text: `${indent}${variation.count} ${variation.text}`, width: 0.85, align: 'LEFT' }
+				])
+			);
+		});
+
+		item.notes.forEach((note) => {
+			lines.push(
+				tableRow([
+					{ text: '', width: 0.15 },
+					{ text: `${indent}+${note}`, width: 0.85, align: 'LEFT' }
+				])
+			);
+		});
+	};
 
 	// Ticket Builder - generates formatted ticket content
 	$: ticketContent = (() => {
@@ -339,23 +365,7 @@
 					])
 				);
 
-				item.variations.forEach((variation) => {
-					lines.push(
-						tableRow([
-							{ text: '', width: 0.15 },
-							{ text: `  ${variation.count} ${variation.text}`, width: 0.85, align: 'LEFT' }
-						])
-					);
-				});
-
-				item.notes.forEach((note) => {
-					lines.push(
-						tableRow([
-							{ text: '', width: 0.15 },
-							{ text: `  +${note}`, width: 0.85, align: 'LEFT' }
-						])
-					);
-				});
+				renderVariationsAndNotes(item, lines);
 			};
 
 			let lastGroupKey = '';
@@ -544,7 +554,7 @@
 
 			lines.push(separator('═'));
 		} else {
-			// Kitchen ticket (no prices)
+			// Kitchen ticket (no prices) - unified format with global ticket
 			lines.push('');
 			if (useEscPos) {
 				lines.push(ESC.DOUBLE_HEIGHT + bold(centerText(`TICKET ${poolLabel.toUpperCase()}`, 16)));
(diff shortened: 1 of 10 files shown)


Tirodem:
- add more space at the end of the ticket (because the be-bop.io url is cut when I cut the ticket)
- also replace the ticket from the "ticket" print CTA
image

gntw:
@Tirodem done, can you check it again plz ?

Tirodem:
Good to merge after conflict fixing ✅

src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.svelte
```diff
@@ -711,26 +749,35 @@
...
+			}))
+		}
+	]}
+	priceInfo={{
+		itemPrices: originalQuantitiesPriceInfo.perItem,
+		total: printTotalAfterDiscount,
+		totalExclVat: printTotalExclVat,
+		vatBreakdown: printVatBreakdown,
+		currency: poolCurrency,
+		...(printHasDiscount && printDiscount
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.svelte]: Why do you need `printHasDiscount` if you're using `printDiscount` anyways?

```
printDiscount.percentage > 0
```

src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.svelte
```diff
@@ -271,12 +271,32 @@
...
 		}, 100);
 	}
+
+	$: printTotals = poolTotals ?? computePoolTotals(originalQuantitiesPriceInfo);
+	$: printDiscount = tab.discount;
+	$: printHasDiscount = printDiscount && printDiscount.percentage > 0;
+	$: printTotalAfterDiscount =
+		printHasDiscount && printDiscount
+			? printTotals.incl * (1 - printDiscount.percentage / 100)
+			: printTotals.incl;
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.svelte]: Do you really need `printHasDiscount` if you're using `printDiscount` anyways?

```
$: printTotalAfterDiscount = printTotals.incl * (1 - printDiscount.percentage / 100);
```

src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
```diff
@@ -329,14 +333,32 @@
...
+				const itemIds = itemsWithNotes.map((item) => item.tabItemId);
+				const notesFormData = new FormData();
+				notesFormData.set('itemIds', JSON.stringify(itemIds));
+
+				fetch('?/clearPrintedNotes', {
+					method: 'POST',
+					body: notesFormData
+				}).catch((error) => {
+					console.error('Error clearing printed notes:', error);
+				});
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte]: Why is this a fetch and not a regular form POST?

src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
```diff
@@ -329,14 +333,32 @@
...
+	async function handleReprintFromHistory(entry: PrintHistoryEntry) {
 		ticketTagGroups = entry.tagGroups;
 		ticketGeneratedAt = new Date();
 
+		printModalOpen = false;
+		await tick();
+
 		setTimeout(() => {
 			window.print();
 		}, 100);
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte]: Why do you need this pattern of 1) awaiting tick and 2) deferring a call to `print()`?

It seems we're hiding a problem here.

src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
```diff
@@ -932,7 +954,12 @@
 />
 
 <!-- Printable Ticket (hidden on screen, visible on print) -->
-<PrintableTicket {poolLabel} tagGroups={ticketTagGroups} generatedAt={ticketGeneratedAt} />
+<PrintableTicket
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte]: Why do we need to always keep in in the screen?

Also, if we really must keep it in the DOM, wouldn't it make more sense for us to only display it when printing? (`class="hidden print:block"`) (and also hide the touch UI on print?)

Edit: I now see `PrintableTicket` changes the behaviour of the **page** embedding it. This is extremely unintuitive: When I add a component, I don't expect it to mess with the already existing components.

src/lib/components/PrintableTicket.svelte
```diff
@@ -15,107 +19,764 @@
...
-				margin: 0;
-				padding: 0;
+				margin: 0 !important;
+				padding: 0 !important;
 			}
 
+			/* Hide everything except .printable */
 			body * {
 				visibility: hidden;
 			}
```
roosemberth: I think this behaviour is problematic: It's very unexpected that by adding a component called `PrintableTicket` for the rest of the UI to disappear and be replaced by the contents of this components (regardless of whether that's during “printing”).

The current behaviour is problematic because:

1) When we “print” a **page** (e.g. file > print), we make-believe that the user's intent is not to print the page they're currently viewing, but to print a ticket instead;

2) To trigger this behavior, we add a **child component** to the page that overrides the page view.

I suggest:

- Regarding (1): Let's avoid such a convoluted behaviour and instead use the same [mechanism we use to render](https://github.com/be-BOP-io-SA/be-BOP/blob/3fc488fe751c9047c5a73e44f65923036a1d690a/src/routes/(app)/order/%5Bid%5D/%2Bpage.svelte#L201-L207) and [print](https://github.com/be-BOP-io-SA/be-BOP/blob/3fc488fe751c9047c5a73e44f65923036a1d690a/src/routes/(app)/order/%5Bid%5D/%2Bpage.svelte#L192) order receipts;
- Regarding (2): This is textbook CSS leakage; code with this pattern composes poorly and yields unpredictable results. Let's not style code outside this component; instead, let's [break out of svelte's layout](https://svelte.dev/tutorial/kit/breaking-out-of-layouts) - this way we don't have to fight against the framework and we can reuse the “ticket layour” across different tickets we may want to print (intermediary, order, ...).

src/routes/(app)/pos/touch/tab/[orderTabSlug]/kitchen-ticket/+page.server.ts
```diff
@@ -0,0 +1,72 @@
...
+	const tab = await getOrCreateOrderTab({ slug: params.orderTabSlug });
+
+	const poolLabel = params.orderTabSlug
+		.split('-')
+		.map((word: string) => word.charAt(0).toUpperCase() + word.slice(1))
+		.join(' ');
+
+	const mode = (url.searchParams.get('mode') ?? 'all') as 'all' | 'newlyOrdered';
+	const tagFilter = url.searchParams.get('tagFilter');
+	const historyIndex = url.searchParams.get('historyIndex');
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/kitchen-ticket/+page.server.ts:14]: You don't need `url`, you can get these from `params` already :)

Comment (gntw) [reply src/routes/(app)/pos/touch/tab/[orderTabSlug]/kitchen-ticket/+page.server.ts:14]: In SvelteKit, params only contains route parameters from the path (e.g., orderTabSlug), while query parameters like ?mode=... are only accessible via url.searchParams. Unless I'm missing something? https://svelte.dev/docs/kit/load

Comment (roosemberth) [reply src/routes/(app)/pos/touch/tab/[orderTabSlug]/kitchen-ticket/+page.server.ts:14]: Oh, you're right. My bad.

roosemberth:
@gntw Can you please fix the conflicts? Then ask Tirodem for a quick review to make sure all is good; afterwards Tirodem can merge.

Tirodem:
image

PR #2317, opened 2026-01-12
🐛🤖 Fix product line comment persistence on mid-ticket print (#2289)

### LINKED ISSUE #2289: 🐛 Issue with product line comment on /pos/touch when printing mid-ticket
Let's say I order 3 steaks on /pos/touch
I want a medium, a rare and a well cooked.
I put the comment :

Image

Image

I print the ticket :

Image

I order a 4th burger, standard, no special cooking required.

I print the ticket :

Image

Comment is persistent, which is understandable, it's legacy behavior and we forgot to cover it with the mid-ticket introduction.

tl;dr when I print mid-ticket, the system should reset the product line comments.

Tirodem:
When printing a mid-ticket on /pos/touch, product comments (e.g., "medium", "rare", "well-cooked") should be cleared after printing. This allows staff to add new comments for subsequent 
orders without old comments reappearing.

  Comments were persisting after printing. For example:
  1. Order 3 steaks with cooking instructions → print ticket
  2. Add a burger → print ticket
  3. The old steak comments would still be visible in the system

  Additionally, if user clicked on a product line to change quantity (without changing the comment), the stale cached comment would be re-saved.

  - Server-side: Comments are cleared when mid-ticket print status is updated
  - Client-side: Comment cache is flushed immediately after printing to prevent stale data 

  Closes #2289

### DIFF
--- a/src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
+++ b/src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
@@ -313,6 +313,16 @@
 				}).catch((error) => {
 					console.error('Error updating print status:', error);
 				});
+
+				// Clear notes locally for printed items to avoid stale cache
+				for (const item of filteredItems) {
+					const idx = items.findIndex((i) => i.tabItemId === item.tabItemId);
+					if (idx !== -1) {
+						items[idx].internalNote = undefined;
+					}
+				}
+				// Force the reactive system (Svelte) to update all 'items' dependencies.
+				items = items;
 			}
 
 			invalidate(UrlDependency.orderTab(tabSlug));


Tirodem:
QA ok after 2nd commit✅ 
Initial commit worked but after printing, if I updated the quantity of a product instead of adding a new one, the comment was sorta cached, so it was put back in the qty / comment form, so it was saved, so it was ready to be printed again...
It's good after last commit.

src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte
```diff
@@ -313,6 +313,15 @@
...
 				});
+
+				// Clear notes locally for printed items to avoid stale cache
+				for (const item of filteredItems) {
+					const idx = items.findIndex((i) => i.tabItemId === item.tabItemId);
+					if (idx !== -1) {
+						items[idx].internalNote = undefined;
+					}
+				}
+				items = items;
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte:325]: Remove, this does nothing.

Comment (roosemberth) [reply src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte:325]: Offline discussion: This abuses the reactive system to force a new frame downstream. This seems to be intended, so please add a comment.

Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/+page.svelte:325]:
```suggestion
				// Force the reactive system (Svelte) to update all 'items' dependencies.
				items = items;
```

PR #2319, opened 2026-01-13
🦋🤖Remove clear session trashbin icon from login page

### LINKED ISSUE #1848: 🦋 Remove the recycle bin icon next to Npub on /login
# Overall
When a user request a session login from nostr using !session and confirmed it, the /login screen show their Npub with a little red recycle bin icon next to it. It has the same behavior than the Disconnect button.

It's a bit confusing for the user and might imply that it will delete their order history or their "account" we should remove it in favor of Disconnect button.

Image

Tirodem:
Fixes #1848


Tirodem:
QA OK ✅

Issue #2270, opened 2025-12-16
🐛 Checkout discount limit issue with fiat unit discounts


CypherJack:
## Overview
When creating a checkout discount with a POS employee account, if the discount amount is greater than 99 fiat units, an "invalid discount" error message appears, and the order cannot be processed.

The system is applying a limit of 99 fiat units for discounts, similar to percentage-based discounts.

Tirodem:
Weird values but still present :

<img width="1516" height="826" alt="Image" src="https://github.com/user-attachments/assets/08891d6d-a027-48c1-8772-9b42bc3c15a5" />

<img width="1531" height="1118" alt="Image" src="https://github.com/user-attachments/assets/ffe7bff1-c2fb-4073-b417-7359f6efd839" />

PR #2328, opened 2026-01-21
🐛🤖 Fix discount validation to allow up to 100% discounts

Tirodem:
Fixes https://github.com/be-BOP-io-SA/be-BOP/issues/2270

  ## Summary
  - Employees with PoS options can now apply 100% percentage discounts (was limited to 
99%)
  - "Offrir la commande" checkbox correctly sets 100% discount
  - Fiat discounts equal to the full order amount now work correctly (e.g., 215€ discount on a 215€ order)

  ## What changed
  When logged as an employee with PoS options on `/checkout`, the discount validation now accepts:
  - Percentage discounts up to 100% (previously limited to 99%)
  - Fiat discounts equal to the exact order total (previously rejected with "La réduction n'est pas valable !")

### DIFF
--- a/src/routes/(app)/checkout/+page.svelte
+++ b/src/routes/(app)/checkout/+page.svelte
@@ -115,16 +115,7 @@
 	);
 	$: deliveryFeesToBill = offerDeliveryFees ? 0 : orderDeliveryFees;
 
-	$: isFreeOfCharge =
-		addDiscount &&
-		((discountType === 'percentage' && discountAmount === 100) ||
-			(discountType === 'fiat' &&
-				discountAmount >=
-					toCurrency(
-						data.currencies.main,
-						priceInfoWithoutDiscount.totalPriceWithVat,
-						UNDERLYING_CURRENCY
-					)));
+	$: isFreeOfCharge = addDiscount && discountType === 'percentage' && discountAmount === 100;
 
 	$: possiblyOutOfBoundsDiscount =
 		addDiscount && discountType
@@ -217,7 +208,9 @@
 		return [...tagEntries, ...otherEntry];
 	})();
 
-	$: nothingToPay = priceInfo.totalPriceWithVat === 0 || isFreeOfCharge;
+	$: nothingToPay =
+		isFreeOfCharge ||
+		toCurrency(data.currencies.main, priceInfo.totalPriceWithVat, priceInfo.currency) <= 0;
 	$: paymentMethods = nothingToPay
 		? ['free']
 		: data.paymentMethods.filter(
@@ -231,12 +224,14 @@
 	$: isDiscountValid =
 		isFreeOfCharge ||
 		(discountType === 'fiat' &&
-			discountAmount <=
+			(discountAmount <=
 				toCurrency(
 					data.currencies.main,
 					priceInfoWithoutDiscount.totalPriceWithVat,
 					UNDERLYING_CURRENCY
-				)) ||
+				) ||
+				toCurrency(UNDERLYING_CURRENCY, discountAmount, data.currencies.main) <=
+					priceInfoWithoutDiscount.totalPriceWithVat)) ||
 		(discountType === 'percentage' && discountAmount <= 100);
 	let showBillingInfo = false;
 	let isProfessionalOrder = false;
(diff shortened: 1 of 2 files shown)


Tirodem:
Tests OK for both % and fiat amount discount ✅ 
@gntw I don't really feel at ease with this fix as :
- I expected a fiat amount only fix but while investigating there also was a % bug to fix following a regression of a PoS discount ticket on classical /checkout
- the fiat amount bug to fix was related to a circular reference and by experience I have some PTSD about this

Can you take a bit of time to deeply review the original bug, the fix / changed code, and tell me if it's okay with you or if you rather fix it yourself from scratch ?

src/routes/(app)/checkout/+page.svelte
```diff
@@ -206,10 +218,16 @@
...
 		(discountType === 'fiat' &&
-			priceInfo.totalPriceWithVat >= toSatoshis(discountAmount || 0, data.currencies.main)) ||
-		(discountType === 'percentage' && discountAmount <= 99);
+			discountAmount <=
+				toCurrency(
+					data.currencies.main,
+					priceInfoWithoutDiscount.totalPriceWithVat,
+					UNDERLYING_CURRENCY
+				)) ||
+		(discountType === 'percentage' && discountAmount <= 100);
```
roosemberth, line 235: Instead of adding a workaround to the circular dependency why not break the circular dependency alltogether?

```
$: isFreeOfCharge = discountType === 'percentage' && discountAmount === 100;
$: nothingToPay = priceInfo.totalPriceWithVat === 0 || isFreeOfCharge;
$: paymentMethods = nothingToPay ? ...
```

In svelte 4, the reactive system is more applicative, so we should also include a short-circuit in `isDiscountValid` for the case (nothingToPay, free).

(For completeness: Svelte 5 is monadic with runtime tracking, so this instance of the problem doesn't exist.)

Tirodem: Yup, I'm OK with your feedback, that's why I proposed to Vadim to continue / redo this ticket.

gntw: DONE

Tirodem:
image

which is ok

but after processing :

image

also :

image

input number here should have 100 maximum

note that 100% discount but filled with Fiat leads to error, but 100% with % is OK and processed as free payment.

Tirodem:
@gntw I did change for 2 things following tests :
- isFreeOfCharge now detects fiat discount >= total price, not just 100% percentage  
- Added max=100 constraint on discount input when type is percentage

I gonna QA it right now.

Tirodem:
Shit :
image
image

gntw:
@Tirodem I reworked your commit 4a90ee48 to follow @roosemberth feedback about not expanding priceInfoWithoutDiscount usage.

What changed:

1. isFreeOfCharge reverted to percentage-only check. The fiat case doesn't need it — see below.

2. nothingToPay now checks toCurrency(mainCurrency, totalPriceWithVat, currency) <= 0 instead of totalPriceWithVat === 0. This naturally catches fiat discounts equal to total price: after applying the discount, a
few SAT may remain due to EUR↔SAT rounding, but they convert to €0.00 — so nothing to pay.

3. Server-side fix in orders.ts: same logic — toCurrency(mainCurrency, totalSatoshis, 'SAT') > 0 instead of totalSatoshis !== 0. This fixes the "You can't use free payment method" error you found (5 SAT remainder
= €0.00 = free).

4. max on input: kept max=100 for percentage only. For fiat, isDiscountValid already handles the validation — no need to duplicate it via HTML max attribute.

Why not use priceInfoWithoutDiscount in isFreeOfCharge: it spreads the workaround into more reactive chains.

Tirodem:
image
image
image
image

To be honest, functionally, it's nearly acceptable as it's better than it was before, but I'm triggered about this limit : how comes it works with 1112,29 and not 1112,30 ?

version.txt = `5306b2def76e499bac1ead6b213c7229e29cba8f`

gntw:
@Tirodem , i fixed last mentioned problem

Tirodem:
image
image
image
image

PR #2330, opened 2026-01-21
🐛🤖 Fix OAuth provider deletion redirecting to 404 error page

### LINKED ISSUE #2044: 🐛 Wrong fallback after deleting Oauth provider
When you delete an Oauth povide in /admin/oauth :

Image

With landing URL being /admin/oauth/{id}/?delete .
We should land on /admin/oauth without error.

Tirodem:
Fixes #2044

  ## Summary
  - Deleting an OAuth provider now redirects to `/admin/oauth` with no errors
  - Fixed 404 error that appeared after deletion

  ## What changed
  In the admin OAuth settings (`/admin/oauth`), when deleting an OAuth provider, the page now properly redirects back to the OAuth providers list instead of showing "Error 404, OAuth provider not found: {label}".

  Previously, the delete action had no redirect, causing the page to reload at the now-invalid URL `/admin/oauth/{id}/?delete` where the provider no longer exists.

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/oauth/[id]/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/oauth/[id]/+page.server.ts
@@ -1,6 +1,7 @@
 import { collections } from '$lib/server/database.js';
 import { runtimeConfig } from '$lib/server/runtime-config';
-import { error } from 'console';
+import { error, redirect } from '@sveltejs/kit';
+import { adminPrefix } from '$lib/server/admin';
 import { z } from 'zod';
 
 export const actions = {
@@ -71,5 +72,7 @@ export const actions = {
 				}
 			}
 		);
+
+		throw redirect(303, `${adminPrefix()}/oauth`);
 	}
 };


Tirodem:
Tests OK ✅

src/routes/(app)/admin[[hash=admin_hash]]/oauth/[id]/+page.server.ts
```diff
@@ -71,5 +72,7 @@ export const actions = {
 				}
 			}
 		);
+
+		throw redirect(303, `${adminPrefix()}/oauth`);
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/oauth/[id]/+page.server.ts:76]: Why not also redirect when the OAuth provider is updated?

Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/oauth/[id]/+page.server.ts:76]: It's an UX thing. When non-power users set things like OAuth provider, payment or anything else with api keys or similars, they tend to fill it, then keep the page to check with the source, then test, then keep the page handy if any change is required. Directly redirecting oauth update to main page can cause confusion and anxiety to user.

PR #2332, opened 2026-01-21
🦋 Add POS cashback calculator with Z-ticket integration

### LINKED ISSUE #2286: 🆕 on /pos/touch, manage real payment vs theoric payment + change amount
# Needs

- People aren't always good in math or focused enough during service to do quick calculation : they need some kind of helper to know how much change they have to give back to customer when paying with cash

- Retail cashback leads sometimes customer to pay more with credit card or twint and have cash back ; we need to make something to manage this cash output while doing the PoS closure / Z Ticket.

- (in the future) we'll need a way to auto-manage tips, like planned in https://github.com/be-BOP-io-SA/be-BOP/issues/1873

# Suggestion

## Calculation helper

Here's the current /order/{id} from /pos/touch :

Image

What we'll need here is something like that :

Image

⚠️ Cashback will be managed for Point of Sale payment method only for now ⚠️ 
To do that with Bitcoin on-chain and Lightning natively from be-BOP, we'll need to adapt some stuff later.

Also note that the "Valider" button is not required (you validate thorugh the "Payé" button).
And the "Cashback" can be deployed by default (we'll spare a click).

Image

When depoyed, "Payé" is empty, so "A rendre" is equal to "A payer" ; "A payer" is dynamically udated when changing "Payé" input. "Cancel cashback" will empty the "Payé" input.

If "Payé" < "A payer" :

Image
- highlight input in error
- disable "Payé" CTA

## Cashback manager

When a pending order is "Payé" :
- if "Payé" is not filled  then "Payé" = "A payer", and no cashback is saved
- if "Payé" is filled and "Payé" = "A payer" then no cashback is saved
- if "Payé" is filled and "Payé" > "A payer" :
  - if payment method is "PoS Cash" then no cashback is saved
  - if payment method is NOT "PoS cash" then save "A rendre" amount as cash outcome while closing PoS session / doing Z ticket

Here's how I imagine things :

Image

⚠️ Note that, as it doesn't appear on my test screenshot, here we should have a non-cash amount on "Daily incomes" : this should be the REAL amount paid, and not the expected to pay amount ⚠️

gntw:
🦋 POS cashback calculator with Z-ticket integration

POS cashiers can now calculate change directly on the payment screen. A new "Cashback" button appears next to "Change", opening a calculator section with:
- Amount to pay (from order)
- Amount paid by customer (editable)
- Change to return (auto-calculated, red if insufficient)

For non-cash payments (card, Twint, crypto), the cashback amount is automatically tracked and appears in the Z-ticket as "Cashback to customer" under Daily outcomes. This helps reconcile the cash drawer when customers pay by card but receive cash back.

Daily incomes now shows the actual amount received (order total + cashback), not just the order amount, giving accurate revenue reporting.

Closes #2286

### DIFF
--- a/src/lib/components/Order/PaymentActions.svelte
+++ b/src/lib/components/Order/PaymentActions.svelte
@@ -23,9 +23,17 @@
 	export let ticketReady = false;
 
 	let openPaymentMethodChange = false;
+	let openCashbackSection = false;
 	let disableInfoChange = true;
+	let paidAmount: number | null = null;
 
 	$: tapToPayInProgress = payment.posTapToPay && payment.posTapToPay.expiresAt > new Date();
+	$: toPay = payment.currencySnapshot.main.price.amount;
+	$: changeAmount = Math.max(0, (paidAmount ?? 0) - toPay);
+	$: isInsufficientPayment = paidAmount !== null && paidAmount < toPay;
+	$: isPosNonCash = payment.method === 'point-of-sale' && payment.posSubtype !== 'cash';
+	$: showCashbackButton =
+		posMode && payment.method === 'point-of-sale' && payment.status === 'pending';
 	$: posSubtype =
 		payment.method === 'point-of-sale' && payment.posSubtype
 			? posSubtypes?.find((s) => s.slug === payment.posSubtype)
@@ -180,6 +188,22 @@
 				{/if}
 			</form>
 
+			<!-- Hidden cashback input for form submission (non-cash only, positive change) -->
+			{#if showCashbackButton && isPosNonCash && changeAmount > 0}
+				<input
+					type="hidden"
+					name="cashbackAmount"
+					value={changeAmount}
+					form="confirmForm-{payment.id}"
+				/>
+				<input
+					type="hidden"
+					name="cashbackCurrency"
+					value={payment.price.currency}
+					form="confirmForm-{payment.id}"
+				/>
+			{/if}
+
 			<!-- Unified button container (POS mode = grid, normal = flex) -->
 			<div class="flex flex-wrap gap-2 unified-buttons">
 				<!-- Cancel button -->
@@ -194,7 +218,12 @@
 
 				<!-- Mark paid button (POS + bank-transfer only) -->
 				{#if payment.method === 'point-of-sale' || payment.method === 'bank-transfer'}
-					<button type="submit" class="btn btn-black" form="confirmForm-{payment.id}">
+					<button
+						type="submit"
+						class="btn btn-black"
+						form="confirmForm-{payment.id}"
+						disabled={isInsufficientPayment}
+					>
 						{t(posMode ? 'pos.btn.paid' : 'pos.cta.markOrderPaid')}
 					</button>
 				{/if}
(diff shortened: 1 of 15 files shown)


src/lib/components/Order/PaymentActions.svelte
```diff
@@ -23,9 +23,17 @@
 	export let ticketReady = false;
 
 	let openPaymentMethodChange = false;
+	let openCashbackSection = false;
 	let disableInfoChange = true;
+	let paidAmount: number | null = null;
 
 	$: tapToPayInProgress = payment.posTapToPay && payment.posTapToPay.expiresAt > new Date();
+	$: toPay = payment.currencySnapshot.main.price.amount;
+	$: changeAmount = (paidAmount ?? 0) - toPay;
```
roosemberth: Shouldn't changeAmount be zero if customer hasn't paid enough?

src/lib/translations/es-sv.json
```diff
@@ -628,6 +628,14 @@
 			"replaceSubmit": "Cambiar",
 			"tapToPay": "Pago con tarjeta"
 		},
+		"cashback": {
+			"title": "Cambio ▾",
```
roosemberth: Cambio de efectivo.

src/lib/translations/es-sv.json
```diff
@@ -628,6 +628,14 @@
...
+			"toPay": "A pagar",
+			"paid": "Pagado",
+			"change": "Cambio",
```
roosemberth: Vuelto

src/lib/server/pos-sessions.ts
```diff
@@ -123,6 +127,26 @@ export async function calculateDailyIncomes(session: PosSession): Promise<PosSes
 	return Array.from(incomes.values());
 }
 
+export async function calculateTotalCashback(
+	session: PosSession
+): Promise<{ amount: number; currency: Currency }> {
+	const orders = await collections.orders
+		.find({
+			createdAt: { $gte: session.openedAt },
+			status: 'paid'
```
roosemberth: Nitpick: I would only fetch those with cashback.

Tirodem:
I opened a pool on a new session.
I've made those payments :
- https://dev-bootik2.pvh-labs.ch/order/f82a113d-e47d-46a7-a30e-ea0286260fab rounded up to 35
- http://dev-bootik2.pvh-labs.ch/order/85f70442-7a01-4e3f-9a5f-cd24d5e0c14b rounded up to 10
- https://dev-bootik2.pvh-labs.ch/order/5e611604-e0a0-44e1-a452-e6e079772a1b rounded up to 670
- https://dev-bootik2.pvh-labs.ch/order/46a399dc-8600-4b7f-affe-7bf9ca95931f rounded up to 35

Here's what I have while closing the PoS session :

image

As you can see, we seem to only take the last cashback, while it supposed to increment for each payment during the session.

I'll test one more time to verify my theory but I think we overwrite the cashback amount for the Z ticket instead of incrementing it ❌

Tirodem:
Oh wait I'm dumb it's possible all those except one were made by cash so it's logical we don't track them x) 🙏

Tirodem:
So, my bad, it works like a charm ✅ 

image

image

image

image

image

image

image

Tirodem:
image

Tirodem:
@gntw there's one thing left , if cashback > current cash in cash drawer. We can end up with negative cash at session closure. We won't process that in this ticket, we'll dig up with @CypherJack following pilot's feedback, we'll probably add a way to add more cash in drawer during the day (opposite of bank deposit) on session closure.

Otherwise, it's good 👌 

@roosemberth if vadim's change are OK for you, for me tests are ✅ , it can be merged directly.

roosemberth (approved):

Issue #1793, opened 2025-04-07
🦋 Schedule widget : error management on form submission should expand the events with error on their form if possible


Tirodem:
Following https://github.com/be-BOP-io-SA/be-BOP/pull/1792#issuecomment-2782965219

PR #2333, opened 2026-01-21
🐛🤖 Auto-expand schedule events with validation errors

Tirodem:
Fixes issue #1793

  ## Summary
  - Schedule events with validation errors now automatically expand
  - Users can immediately see and fix errors without manually expanding each event     

  ## What changed
  In the admin schedule editor (`/admin/schedule/{id}`), when a form validation error occurs on a specific event, that event's details block now automatically opens to display the error message.

ually expand each event to find the error.                                             ually expand each event 

  **After:** Any event with an error automatically expands, making the error immediately visible with the red error message.


Tirodem:
QA OK ✅

PR #2336, opened 2026-01-21
🦋 Masking nsec like API keys and passwords with dots placeholders and unmask zpub

### LINKED ISSUE #2263: 🦋 Masking nsec like API keys and passwords with dots placeholders and unmask zpub
## Overall
be-BOP treats passwords and API Keys in a different manner than Nostr Private Keys or Secret Keys known as nsec.
Masking nsec with these password dots placeholders would make admin dashboard more consistent and a bit more secure too.

Tirodem:
Fix #2263

  - Unmask ZPub public key field on /admin/bitcoin-nodeless (change from password to text input)
  - Mask nsec private key field on /admin/nostr (add password input type)
  - Improves security by hiding sensitive private keys
  - Improves usability by showing public keys that are not sensitive

For reminder, any page can be restricted to employees through Settings > ARM (/admin/arm)

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
@@ -80,6 +80,7 @@
 		Private Key (nsec format)
 		<input
 			bind:this={nsecInputEl}
+			type="password"
 			class="form-input"
 			name="privateKey"
 			bind:value={data.nostr.privateKey}
--- a/src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.svelte
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/bitcoin-nodeless/+page.svelte
@@ -25,7 +25,7 @@
 	<label class="form-label">
 		Public key (ZPub)
 		<input
-			type="password"
+			type="text"
 			name="publicKey"
 			class="form-input"
 			required


Tirodem:
@CypherJack it works but we should ticket a new feature for nsec recovery (send to admin, either by email or nostr).

Tirodem:
QA OK ✅

src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte
```diff
@@ -80,6 +80,7 @@
 		Private Key (nsec format)
 		<input
 			bind:this={nsecInputEl}
+			type="password"
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte:83]: Why send the secret to the frontend if it's just to hide it?

If what we want is to let the user know that the secret is set, we should [project](https://github.com/be-BOP-io-SA/be-BOP/blob/91d0da04b71d8889d97c70b60eec7f38617515bb/src/routes/(app)/admin%5B%5Bhash%3Dadmin_hash%5D%5D/s3/%2Bpage.server.ts#L15-L19) that information, [display it accordingly](https://github.com/be-BOP-io-SA/be-BOP/blob/91d0da04b71d8889d97c70b60eec7f38617515bb/src/routes/(app)/admin%5B%5Bhash%3Dadmin_hash%5D%5D/s3/%2Bpage.svelte#L262-L275) and [only allow updates](https://github.com/be-BOP-io-SA/be-BOP/blob/91d0da04b71d8889d97c70b60eec7f38617515bb/src/routes/(app)/admin%5B%5Bhash%3Dadmin_hash%5D%5D/s3/%2Bpage.server.ts#L85-L88).

IMHO, we should add a button to copy the secret to the user's clipboard: The nsec may be generated by be-BOP and we should provide the user with a mechanism to retrieve it. I think however, it would be best if this is a deliverate action and we don't send the password every time the user accesses the page (e.g. when poking around).

Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte:83]: Fair point.
And we can say the same for exposed API secret as password in payment settings forms.
But if we introduce something about this (which is pertinent), I'd rather do it on a new, consecrated ticket, for the whole admin interface, and not just nostr / zpub forms.

Comment (roosemberth) [reply src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte:83]: We already do the right thing for S3 and SMTP. I'd imagine it's easier to improve it opportunistically when we make any changes to each settings page.

What do you think?
If you still want to address this separately, please create the issue and link it here to resolve this discussion.

Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/nostr/+page.svelte:83]: Deal : https://github.com/be-BOP-io-SA/be-BOP/issues/2339

I rather do that separately because that'll force us to do this properly everywhere. Imagine we never touch Paypal page anymore... it would be bad to stick to opportunity.

I'm sticking this to MVP S1 2026 wagon (the next one).

Tirodem:
Damn, we have a release error : https://github.com/be-BOP-io-SA/be-BOP/actions/runs/21251666838

PR #2352, opened 2026-01-30
🐛 Fix real-time Total/VAT on /pos/session

### LINKED ISSUE #2294: 🐛 Total price and VAT are not updated in real time on /pos/session
In order to display total and VAT prices, the /pos/session tab need to be manually updated. However the products are correctly added in real time.

Image

gntw:
Total and VAT stayed at zero while products updated in real-time via /pos/session.
Now priceInfo is computed client-side reactively

Closes #2294


Tirodem:
Seems good , updated in real time with good price, even with PWYW and deposit products ✅ 
@roosemberth for review

PR #2353, opened 2026-01-30
Remove IP2Location CSVs from LFS and fetch GeoIP assets during release

roosemberth:
The large IP2Location CSVs should not be stored in Git LFS to avoid exceeding
LFS quotas and to simplify CI and developer workflows.

These databases are no-longer stored in the repository and are downloaded from
a mirror while building the artifacts.

There are no functional changes, you get the same be-BOP experience, with better plumbing.


roosemberth:
Closing for now. Will address later.

PR #2356, opened 2026-02-02
🦋 Restrict pool changes after mid-ticket print

gntw:
After printing a kitchen ticket (mid-ticket) in POS Touch, printed items are now protected: they cannot be deleted, and their quantity cannot be reduced below the printed amount. Pools containing printed items cannot be deleted either. This prevents revenue loss when staff removes items after they have already been sent to the kitchen/bar.

- Quantity increase and adding new items remain unrestricted
- Moving items between pools is allowed — printedQuantity transfers
    correctly using "unprinted first" principle
- Config toggle in Admin → POS: "Forbid item deletion / qty reduction,
    after mid-ticket print" (enabled by default)

### DIFF
--- a/src/lib/components/ItemEditDialog.svelte
+++ b/src/lib/components/ItemEditDialog.svelte
@@ -8,6 +8,7 @@
 		product: { name: string; maxQuantityPerOrder?: number };
 	};
 	export let onClose: (_: { note?: string; quantity?: number }) => Promise<void>;
+	export let errorMessage = '';
 
 	let note = item.internalNote?.value || '';
 	let quantity = item.quantity;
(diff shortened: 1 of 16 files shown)


Tirodem:
Let's say I add 5 beers.
I print mid-ticket.
I pay 4 beers.
I go back to pool.
I have one beer.
I add 4 more beers, now there's 5.
But, plot twist, it was a mistake, I only wanted to add 2, so, I want to remove 2.
Here's what I have :
image
image
❌ It keeps 5 as it what was printed on the first mid-ticket.
But, as we paid 4 with product split, minimum should now be 1.
So, when I add 4 new beers and don't print mid-ticket, I should be able to remove all 4 to go back to 1.
Expected result here is either "Impossible de réduire la quantité en dessous de la quantité imprimée non-payée : 1", or just no error as we have more product than minimal qty (printed minus paid).

gntw:
fixed ! @Tirodem

Tirodem:
Works ✅ 
@roosemberth for final review

PR #2358, opened 2026-02-03
🦋 Clean order page layout for POS touch mode

### LINKED ISSUE #2354: 🦋 Clean order page on /pos/touch
Feedback from Iva

🆕 updated, see https://github.com/be-BOP-io-SA/be-BOP/issues/2354#issuecomment-3824701604 for final target

> There's too many buttons when an order is done on /pos/touch

Image

Ideas :

- Move this to a collapsed / deploy on demand block after the most useful CTAs

Image

- Keep only 1 CTA on those (the black one to go back to split payment, rename it simply "Continue")

Image

gntw:
The order page (`/order/[id]`) now has a simplified, touch-friendly layout when accessed in POS mode. Print buttons appear at the top, payments are collapsible, and staff notes are displayed inline instead of
requiring navigation to a separate page.

- **Compact POS layout** — optimized for touch screens with larger tap targets
- **Preserved POS context** — changing payment method, paying by card, or saving notes no longer exits POS mode
- **Inline staff notes** — view and add notes without leaving the order page
- **Fixed "undefined" bug** — empty payment detail field no longer shows "undefined"

Closes #2354

### DIFF
--- a/src/lib/components/Order/CardPayment.svelte
+++ b/src/lib/components/Order/CardPayment.svelte
@@ -8,10 +8,15 @@
 
 	export let payment: SerializedOrderPayment;
 	export let orderId: string;
+	export let returnTo: string | undefined = undefined;
+
+	$: payUrl = returnTo
+		? `/order/${orderId}/payment/${payment.id}/pay?returnTo=${encodeURIComponent(returnTo)}`
+		: `/order/${orderId}/payment/${payment.id}/pay`;
 </script>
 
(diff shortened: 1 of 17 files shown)


Tirodem:
Ok except translation ❌

gntw:
> Ok except translation ❌

fixed!  @Tirodem

Tirodem:
Go ✅ 
@roosemberth , ok to merge ? (required for 2pm)

PR #2362, opened 2026-02-04
Font size increasing for items in POS mobile

gntw:
- font size increasing for items in POS mobile


Tirodem:
@roosemberth sorry, I'm merging this one directly considering the small diff.

PR #2372, opened 2026-02-05
🦋 POS tickets - show datetime when first item was added to pool

### LINKED ISSUE #2237: 🦋 Add Date and Time of first added Product to a Pool on printed Ticket
When waiter start adding Products to a pool, the date and time of the first added product should be saved. When a ticket is printed (even if it have no accounting value) the information should be printed as well under the identity of shop/entity like this:

## Translation
Would be nice if it is translated in other supported languages.

## Warning
If the pool is cleared/reset or the all the items have been removed manually this information should be removed too, waiting for the next first added product.

gntw:
POS tickets now display when the table was opened (first item added). This helps track service time from order start to payment.

- Timestamp recorded automatically on first item
- Resets when pool is emptied
- Shows on kitchen tickets, global tickets, and payment receipts

Closes #2237

### DIFF
--- a/src/lib/components/PrintableTicket.svelte
+++ b/src/lib/components/PrintableTicket.svelte
@@ -61,6 +61,9 @@
 	// Number of blank lines at top of each kitchen ticket section (for clamping)
 	export let topBlankLines = 0;
 
+	export let poolOpenedAt: Date | undefined = undefined;
+	export let orderCreatedAt: Date | undefined = undefined;
+
 	// ESC/POS Command Constants
 	const ESC = {
 		// Text formatting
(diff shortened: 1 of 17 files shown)


roosemberth (changes requested):
I think that having “the date of the first object” persisted separatedly from the object is problematic.

However, if we do “date at which the ticket was opened” is fine, but the property is called “firstItemAddedAt”.

So, either rename the property to something like “pool opened at” or retrieve the date from the **item** in the pool and not from the pool.

gntw:
> So, either rename the property to something like “pool opened at” 

Renamed to poolOpenedAt,  @roosemberth

Tirodem:
QA ok ✅

roosemberth:
Thank you!

PR #2373, opened 2026-02-05
🐛 Fix VAT showing 0% in POS Global Ticket

### LINKED ISSUE #2366: 🐛 VAT is set to 0 for global ticket print
When printing a global ticket the VAT is set to zero unlike the receipt ticket which has the correct VAT rate.

gntw:
Global Ticket now correctly shows VAT rate even when products don't have a custom VAT profile assigned.

Closes #2366


roosemberth (approved):
Approving... But:

Why is the VAT computed in the “ticket” module? is the VAT specific to a ticket? I would expect this computation to take place in a “VAT-something” module; this can be a function (since the computation seems complex).

Where are we computing it and how do we make sure the computation is the same everywhere?

You can go ahead and merge, but please look at this and open an issue if need be.

PR #2377, opened 2026-02-05
🦋🤖 Forward receipt to customer from order page (#2375)

### LINKED ISSUE #2375: 🦋 Allow ticket / receipt forward from /order/{id} page
As an employee, when viewing a paid order, I want to be able to forward the receipt  
  to a customer by entering their email address or Nostr npub.

  Behavior:
  - A "Forward receipt" link appears on paid orders, visible to staff only (both       
  classic POS and POS Touch)
  - Clicking it expands an inline form with a contact address input (email or Nostr    
  npub)
  - If an email is entered, the customer receives the order.paid email
  - If a Nostr npub is entered, the customer receives a DM with the ticket content and 
  order link
  - A confirmation message is shown once the receipt has been forwarded

Tirodem:
- Employees can now forward a receipt to a customer directly from the order page     
- Available on paid orders only, for both classic POS and POS Touch
- Supports sending via email (sends the order.paid email) or Nostr DM (sends ticket content + order link)
- Inline form with contact address input (email or Nostr npub), same pattern as the RSVP form
- Success confirmation displayed after forwarding

Fixes #2375

### DIFF
--- a/src/lib/components/Order/ForwardReceiptForm.svelte
+++ b/src/lib/components/Order/ForwardReceiptForm.svelte
@@ -1,0 +1,58 @@
+<script lang="ts">
+	import { enhance } from '$app/forms';
+	import { useI18n } from '$lib/i18n';
+
+	const { t } = useI18n();
+
+	export let actionUrl: string;
+	export let paymentId: string;
+
+	let isOpen = false;
+	let isSuccess = false;
+</script>
+
+{#if isSuccess}
+	<p class="text-green-500">{t('order.forwardReceipt.success')}</p>
+{:else if isOpen}
+	<form
+		action={actionUrl}
+		method="post"
+		class="flex flex-col gap-2"
+		use:enhance={() => {
+			return async ({ result, update }) => {
+				if (result.type === 'success') {
+					isOpen = false;
+					isSuccess = true;
+				} else {
+					await update();
+				}
+			};
+		}}
+	>
+		<input type="hidden" name="paymentId" value={paymentId} />
+		<label class="form-label body-secondaryText">
+			{t('login.authenticate.inputLabel')}
+			<input
+				class="form-input"
+				type="text"
+				name="address"
+				placeholder="email / npub1..."
+				required
+				pattern="^(?!nsec).*"
+				title={t('login.nsecBlockTitle')}
+			/>
+		</label>
+		<div class="flex gap-2">
+			<button type="submit" class="btn btn-black">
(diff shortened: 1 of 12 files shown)


Tirodem:
@gntw @roosemberth Either vibefix or vibeprototype, depending of your's review feedbacks. Pilot's request. QA is OK ✅

gntw:
@Tirodem  got a questsion about the Nostr DM content in forwardReceipt: currently the DM sent to the customer only contains a link — Order #N receipt: <url>. The original issue description mentions "the customer receives a DM with the ticket content and order link". Is this intentional to keep it minimal (just the link), or should we include more order details in the DM (items, total amount, payment method etc.)?

Tirodem:
yeah, I changed the behavior after test-exploring. Trying to recreate the
ticket on nostr dm was a bad idea, sending the link is better. In the
future when implemented we'll use the order.paid nostr template instead,
and for now we'll send the link.

Tirodem:
Filling partial npub =
image
Otherwise, a valid npub is OK, a valid email is OK ✅ 
It's clearly acceptable, but if it can be done in front, with error message + highlighted input field, it would be **great**
BTW, it can be another ticket, as it's currently not done in `/schedule/{id_schedule}/rsvp/{schedule_event}?/notify`.
So :
@gntw , if you're OK with @roosemberth , we can merge like this
@CypherJack , can you create the new ticket to improve this input please

roosemberth (changes requested):
I'm not a fan of introducing behaviour in “Various code grooming” commits:

- The refactor and added permission check in this change is a band-aid.
- We hid new behavior in a “grooming” commit, if we find this behaviour to be faulty, it will waste people's time to figure out the extent of the problem.

Please move this to a separate commit.

gntw:
> I'm not a fan of introducing behaviour in “Various code grooming” commits:
> 
> * The refactor and added permission check in this change is a band-aid.
> * We hid new behavior in a “grooming” commit, if we find this behaviour to be faulty, it will waste people's time to figure out the extent of the problem.
> 
> Please move this to a separate commit.

If we are going to merge all of this, then we probably wanted to squash it into one commit, as we do it in most of our pull requests? There are currently five commits in the branch, and I don't see much point in keeping them separate. I agree, the topic of my commit is ridiculous, but I thought that after the review I would squash all these commits into one so as not to clutter up the main branch. And then I could write a normal description for this single commit.  

@roosemberth is that ok ?

PR #2378, opened 2026-02-05
🐛 Fix POS discount shows "because of subscription" in order notes

### LINKED ISSUE #2369: 🐛 When offering 100% discount on /pos/touch they are displayed as subscription in order notes without the commentary
Image

"because of subscription" is false it is a /pos/touch discount unrelated to any subscription. We should put text set on discount on /pos/touch instead.

gntw:
POS discounts now display correctly in order notes with proper scope and motive.

Before: `Discount applied: product: 100% because of subscription` (bug)
After: `Discount applied (100% for ALL): product: 100% (motive: Passeport gourmand)`

Closes #2369

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -1229,6 +1229,20 @@ export async function createOrder(
 			);
 		}
 
+		const posDiscount = params.cart?.orderTabId
+			? (
+					await collections.orderTabs.findOne(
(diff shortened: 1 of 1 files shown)


Tirodem:
image
image
👌

roosemberth (approved):

roosemberth:
Thank you!

PR #2386, opened 2026-02-09
🦋🤖 Allow yearly subscription (#2385)

### LINKED ISSUE #2385: 🦋 Allow yearly subscription
Right now subscriptions are daily, weekly and monthly.
We must manage yearly subscription.

Tirodem:
Adds "yearly" and "weekly" option to the subscription duration setting.

Administrators can now select "year" or "week" as subscription duration in **Settings > General** (Timing section). When chosen, subscription-type products grant access for one full year/week from purchase date, or one year/week from the current expiration on renewal.
Fixes #2385

### DIFF
--- a/src/lib/server/runtime-config.ts
+++ b/src/lib/server/runtime-config.ts
@@ -3,6 +3,7 @@ import { ObjectId } from 'mongodb';
 import { collections } from './database';
 import { defaultExchangeRate, exchangeRate } from '$lib/stores/exchangeRate';
 import type { Currency } from '$lib/types/Currency';
+import type { SubscriptionDuration } from '$lib/types/SubscriptionDuration';
 import type { DeliveryFees } from '$lib/types/DeliveryFees';
 import { currencies } from '$lib/stores/currencies';
 import {
@@ -71,7 +72,7 @@ const baseConfig = {
 	lightningQrCodeDescription: 'brand' as 'orderUrl' | 'brand' | 'brandAndOrderNumber' | 'none',
 	maintenanceIps: '',
 	brandName: 'My be-BOP',
-	subscriptionDuration: 'month' as 'month' | 'day' | 'hour',
+	subscriptionDuration: 'month' satisfies SubscriptionDuration,
 	subscriptionReminderSeconds: 24 * 60 * 60,
 	reserveStockInMinutes: 20,
 	confirmationBlocksThresholds: {
--- a/src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts
@@ -1,5 +1,4 @@
 import { collections } from '$lib/server/database';
-import { pojo } from '$lib/server/pojo';
 import { error, fail } from '@sveltejs/kit';
 import { userQuery } from '$lib/server/user';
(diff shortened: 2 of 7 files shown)


Tirodem:
QA OK ✅ 
Previously failed :
- monthly stayed monthly even after a change
- going to "subscribers" tab on product led to 500 error

src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts
```diff
@@ -21,7 +21,8 @@ export const load = async ({ params }) => {
 		subscriptions: subscriptions.map((subscription) => ({
 			...pojo(subscription),
 			user: pojo(subscription.user),
-			notifications: subscription.notifications.map(pojo)
+			notifications: subscription.notifications.map(pojo),
+			notificationHistory: (subscription.notificationHistory ?? []).map(pojo)
```
Comment (roosemberth) [inline src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts]: Why is this change needed?

Additionally, `pojo` is a code-smell to leak business models (or database models) into the frontend. It would be more readable to define a frontend model and an explicit projection instead.

src/lib/server/runtime-config.ts
```diff
@@ -71,7 +71,7 @@ const baseConfig = {
 	lightningQrCodeDescription: 'brand' as 'orderUrl' | 'brand' | 'brandAndOrderNumber' | 'none',
 	maintenanceIps: '',
 	brandName: 'My be-BOP',
-	subscriptionDuration: 'month' as 'month' | 'day' | 'hour',
+	subscriptionDuration: 'month' as 'year' | 'month' | 'week' | 'day' | 'hour',
```
roosemberth: I see the same list hardcoded at multiple places.

Can we add a module subscriptionDuration.ts that defines an array of all variants? Then, in the same module, we can define `type SubscriptionDuration = typeof SUBSCRIPTION_DURATIONS[number]`.

The expression on this file would simply be `'month' satisfies SubscriptionDuration`.

The same array could be used to parse the enum in the config page.

Tirodem: Built on legacy, but changed in last push (with SubscriptionDuration). QA WIP.

src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts
Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/product/[id]/subscribers/+page.server.ts]: Commit message updated

src/lib/server/runtime-config.ts
```diff
@@ -71,7 +72,7 @@ const baseConfig = {
 	lightningQrCodeDescription: 'brand' as 'orderUrl' | 'brand' | 'brandAndOrderNumber' | 'none',
 	maintenanceIps: '',
 	brandName: 'My be-BOP',
-	subscriptionDuration: 'month' as 'month' | 'day' | 'hour',
+	subscriptionDuration: 'month' as SubscriptionDuration,
```
roosemberth: `'month' satisfies SubscriptionDuration,`

'as' makes no check and should not be used unless there's a good reason typescript is unable to prove a type.

roosemberth: Edit: updated expression.

roosemberth:
@Tirodem  You can ignore the CI failure in step "Build and release artifacts / Build artifacts (pull_request)" and merge.

PR #2393, opened 2026-02-10
🐛 Fix POS discount with tag filter applying to entire pool instead of matching items only

### LINKED ISSUE #2391: 🐛 Discount on tag now applies on whole pool
Image

Image

Image

Image

As only "Café" has the "Boissons non-alcoolisées" tag, discount should apply only to it, not to the whole pool.

We don't seem to filter the discount by tag anymore.

gntw:
POS discount with a tag filter (e.g. 100% off "Alcohol") was incorrectly applied to the entire pool total in the split page and Global Ticket. Now only matching items are discounted; other items remain at full price.

Closes #2391

### DIFF
--- a/src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts
+++ b/src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts
@@ -20,9 +20,10 @@ import { paymentMethods, type PaymentMethod } from '$lib/server/payment-methods'
 import { resolvePoolLabel } from '$lib/types/PosTabGroup';
 
 type Locale = App.Locals['language'];
-type ProductProjection = ProductForPriceInfo & { name: string };
+type ProductProjection = ProductForPriceInfo & { name: string; tagIds?: string[] };
 
 type HydratedTabItem = {
+	discountPercentage?: number;
 	internalNote?: {
 		value: string;
 		updatedAt: Date;
@@ -52,7 +53,8 @@ async function hydratedOrderItems(
 			vatProfileId: 1,
 			variationLabels: {
 				$ifNull: [`$translations.${locale}.variationLabels`, '$variationLabels']
-			}
+			},
+			tagIds: 1
 		})
 		.toArray();
 	const productById = new Map(products.map((p) => [p._id, p]));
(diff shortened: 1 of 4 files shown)


Tirodem:
Global ticket amount is inconsistent with a discount on only one product type.
Price in UI is good, global ticket is not good, order amount is good, order receipt is good.
I don't understand the origin of the difference.
Otherwise, the discount by tag mechanic is working.

src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts
```diff
@@ -77,9 +79,20 @@ async function hydratedOrderItems(
...
-		items: await hydratedOrderItems(locale, tab.items),
+		items:
+			percentage && percentage > 0
+				? items.map((item) => ({
+						...item,
+						...(!tagId || (item.product.tagIds ?? []).includes(tagId)
+							? { discountPercentage: percentage }
+							: {})
+				  }))
+				: items,
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/split/+page.server.ts]: This is so hard to read. Can we refactor this into something like

```
function applyDiscountPercentage(item: .., percentage: .., tagId: ..) {
  const percentage = percentage ?? 0;
  if (percentage <= 0 || ...) {
    return item;
  } else {
    return {...item, discountPercentage: percentage };
  }
}
...
items: items.map((item) => applyDiscountPercentage(item, percentage, tagId)),
```

src/routes/(app)/pos/touch/tab/[orderTabSlug]/ticket/+page.server.ts
```diff
@@ -80,12 +97,27 @@ export async function load({ locals, params }) {
...
-	const totalAfterDiscount = totalInclVat * discountMultiplier;
 
-	const currency = ((tab.items[0] &&
-		productById.get(tab.items[0].productId.toString())?.price.currency) ??
-		UNDERLYING_CURRENCY) as Currency;
+	let totalAfterDiscount = totalInclVat;
+	if (hasDiscount) {
+		const discountableIncl = itemsForTicket
+			.filter((item) => item.matchesDiscountTag)
+			.reduce((sum, item) => sum + item.totalExcl * (1 + item.vatRate / 100), 0);
```
Comment (roosemberth) [inline src/routes/(app)/pos/touch/tab/[orderTabSlug]/ticket/+page.server.ts]: Can we store the `item.totalIncl` in `itemsForTicket`.

It's **really** unexpected to re-compute the price with VAT in the view (specially since we the price with VAT was already computed somewhere else to come up with `totalInclVat`).

Tirodem:
Fix seems functional ✅ Were change request applied previously ?

Tirodem:
@roosemberth I cannot validate requested changes, will you have the time to confirm them ? Otherwise I'll bypass tomorrow. Thanks.

PR #2424, opened 2026-02-27
🦋🤖 Change default VAT rate to use seller's country (#2415)

### LINKED ISSUE #2415: 🦋 Change the default VAT rate
TBD
Make VAT from seller by default on new be-BOP as it makes more sense for most of our users. Don't update if default is unchanged on in-production be-BOPs that update, only for new instances.

Tirodem:
Enable "Use VAT rate from seller's country" (vatSingleCountry) by default on new be-BOP instances, as it's the most common use case.

Fixes #2415


Tirodem:
@gntw simple diff, merge if convenient

PR #2425, opened 2026-02-27
🐛🤖 Add option to forbid /pos/touch when POS session is closed (#2416)

### LINKED ISSUE #2416: 🦋 Add option in back-office to disable /pos/touch page when POS session is closed
Add a setting in the back-office (ex: "Enforce Open Session for Touch UI") to prevent unauthorized orders or navigation to the touch interface when a POS session is officially closed.

When enabled, if a user attempts to access /pos/touch without an active session, they should be redirected or shown a "Session Closed" screen or similar. Waiters cannot interact with the ordering UI unless an admin has officially opened the POS session.

To be work on

Tirodem:
When a new option is enabled on /admin/pos, if PoS Session is not open, users are redirected from /pos/touch to /pos with information message.

Fixes #2416

### DIFF
--- a/src/routes/(app)/pos/+page.server.ts
+++ b/src/routes/(app)/pos/+page.server.ts
@@ -13,7 +13,7 @@ import {
 	removeOrderTab
 } from '$lib/server/orderTab';
 import { removeUserCarts } from '$lib/server/cart';
-import { getCurrentPosSession } from '$lib/server/pos-sessions';
+import { getCurrentPosSession, requireOpenPosSession } from '$lib/server/pos-sessions';
 import { runtimeConfig } from '$lib/server/runtime-config';
 import { resolvePoolLabel } from '$lib/types/PosTabGroup';
 
@@ -79,6 +79,7 @@ export const load = async (event) => {
 
 export const actions: Actions = {
 	addToTab: async ({ request }) => {
+		await requireOpenPosSession();
 		const formData = await request.formData();
 		const { tabSlug, productId } = z
 			.object({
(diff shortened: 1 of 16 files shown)


Tirodem:
Option should be enabled by default ; /pos/touch redirection works, but /pos/touch/tab/tables-0 is still available, and it shouldn't. /pos/touch and /pos/touch/* should be redirected when option is enabled and PoS session is closed ❌

src/routes/(app)/pos/+page.server.ts
```diff
@@ -67,6 +67,9 @@ export const load = async (event) => {
 
 export const actions: Actions = {
 	addToTab: async ({ request }) => {
+		if (runtimeConfig.posSession.forbidTouchWhenSessionClosed && !(await getCurrentPosSession())) {
+			throw redirect(303, '/pos?errorMessage=pos-touch-session-required');
+		}
```
gntw: This 3-line guard is duplicated 10 times across 4 files. If the redirect URL, error key, or check logic changes, all 10 places need updating.

Lets extract into a shared helper, for example: 

```ts
// src/lib/server/pos-sessions.ts
export async function requireOpenPosSession(): Promise<void> {
  if (runtimeConfig.posSession.forbidTouchWhenSessionClosed && !(await getCurrentPosSession())) {
    throw redirect(303, '/pos?errorMessage=pos-touch-session-required');
  }
}
```

Then each call site becomes: `await requireOpenPosSession();`

Tirodem: It's done !

PR #2431, opened 2026-02-27
🦋🤖 Improve email/npub validation error messages (#2406)

### LINKED ISSUE #2406: 🦋 Improve incomplete/wrong npub error output
## Overview
Related to https://github.com/be-BOP-io-SA/be-BOP/pull/2377#issuecomment-3946532074 

An incomplete npub or with error like added character is treated like an email sending user to an ugly error 422 page.

## How to improve
Show the error message in front for the user highlighting the input field without sending to a new page.

If the user input start with `npub` it should not be referred as email. If it has a `@` it then should be referred as email. If be-BOP is unable to determine if it is a nostr key or an email address we could display a generic error such as `Your contact address is not recognize.`

Tirodem:
Context-specific error messages for hybrid email/npub input fields: invalid email, invalid npub, or generic invalid contact address. Applied to login, RSVP, and schedule subscribe forms with i18n support.

Fixes #2406

### DIFF
--- a/src/routes/(app)/login/+page.server.ts
+++ b/src/routes/(app)/login/+page.server.ts
@@ -15,7 +15,7 @@ import {
 	TWITTER_ID,
 	TWITTER_SECRET
 } from '$lib/server/env-config';
-import { zodNpub } from '$lib/server/nostr.js';
+import { validateEmailOrNpub } from '$lib/server/nostr.js';
 import { renewSessionId } from '$lib/server/user.js';
 import { rateLimit } from '$lib/server/rateLimit.js';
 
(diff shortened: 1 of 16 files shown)


Tirodem:
Hmm, should have wait for https://github.com/be-BOP-io-SA/be-BOP/pull/2377 before doing this one 😅

Tirodem:
Check OK ✅ 
To keep in mind : we hit quite quickly 429 error when sending a nostr receipt then another one
image

src/routes/(app)/login/+page.server.ts
gntw: } catch (err) {
    return { ...base, error: 'invalidOrExpiredToken' };
}

this string is now key for translation (look at the page.svelte), it could not be just a string

PR #2437, opened 2026-03-01
🦋 Add PoS payment subtype filter to reporting page

### LINKED ISSUE #2413: 🦋 Add PoS Payment Subtypes filters to /admin/transactions/reporting
Currently admins can filter PoS Payments as a whole. Adding an individual filter for each one of them would be nice to have.

gntw:
Admins can now filter by individual PoS payment subtypes (cash, check, external-tpe, etc.) on the reporting page. A secondary dropdown appears when "Point of sale" is selected as payment method.

Closes #2413

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/order/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/order/+page.server.ts
@@ -8,7 +8,7 @@ import type { Filter } from 'mongodb';
 import { z } from 'zod';
 
 export async function load({ url, locals }) {
-	const methods = paymentMethods({ hasPosOptions: locals.user?.hasPosOptions });
+	const methods = paymentMethods({ hasPosOptions: locals.user?.hasPosOptions, includePOS: true });
 
 	const querySchema = z.object({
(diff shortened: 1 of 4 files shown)


Tirodem:
Good start, but we have some leftovers.

For example :

### 1
image
image

### 2
image
image

### 3
image
image

I think the issue is about orders with multiple payment PoS subtypes. In case of multiple payments, the other payments should not appears on payment detail and payment synthesis.

Tirodem:
Hmm, my bad ! It's indeed a mistake, but i'ts already on main and seems legacy-_esque_.

Here's on main with Lightning :
image
image

And without payment filter :
image

So this PR is good, but we need another bugfix on orders with multiple payments, because they have weird interaction with reporting.

Tirodem:
BTW, the initial ticket only told about /admin/reporting, but I've expected /admin/order.
@gntw would you rather do the same work on /admin/order, or do you think it's better on a new issue / PR ?

gntw:
@Tirodem will do one more update for this PR that fixes /admin/order

gntw:
@Tirodem done

PR #2448, opened 2026-03-10
🦋 Add sorting for POS payment subtypes

### LINKED ISSUE #2419: 🦋 Add PoS Payments Sub-types as individual Payment methods in /admin/config
We should treat individual sub-types like regular payment methods. Allowing admin to reorder them as they want, mixing them with regular types if wanted.

Currently we have a single option named Point of sale (only for users with POS role), split it into every subtypes that exists and add new ones when they are created.
Image

gntw:
Admins can now reorder POS payment subtypes (Cash, Check, Monero, etc.) using up/down arrows on the POS Payments page (`/admin/pos-payments`), below Edit/Delete on each subtype card. Order is saved automatically on each arrow click. The configured order will be applied for: checkout dropdown, POS split page, and order details.

Closes #2419

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/pos-payments/+page.server.ts
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/pos-payments/+page.server.ts
@@ -193,6 +193,22 @@ export const actions: Actions = {
 		return { success: true };
 	},
 
+	saveSortOrder: async ({ request }) => {
+		const formData = await request.formData();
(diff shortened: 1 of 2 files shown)


Tirodem:
Works, but I'd rather disable the autosave and add a button "Save payment order".

gntw:
@Tirodem now save is manual

Tirodem:
Like a charm ✅

PR #2450, opened 2026-03-10
🦋🤖 Add QR code icon on /order/id/ for event tickets (#2351)

### LINKED ISSUE #2351: 🦋 Add a QR code icon on /order/id/ when there is event ticket
Add a simple icon next to the link to the actual tickets to let people know that there is an event ticket in there. Also doing it next to Display Everything global order if there is ticket

https://www.svgrepo.com/show/490299/qr-code.svg

Tirodem:
Give more focus through a SVG on /order/{id} when a ticket is inside an order.

Fixes #2351


Tirodem:
image

PR #2456, opened 2026-03-11
🐛🤖 Fix product price preview in /pos/touch to use custom VAT rate

### LINKED ISSUE #2454: 🐛 Product price preview in /pos/touch ignores Custom VAT rate
## Overview
The product price preview displayed on the product list in /pos/touch always calculates the ATI (All Taxes Included) price using the default VAT rate, ignoring any Custom VAT rate assigned to the product. The pool order total is correct, only the preview is wrong.

## Steps to Reproduce
1. Assign a Custom VAT rate to a product (ex: 2.6% instead of default 8.1%)
2. Navigate to /pos/touch
3. Observe the price preview displayed on the product tile
4. Add the product to the pool and see real ATI pricing.

## Observed Behavior
- Product tile preview shows ATI price calculated with the default VAT rate (on attached picture: 8.1%) ❌
- Pool order total shows ATI price calculated with the correct Custom VAT rate (on attached picture: 2.6%) ✅
- Results in a price discrepancy between the product list and the actual order total, causing confusion

## Expected Behavior
The product tile price preview should reflect the ATI price using the product's Custom VAT rate when one is assigned to it.

Image

Tirodem:
- Fixed product price preview on /pos/touch tiles always using the default VAT rate instead of the product's custom VAT rate
- Tile prices now correctly reflect each product's assigned VAT profile, matching the actual order total
- Only the preview was affected — pool order totals were already correct

Fixes #2454

### DIFF
--- a/src/routes/(app)/pos/+layout.server.ts
+++ b/src/routes/(app)/pos/+layout.server.ts
@@ -7,21 +7,35 @@ async function getProductsToDisplay(params: {
 	query: Record<string, unknown>;
 	language?: string;
 }): Promise<
-	Pick<Product, '_id' | 'price' | 'name' | 'preorder' | 'availableDate' | 'tagIds' | 'stock'>[]
+	Pick<
+		Product,
+		'_id' | 'price' | 'name' | 'preorder' | 'availableDate' | 'tagIds' | 'stock' | 'vatProfileId'
+	>[]
 > {
 	return collections.products
 		.find({
 			...params.query
 		})
 		.project<
-			Pick<Product, '_id' | 'price' | 'name' | 'preorder' | 'availableDate' | 'tagIds' | 'stock'>
+			Pick<
+				Product,
+				| '_id'
+				| 'price'
+				| 'name'
+				| 'preorder'
+				| 'availableDate'
+				| 'tagIds'
+				| 'stock'
+				| 'vatProfileId'
+			>
 		>({
 			price: 1,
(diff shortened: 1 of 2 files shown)


Tirodem:
image
I'll check in detail tomorrow.
@CypherJack get ready for morning testing and peer with Thomas.
(the product with 2 VAT rates is _Colonel sur place_ and _Colonel à emporter_)

CypherJack:
Seems to work as intended , however the pilot using /pos/touch for selling take-away too seems to be a bit annoyed by the default VAT still appearing in the pool preview. For them it's:

Dont TVA 2.6% CHF XX
Dont TVA 8.1% CHF 0

As you can see it shows CHF 0 so there is no functional problem but if there is no product in the pool that are using the default VAT, it shouldn't appear. This is very minor annoyance with low priority. Should I open a new issue for that?

Tirodem:
We should check this one before mergin :

image

displayed price on right part is with 8.1% while being 2.6% on left part.

Tirodem:
image

Another bug was here : when price currency = EUR, if main currency is CHF, product was displayed on /pos/touch in EUR and when added to a pool it was CHF.
Now, it's converting to main currency, as it should.

Tirodem:
Seen with @gntw : code OK ✅ merging by violence because cannot validate co-authored code

Tirodem:
oopsie, wrong PR merge 😓 was #2490

PR #2457, opened 2026-03-12
🐛 Fix inability to deactivate POS session checkbox settings

### LINKED ISSUE #2382: 🐛 Impossible to deactivate Forbid item deletion / qty reduction, after mid-ticket print for PoS
When unchecked the setting isn't properly applied.

gntw:
Unchecking "Forbid item deletion / qty reduction after mid-ticket print" or "Forbid use of /pos/touch when PoS Session is not opened" had no effect — values stayed `true` in the database. Fixed for both cases. 

Closes #2382


Tirodem:
Works like a charm ✅

Tirodem:
Testing now with redeploy

Tirodem:
Ok ✅ merging

Issue #2463, opened 2026-03-13
🦋 Allow owner to see bookings of a product on schedule


Tirodem:
# Current situation

Right now bookings from a booking product are stored into schedule.events, not in schedule collection.
So we cannot see them while consulting the product's schedule.
It's bothersome for the owner to not have this information.

# Improvement

On /admin/product/[id], on "Booking & Tickets" zone, after "Edit associated schedule" :

- if there's at least 1 entry with corresponding slug (schedule.events.scheduleId = "product:" + products.slug), with status=confirmed (beware of https://github.com/be-BOP-io-SA/be-BOP/issues/2462 , should be done first) and with beginsAt in the future, list up to the first 5 bookings (beginsAt asc). List shall include : beginsAt, endsAt, order ID, link to corresponding order, lenght (endsAt minus beginsAt).

- if there's at least 1 entry with corresponding slug (schedule.events.scheduleId = "product:" + products.slug), with status=confirmed, with beginsAt in the past but with endsAt in the future, list the corresponding booking (supposedly one, but just in case, display all corresponding, for debug cases). List shall include : ⌚, beginsAt, endsAt, order ID, link to corresponding order, lenght (endsAt minus beginsAt).

On /admin/schedule/[id], if there's at least 1 entry with corresponding slug (schedule.events.scheduleId = "product:" + products.slug), with status=confirmed, add a "System event" tabs. On this tab, list all events, with sort by beginsAt desc. Future bookings should be with 🔵, current booking should be with 🟠, previous booking should be with 🟢. Before, add checkbox "Hide past bookings" (default : false). Also add beginsAt filters, "From" to "Until".

PR #2464, opened 2026-03-13
🐛🤖 Fix schedule event status staying pending for non-subscription bookings

### LINKED ISSUE #2462: 🐛 schedule.events status is always "pending" for product booking with no subcription
Self-explaining.

The booking status stays 'pending' because the code that updates it to 'confirmed' is gated behind a subscription check — it only confirms bookings tied to subscription products. For regular (non-subscription) product bookings, that confirmation path never runs, so the status remains 'pending' forever.

Tirodem:
Move scheduleEvents status update to 'confirmed' outside the subscription check so it runs for all paid orders, not just subscription products.

Transparent for end-user but required for #2463 

Fixes #2462

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -2440,12 +2440,15 @@ export async function updateAfterOrderPaid(order: Order, session: ClientSession)
 		await handleOrderTabAfterPayment({ order, session });
 	}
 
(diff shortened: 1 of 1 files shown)


Tirodem:
QA ok ✅

PR #2465, opened 2026-03-13
🐛🤖 Remove receiptReady gating on invoice download buttons

### LINKED ISSUE #2461: 🐛 "Download Invoice" button is disabled and cannot be clicked
### **Description**
The **"Download Invoice"** button is incorrectly disabled (`disabled=""` attribute in HTML), preventing users from downloading their invoice PDF. When the `disabled` attribute is manually removed via the browser's developer tools, the button becomes clickable and the PDF downloads successfully.

### **Steps to Reproduce**
1. Open the link received by mail.
2. Navigate to the section and click to the arrow.
3. Observe that the button is grayed out and cannot be clicked.
4. Open the browser's developer tools (F12), remove the `disabled` attribute from the button's HTML.
5. The button is now clickable and the PDF downloads as expected.

Tirodem:
Accessing to an order from a link, after payment, doesn't allow to download receipt (button is displayed but disable).

Now, if payment exists, button will be always available.

Fixes #2461

### DIFF
--- a/src/lib/components/Order/PaymentActions.svelte
+++ b/src/lib/components/Order/PaymentActions.svelte
@@ -20,8 +20,6 @@
 	export let tapToPayInUseByOtherOrder = false;
 	export let printReceipt: () => void;
 	export let printTicket: () => void;
-	export let receiptReady = false;
-	export let ticketReady = false;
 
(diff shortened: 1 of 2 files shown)


Tirodem:
QA ok ✅

Issue #1867, opened 2025-05-05
🦋 Include (with optional setting) credit card form in /order instead of payment


Tirodem:
# Overview

To spare 1 click for a basic credit card payment and not confuse user.

⚠ Make something smart to cover the payment on deposit / multiple payment order

tl;dr : include the payment form directly in /order/{id} instead of putting a link to it.

PR #2467, opened 2026-03-14
🧪🤖 Embed Stripe payment form in order page (optional setting)

Tirodem:
Add "Include payment form in order page" checkbox on /admin/stripe. When enabled, the Stripe payment form is rendered inline on /order/[id] instead of linking to a separate /pay page.

Refs #1867


Tirodem:
image

Tirodem:
@gntw this little experimentation is satisfying, can we take a bit of time this morning to review that and to see how to make this PR properly and not only for Stripe while keeping (or re-introducing) good code practices ?

PR #2480, opened 2026-03-19
📱🤖 Add option to merge topbar and navbar burger menus on mobile

### LINKED ISSUE #2479: 🦋 Provide alternative management behaviors with burger menus
Some partners expressed their concern about the 2-level navigation on mobile and the weird impression given by a double burger menu :

Image

To be clearer :

Image

# First option

First (on right) is the top bar, second (on left) is the navbar.

While it's easy to say "_if you don't want 2 burgers menu, juste put all your navigation entries in either topbar or navbar, not both_", but the option is still interesting and convenient for desktop :

Image

So, my suggestion for a quickwin here : add new display option in Merch > Layout.

First option would be to not make a burger menu if we have only a few options on the topbar.
For example, instead of this :

Image

We could have this :

Image

But it won't be manageable with more than one link

# Second option

Second option would be to put every topbar & navbar entries in only one list.

We have two choices to consider.

## Every options in navbar menu

Image

With topbar entries first, a space, then navbar entries.

➖ : users are used to burger menu on top, not under a logo.

## Every options in topbar menu

Image

➖ : second line might seem useless with only icons when burger menu is collapsed

Image

Tirodem:
Add a new option on merge > layout of have entries from both navigations (topbar & navbar) in the same burger menu on tablet / mobile.

Fixes #2479

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -178,7 +178,7 @@
 				</nav>
 				<button
 					class="inline-flex flex-col justify-center {data.notResponsive ||
-					!data.links.topbar.length
+					(!data.links.topbar.length && !(data.mergeMobileMenus && data.links.navbar.length))
 						? 'hidden'
 						: 'lg:hidden'} cursor-pointer text-4xl transition header-tab"
 					class:rotate-90={topMenuOpen}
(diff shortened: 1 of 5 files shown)


src/routes/(app)/+layout.svelte
```diff
@@ -191,7 +191,9 @@
 		{#if topMenuOpen}
 			<nav
 				transition:slide
-				class="header print:hidden header-tab flex flex-col lg:hidden text-[22px] font-semibold border-x-0 border-b-0 border-opacity-25 border-t-1 border-white px-10 py-4"
+				class="header print:hidden header-tab flex flex-col lg:hidden text-[22px] font-semibold border-x-0 border-b-0 border-opacity-25 border-t-1 border-white px-10 py-4{data.mergeMobileMenus
+					? ' text-right'
+					: ''}"
```
gntw, line 196: The text-right class gets applied to the entire topbar dropdown nav when mergeMobileMenus is on. This right-aligns both the original topbar links AND the merged navbar links. Was that intentional as a design choice, or should only the navbar group be right-aligned (or neither)? Just want to make sure it looks as expected on mobile.

Tirodem: It's as intended :
image
The important thing I didn't test was the brand name that should be left-align when logo isn't wide, so it's OK 👌 
Thanks for the warning

PR #2488, opened 2026-03-20
🧪🤖 Add theme creation assistant with auto color generation

Tirodem:
Allow to quick create new theme by choosing 2 colors (main, secondary) and 2 fonts (main, secondary).
Also allow to apply a theme from /admin/theme/{id}

Closes #2514

### DIFF
--- a/src/lib/server/theme-assist.ts
+++ b/src/lib/server/theme-assist.ts
@@ -1,0 +1,322 @@
+import type { ThemeData } from './theme';
+
+// --- Color utility functions (inspired by WordPress twentytwenty color-calculations.js) ---
+
+function hexToRgb(hex: string): { r: number; g: number; b: number } {
+	const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+	if (!result) {
+		throw new Error(`Invalid hex color: ${hex}`);
+	}
+	return {
+		r: parseInt(result[1], 16),
+		g: parseInt(result[2], 16),
+		b: parseInt(result[3], 16)
+	};
+}
+
+/** WCAG 2.1 relative luminance */
+function getLuminance(hex: string): number {
+	const { r, g, b } = hexToRgb(hex);
+	const [rs, gs, bs] = [r, g, b].map((c) => {
+		const s = c / 255;
+		return s <= 0.03928 ? s / 12.92 : Math.pow((s + 0.055) / 1.055, 2.4);
+	});
+	return 0.2126 * rs + 0.7152 * gs + 0.0722 * bs;
+}
+
+/** WCAG 2.1 contrast ratio between two colors */
+function getContrastRatio(hex1: string, hex2: string): number {
+	const l1 = getLuminance(hex1);
+	const l2 = getLuminance(hex2);
+	const lighter = Math.max(l1, l2);
+	const darker = Math.min(l1, l2);
+	return (lighter + 0.05) / (darker + 0.05);
+}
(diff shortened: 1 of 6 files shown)


Tirodem:
image

src/routes/(app)/admin[[hash=admin_hash]]/theme/assist/+page.server.ts
```diff
@@ -0,0 +1,51 @@
...
+import { generateThemeData } from '$lib/server/theme-assist';
+import { ObjectId } from 'mongodb';
+import { z } from 'zod';
+import { systemFonts } from '$lib/types/Theme';
+
+const assistFormValidator = z.object({
+	name: z.string().min(1),
+	primaryColor: z.string().regex(/^#[0-9a-f]{6}$/i),
+	secondaryColor: z.string().regex(/^#[0-9a-f]{6}$/i),
+	mainFont: z.enum([systemFonts[0], ...systemFonts.slice(1)]),
```
Comment (gntw) [inline src/routes/(app)/admin[[hash=admin_hash]]/theme/assist/+page.server.ts]: `systemFonts` is declared without `as const`, so the Zod enum built from it provides zero compile-time safety -- any string passes the type checker. Please add `as const` to the `systemFonts` declaration in `Theme.ts`, then use `z.enum(systemFonts)` directly (Zod accepts readonly tuples) instead of the manual `[systemFonts[0], ...systemFonts.slice(1)]` destructuring

src/lib/server/theme-assist.ts
```diff
@@ -0,0 +1,322 @@
...
+	const darkText = '#ffffff';
+
+	const activeTabColor = textOnPrimary === '#ffffff' ? '#000000' : '#ffffff';
+	const darkActiveTabColor = textOnDarkPrimary === '#ffffff' ? '#000000' : '#ffffff';
+
+	// Neutral dark surface (slightly lighter than darkBg for cards/surfaces)
+	const darkSurface = '#1a1a1a';
+	const textOnDarkSurface = '#ffffff';
+
+	return {
```
gntw, line 167: The `{ backgroundColor: colorPair(...), color: colorPair(...) }` object is constructed identically ~10 times for main CTA and ~5 times for secondary CTA. So please extract into `mainCtaColors` and `secondaryCtaColors` variables, then spread where needed to eliminate repetition and make future color changes one-line edits

src/routes/(app)/admin[[hash=admin_hash]]/theme/assist/+page.server.ts
Comment (Tirodem) [reply src/routes/(app)/admin[[hash=admin_hash]]/theme/assist/+page.server.ts]: Fixed in the latest push. systemFonts is now as const in Theme.ts, and the validator uses z.enum(systemFonts) directly. For secondaryFont   (which needs to allow ''), I went with z.enum(['', ...systemFonts] as const) so the readonly tuple inference is preserved.

src/lib/server/theme-assist.ts
Tirodem: Done — extracted mainCtaColors and secondaryCtaColors and spread them at all 9   call sites. While I was at it, I applied the same DRY principle to other repeated colorPair(...) calls and extracted them into named *Pair constants (primaryBgPair, onPrimaryTextPair, secondaryBgPair, onSecondaryTextPair, mainBgPair, surfaceBgPair, mainTextPair, primaryOnDarkPair) so future color tweaks are one-line edits across the whole theme. Let me know if you'd prefer I revert the bonus extractions and keep the diff strictly to your two original asks.

Tirodem:
QA'd again after changes, ok ✅

PR #2495, opened 2026-03-30
⚗ Per-product payment credentials for marketplace (Phase 1)

### LINKED ISSUE #2493: Attach Individual Swiss Bitcoin Pay API Key and IBAN to Product (Marketplace Preparation)
As part of exploring future marketplace functionality, we need to implement the ability to attach unique payment credentials to a specific product. This includes both a Swiss Bitcoin Pay API key and an individual IBAN for bank transfers.

Currently, all payments (crypto or bank wire) are routed through the global store credentials. Our long-term goal is to allow different sellers to list their products so that payments—regardless of the method—are routed directly to their respective accounts.

The current task (Phase 1) is a prototype designed to test the core mechanics of credential substitution during checkout for a single item.

**Future Context (Important for Architecture)**
Phase 2 (Out of scope for this task): In the future, we plan to support mixed carts. If a cart contains products from different merchants, the system will need to generate separate payment instructions (e.g., two different QR codes for Bitcoin or two different PDF invoices with different IBANs) for each unique seller.
The current architecture must be designed with the understanding that checkout logic will eventually be expanded to handle multiple invoices/recipients.

**Technical Notes**
Update the invoice generation service (both for SBP API and PDF generation) to dynamically accept payment credentials as arguments.

Ensure the PDF template for bank transfers can dynamically swap the IBAN, BIC/SWIFT, and Account Holder name if those fields are provided at the product level.

The cart's data structure should remain flexible enough to group items by "Payment Recipient" for Phase 2.

**Acceptance Criteria**
Product Configuration:
Two optional fields are added to the Product entity (database/admin panel):
Swiss Bitcoin Pay API Key (for crypto payments).
Merchant IBAN (for bank transfer payments).

**Checkout Logic (Phase 1):**
Scenario A (Custom Credentials): If a user proceeds to checkout and the cart contains only one product that has individual credentials assigned:

If paying via Bitcoin/Lightning: The system must use the product's specific API key to generate the invoice.
If paying via Bank Transfer: The generated invoice (PDF/Download) must display the product's specific IBAN instead of the global store IBAN.

Scenario B (Standard): If the product does not have individual credentials OR there are other/multiple products in the cart, the system defaults to the standard (global) merchant credentials.

No Regression:
The current standard payment process for regular products must remain intact and fully functional.

gntw:
Products can now have their own Swiss Bitcoin Pay API key and bank transfer details (IBAN, BIC, account holder). When a customer checks out with a single product that has merchant credentials configured, payments go directly to the merchant's account — both for Lightning and bank transfers. Orders with multiple products or without custom credentials work exactly as before.

Admin panel: new "Merchant payment credentials" section in product settings.

Closes #2493

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -91,6 +91,7 @@
 
 	let paymentMethods = product.paymentMethods || [...availablePaymentMethods];
 	let restrictPaymentMethods = !!product.paymentMethods;
+	let hasMerchantCredentials = !!product.merchantCredentials;
 	let vatProfileId = product.vatProfileId || '';
 	let formElement: HTMLFormElement;
 	let priceAmountElement: HTMLInputElement;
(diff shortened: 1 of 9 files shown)


Tirodem:
Mix cart does not work (no need to hurry, it's just for tracking, you weren't expected to dig that far). Displayed IBAN is the default one (from the shop).

Otherwise, for a single product in cart, alternative iban is properly displayed when only said product is on the order.

I'm checking with Lightning (it'll require payment to properly check on SBP dashboard).

Tirodem:
Ok, so, I can either provide an invoice from a SwissBitcoinPay account A for all my products, except my marketplace product that will trigger the invoice from SwissBitcoinPay account B.

With this and the same mechanism for IBAN, we have our proof of concept.

Let's put it on the burner, it'll help me as a foundation for future marketplace & direct merchant payment issue writing.

image

PR #2518, opened 2026-04-10
🐛🤖 Fix VAT breakdown display on /pos/touch and its split view

### LINKED ISSUE #2517: 🐛 /pos/touch Displays Unused Default VAT Profile in Pool & Incorrect VAT Layout in Ticket Preview
## Overview
Two related VAT display issues in /pos/touch:
1. The pool preview always shows the default VAT profile line even when no products in the pool use it (displaying as 0)
2. The ticket preview incorrectly renders all VAT profiles on a single line instead of one line per profile

Linked to https://github.com/be-BOP-io-SA/be-BOP/pull/2516

Image

Image

Tirodem:
- On /pos/touch/tab/{id}: only show "Incl. VAT X%" rows when a product with that rate is actually in the pool (filter zero-amount entries from the VAT breakdown).
- On /pos/touch/tab/{id}/split: replace the confusing "incl. VAT 2.6%, 8.1% CHF Y" subtotal line with just "incl. VAT CHF Y" since the VAT rate detail is already shown in the upper-left breakdown.

That'll lead to a less confusing and more compliant UI for /pos/touch.

Fixes #2517

### DIFF
--- a/src/lib/components/MoveItemsModal.svelte
+++ b/src/lib/components/MoveItemsModal.svelte
@@ -337,7 +337,6 @@
 								totalExcl={leftTabPriceInfo.partialPrice}
 								totalIncl={leftTabPriceInfo.partialPriceWithVat}
 								currency={leftTabPriceInfo.currency}
-								vatRates={leftTabPriceInfo.vat.map((vat) => vat.rate)}
 							/>
 						</div>
 					{/if}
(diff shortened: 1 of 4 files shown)


Tirodem:
QA : ✅ 

image

gntw (approved):

PR #2521, opened 2026-04-13
🐛 Fix wrong VAT-included price on POS touch product tiles

### LINKED ISSUE #2520: 🐛 Wrong VAT included price displayed on product list on /pos/touch
Image

Price VAT excluded being 4.6253 with VAT of 8.1, it should be 5, not 5.01.

I'm pretty sure on this zone we do rounding THEN vat included price calculation.

gntw:
Tile price could be one cent higher than the cart price for the same
product (e.g. 5.01 vs 5.00 CHF). Now they match.

Closes #2520


Tirodem (approved):

Tirodem:
image

PR #2528, opened 2026-04-14
🦋 New discount conditions and promo codes on cart

### LINKED ISSUE #2502: 🆕 Define temporal product discount with new modes
Right now discount is only managed through :
- /checkout with employee hasPOS
- /pos/touch discount per tag
- discount with subscription

With #2500 we're introducing manual discount per product line on /cart for employee hasPOS.

With this issue, we'll improve the discount manager to work outside subscriptions.

Right now, on Merch > Discount, here's what we have :

Image

We intend to make those works for those situations :
- default discount, no condition
- required subscription
- required code on cart
- other cases

So first thing is to **make subscription optional**. It'll become one of many criterias, now.

Second thing is to improve discount creation / update form :

Image

Image

Image

Every element on the form is a "AND" condition to apply discount or not.
No particular element selected (like payment method or delivering country) = considered as "true".
Also, "Beginning date" and "Ending date" should be a timestamp, not a date.

Third thing is to allow promocode display on cart. It should be displayed only if a discount with non-empty promocode exists.

Fourth thing is to prioritize discount application.
Rules are :
- the greatest discount should be applied ; greatest = the one with the greatest global discount for the whole order, not the one with the greatest discount for a product line on each product line
- manual employee actions can override automatic discount

And we should obviously keep ISCA mechanism for mean product price on the last 30 days.

gntw:
Shop admins can now create discounts that depend on many conditions at once — promo code entered on cart, sales channel (web / POS / Nostr bot), payment method, delivery country, customer's contact info, product tags, and product combinations. Subscription is no longer required to grant a discount.

Key points:
- Promo code input appears on the cart page only when an active discount uses one, with anti-brute-force rate limiting (20 attempts per 3 hours)
- On checkout, prices recalculate live as the customer switches payment method — the customer sees the discount applied immediately without reloading
- Product page shows the best applicable discount with a red "-X%" badge and  an optional expiration reminder (both toggleable per discount in admin)
- When several discounts apply, the one giving the biggest total saving on the whole order wins
- Employee-applied POS discounts always take precedence over automatic ones

Closes #2502

### DIFF
--- a/src/app.d.ts
+++ b/src/app.d.ts
@@ -25,6 +25,7 @@ declare global {
 				role?: Role;
 				alias?: string;
 				hasPosOptions?: boolean;
+				recoveryEmail?: string;
 			};
 			email?: string;
 			npub?: string;
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -224,7 +224,8 @@ const handleGlobal: Handle = async ({ event, resolve }) => {
 						login: user.login ? user.login : '',
 						roleId: user.roleId,
 						alias: user.alias,
-						hasPosOptions: user.hasPosOptions
+						hasPosOptions: user.hasPosOptions,
+						recoveryEmail: user.recovery?.email
 					};
 				}
 			}
--- a/src/lib/components/ProductCombinationRow.svelte
+++ b/src/lib/components/ProductCombinationRow.svelte
@@ -1,0 +1,68 @@
+<script lang="ts">
+	import { MultiSelect } from 'svelte-multiselect';
+	import { createEventDispatcher } from 'svelte';
+
+	export let product: { productId: string; quantity: number };
+	export let comboIdx: number;
+	export let prodIdx: number;
+	export let availableProductList: Array<{ _id: string; name: string }>;
+
+	const dispatch = createEventDispatcher<{ remove: null; change: null }>();
+
+	function hasStringValue(o: unknown): o is { value: string } {
+		return (
+			!!o &&
+			typeof o === 'object' &&
+			'value' in o &&
+			typeof (o as { value: unknown }).value === 'string'
+		);
+	}
+
+	function handleSelectChange(
+		e: CustomEvent<{ option?: unknown; type: 'add' | 'remove' | 'removeAll' }>
+	) {
+		const { type, option } = e.detail;
+		if (type === 'add' && hasStringValue(option)) {
+			product.productId = option.value;
+		} else if (type === 'remove' || type === 'removeAll') {
+			product.productId = '';
+		}
+		dispatch('change');
+	}
+</script>
+
+<div class="flex gap-2 items-center mb-1">
+	<div class="flex-[4] min-w-0">
+		<MultiSelect
+			--sms-options-bg="var(--body-mainPlan-backgroundColor)"
+			maxSelect={1}
+			options={availableProductList.map((p) => ({ label: p.name, value: p._id }))}
+			selected={product.productId
+				? [
+						{
+							value: product.productId,
+							label:
+								availableProductList.find((p) => p._id === product.productId)?.name ??
+								product.productId
+						}
+				  ]
+				: []}
+			on:change={handleSelectChange}
+		/>
+	</div>
+	<input
+		type="hidden"
+		name="combinations[{comboIdx}][{prodIdx}][productId]"
+		value={product.productId}
+	/>
+	<input
+		type="number"
+		min="1"
+		class="form-input flex-1 min-w-0 !h-[30px] !py-0"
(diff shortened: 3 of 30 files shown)


Tirodem:
To keep track of yesterday testint session :
- 🙇 screenshots were cut, we should also allow discount on billing address country, not only delivery address country
- 🐛 discount on delivery address doesn't trigger at checkout
- 🐛 discount on customer address doesn't trigger (seen yesterday, customer != user, should use the session contact for this)
- 🐛 if 2+ discounts are valid for an order, the biggest doesn't apply as expected (tested with -60% on bitcoin payment AND -50% on promocode, -60% should apply on bitcoin payment even if -50% promocode is set on cart)

Once it'll be fix I'll check the other discount criterias, like product combos.

gntw:
> To keep track of yesterday testint session :
> 
> * 🙇 screenshots were cut, we should also allow discount on billing address country, not only delivery address country
> * 🐛 discount on delivery address doesn't trigger at checkout
> * 🐛 discount on customer address doesn't trigger (seen yesterday, customer != user, should use the session contact for this)
> * 🐛 if 2+ discounts are valid for an order, the biggest doesn't apply as expected (tested with -60% on bitcoin payment AND -50% on promocode, -60% should apply on bitcoin payment even if -50% promocode is set on cart)
> 
> Once it'll be fix I'll check the other discount criterias, like product combos.

All points and rebase are Done

Tirodem:
❌ It's not "delivering or billing country", it's two different (possibly cumulative) criterias :
image

❌ We should have country labels, not 2 letters country ID :
image

❌ Still have the display issue on /checkout :
image

✅ Country discount application works on /checkout

✅ Application of the greatest discount : OK so far, I need some more destructive testing

Continuing QA

Tirodem:
I tested as customer, instead of employee :
image
Better display but we talked about moving the qty.

Tirodem:
For product combination, alignment issue + can't we just use the same input as "Products" at the end of the form ?
It's better looking, you can type a few letters and find your product, and it's more usable because it uses name instead of slug.
image

Tirodem:
✅ Product combination discount : works
✅ Discount on particular products : works

❌ Discount on product tag : this one is particular, it works, but it needs a change of form.
image
image
It would be weird to say "_I have products from this tag in my cart SO I have 20% discount on all other products_".
Tag product is more a target than a condition.

So change the form like this with a little change of wording and behavior :

image

In that case, if "The discount applies to the whole catalog" is :
- unchecked : apply product filter and tag product filter (OR, to discount either target products AND products with target tags)
- checked : apply to anything, ignore products filter and procut tag filter

Tirodem:
✅ Discount on email : works
✅ Discount on nostr : works
💡 Suggestion : to be able to whitelist `npub*` so every user logged with Nostr npub will have the discount
✅ Discount on complex combination (>= 3) : works

gntw:
@Tirodem all points are Done!

Tirodem:
- billing address discount : ✅ 
- delivery address discount : ✅ 
- npub* discount : ✅ 
- *@* discount : ❌ but we didn't demand it explicitely
- product tag discount : ✅ 
- product combination selector : 

❌ issue on discount display on product page
image

❌💡  Discounts on whole catalog are never applied to delivery fees. It could be deployed as it is but we need to track that and at delivery fees as optional target for discount. Also, can be covered by https://github.com/be-BOP-io-SA/be-BOP/issues/2535#issuecomment-4288615721

❌ Can be done on a future ticket but mobile cart doesn't feel satisfiying :
image
I'd expect something like that :
image

Nearly done !

Delivery fees discount can be done on another ticket. Let's just improve mobile cart and product page with discont and we'll be good.

gntw:
> * billing address discount : ✅
> * delivery address discount : ✅
> * npub* discount : ✅
> * _@_ discount : ❌ but we didn't demand it explicitely
> * product tag discount : ✅
> * product combination selector :
> 
> ❌ issue on discount display on product page image
> 
> ❌💡 Discounts on whole catalog are never applied to delivery fees. It could be deployed as it is but we need to track that and at delivery fees as optional target for discount. Also, can be covered by [#2535 (comment)](https://github.com/be-BOP-io-SA/be-BOP/issues/2535#issuecomment-4288615721)
> 
> ❌ Can be done on a future ticket but mobile cart doesn't feel satisfiying : image I'd expect something like that : image
> 
> Nearly done !
> 
> Delivery fees discount can be done on another ticket. Let's just improve mobile cart and product page with discont and we'll be good.

image

In reality, there isn't enough space to the right of the image because the price can be quite long (for example, in XPF), so I suggest this compromise for mobile cart (its Iphone 12 on the screenshot)

gntw:
Also mobile product page 

image

Tirodem (approved):

PR #2533, opened 2026-04-16
🦋 Add OSB payment provider for French Polynesia

gntw:
A new payment method "OSB" is now available for merchants in French Polynesia. OSB is a card payment gateway (Visa, Mastercard, CB, Amex) that works with the local currency XPF (CFP Franc).

How it works:
- Admin configures OSB credentials in a new settings page (/admin/osb)
- Customers see "OSB" as a payment option at checkout
- Clicking the payment link redirects to OSB's hosted payment page
- After payment, customers return to the order page where the status updates automatically

Key points:
- Supports XPF (CFP Franc) — the standard currency in French Polynesia
- Same architecture as PayPal — redirect to external page + automatic status polling
- No additional JavaScript or external scripts needed on the shop side

Closes #2513

### DIFF
--- a/src/lib/translations/de.json
+++ b/src/lib/translations/de.json
@@ -32,6 +32,19 @@
 			"discardItem": "Verwerfen",
 			"view": "Warenkorb anzeigen"
 		},
+		"discount": {
+			"button": "% Rabatt",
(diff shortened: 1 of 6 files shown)


gntw:
Recreating from correct branch

PR #2537, opened 2026-04-21
🐛 Fix nostr-relaypool reconnect backoff never resetting

gntw:
Fixes the nostr bot going silent after several hours of uptime and staying silent for many hours after a transient network issue — users had to wait for a process restart to get replies to !help, !cart, etc.

Root cause is a reconnect backoff in the upstream nostr-relaypool library that kept growing and was never reset after a successful reconnect, so any prolonged network blip effectively put the pool to sleep until the process was restarted.

After the fix the bot recovers on its own within 10 minutes of any network issue, regardless of how long it lasted. No restart needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,8 @@
 	},
 	"pnpm": {
 		"patchedDependencies": {
-			"date-fns@3.6.0": "patches/date-fns@3.6.0.patch"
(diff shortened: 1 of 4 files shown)


Tirodem (approved):

Tirodem:
Both DM answers and zap nostrevent publication are mostly working (95% of the time I've tested), which is a positive first.
I'm merging this one.

PR #2545, opened 2026-04-23
🐛 Fix missing order numbers after failed checkout

### LINKED ISSUE #2541: 🐛 When payment generation fails, order is not created, but order.number is incremented wrongly
Example : bad stripe api key, bad sumup currency, bad swiss bitcoin pay setting

When I try to generate order with this wrongly set payment method, checkout validation leads to /error redirection

Image

Here we are missing the 25 368 order on sandbox.

gntw:
When a customer tried to pay with a misconfigured payment provider (wrong Stripe API key, incompatible SumUp currency, broken Swiss Bitcoin Pay settings, etc.), the checkout failed and sent them to the error page — but the order counter had already moved forward. That left visible gaps in the order list (e.g. #25367, then #25369, with #25368 missing).

After this change the counter is only consumed when the order is actually saved. A failed checkout no longer leaves a hole in the numbering, so bookkeeping and auditing stay clean.

- Order numbers remain continuous even after payment-provider errors
- No change to existing orders or to the checkout UX
- Booking calendar entries from a failed checkout no longer linger in
  the database either

Closes #2541

### DIFF
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -89,11 +89,11 @@ export async function conflictingTapToPayOrder(orderId: string): Promise<string 
 	return null;
 }
 
-async function generateOrderNumber(): Promise<number> {
+async function generateOrderNumber(session?: ClientSession): Promise<number> {
 	const res = await collections.runtimeConfig.findOneAndUpdate(
 		{ _id: 'orderNumber' },
 		{ $inc: { data: 1 as never } },
-		{ upsert: true, returnDocument: 'after' }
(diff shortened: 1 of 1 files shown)


Tirodem:
https://github.com/be-BOP-io-SA/be-BOP/issues/2546

Tirodem:
https://github.com/be-BOP-io-SA/be-BOP/issues/2547

Tirodem:
Otherwise, it's OK for me ✅ 

image

Before, it was like this :

image

Tirodem (approved):

Issue #2556, opened 2026-04-28
🆕 Wordpress to be-BOP migration tool


Tirodem:
Self-explained.

PR #2558, opened 2026-04-28
🆕🤖 Embed WordPress block CSS scoped under .wp-imported (#2556)

Tirodem:
## Summary

Refs #2556. Stacked on top of #2557 (migration skeleton). Targets the skeleton branch — once that merges to main, this PR's base will be retargeted to main automatically.

When migrated WordPress content is rendered in be-BOP, Gutenberg blocks rely on WP's `block-library` CSS to look right. Without it: `<p class="wp-block-paragraph">`, `<figure class="wp-block-image">`, `.has-text-align-center` etc. fall back to unstyled defaults and the layout breaks.

This PR embeds `@wordpress/block-library` at build time, prefixes every selector with `.wp-imported` via PostCSS (`postcss-prefix-selector`), and serves the result as a single static stylesheet at `/css/wp-imported.css`. Scoped — no leakage on the rest of be-BOP.

## License

`@wordpress/block-library` is **GPL-2.0-or-later**. be-BOP is **AGPLv3**. The "or later" makes them compatible: the combined work stays AGPLv3 (which it already is). No license blocker.

## Changes

- `package.json` — add `@wordpress/block-library` (deps) and `postcss-prefix-selector` (devDeps). Hook `build` and `dev` scripts to run the CSS generation first; expose `build:wp-css` for manual re-runs.
- `pnpm-lock.yaml` — synced.
- `scripts/build-wp-imported-css.mjs` — reads `@wordpress/block-library/build-style/{style,theme}.css`, runs it through PostCSS with `postcss-prefix-selector` to scope every selector under `.wp-imported` (rewriting `:root` / `html` / `body` to the scope as well), writes to `static/css/wp-imported.css`.
- `.gitignore` — output is generated, ignored.
- `src/lib/server/migration/promoters/cmsPage.ts` — promoted CMS pages get their content wrapped in `<div class="wp-imported">…</div>` so the scoped CSS targets exactly the imported content.
- `src/routes/(app)/admin[[hash=admin_hash]]/migration/staged/[id]/preview/+page.svelte` — preview route loads `/css/wp-imported.css` via `<link>` and wraps content in `.wp-imported`. Admin sees properly styled migrated content during review.

## Not in this PR (deferred)

**Public-side rendering of promoted CMS pages still doesn't load `/css/wp-imported.css`.** A customer visiting `/cms/[slug]` for a promoted WP page sees the wrapped content but the CSS isn't loaded yet, so blocks render unstyled — same as before this PR for promoted pages.

A followup PR will wire that on the public `/cms/[slug]` route, probably conditionally (only inject the link when the page content contains `.wp-imported`). Out of scope here because it touches existing customer-facing rendering code, which deserves its own focused review.

## Test plan

- [ ] `pnpm install` succeeds (warning about React peer deps from Gutenberg's transitive deps is expected — we don't use the JS, only the CSS).
- [ ] `npm run dev` runs `scripts/build-wp-imported-css.mjs` first; `static/css/wp-imported.css` is created and prefixed under `.wp-imported`.
- [ ] Migration preview at `/admin/migration/staged/[id]/preview` for a `cmsPage` shows WP-styled content (proper paragraph spacing, headings, image figures, alignment classes…).
- [ ] After promote, the resulting CMSPage in DB has `content` wrapped in `<div class="wp-imported">…</div>`.
- [ ] Public CMS route `/cms/[slug]` for a promoted page still renders (no CSS yet, but no regression).


Tirodem:
Folded into #2557 — single branch as intended.

PR #2563, opened 2026-05-09
🆕🤖 Create /booking page for customer to list bookings with sort and i…

### LINKED ISSUE #2562: 🆕 Create /booking page for customer to see booking products and sort by product name and date / hour
Self-explaining
Same philisophy than /order or /subscription

Tirodem:
…Cal export

For coworking use, when you book quite often, even few times a day, users need a booking dashboard more practical than /orders

Now, people can check directly their booking without accessing their orders one by one.

image

image

Close #2562

### DIFF
--- a/src/lib/translations/de.json
+++ b/src/lib/translations/de.json
@@ -26,6 +26,21 @@
 		"agreement": "Ich habe die Zugangsbedingungen gelesen, erfülle sie und respektiere sie",
 		"navigate": "Weiter stöbern"
 	},
+	"booking": {
+		"exportIcal": "iCal exportieren",
+		"listTitle": "Ihre Buchungen",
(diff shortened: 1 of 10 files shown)


CypherJack (commented):
### Review
It works according to my quick tests. ~~However I find it counter intuitive that sorting by `Booking date` is the date when the order was complete (billing date) and not when the appointment is taking place in the customer agenda. To me it should be important to have Sort by Booking Date and by billing date with the default being Booking Date.~~

#### UI is hard to read
There is no hierarchy of informations. Making it hard to read honestly.

Image
.

Also if the schedule is taking place in one day the date should appear only once.

Image
.

Here is a proposal that I find more intuitive

Image

CypherJack (approved):
Good!

PR #2571, opened 2026-05-19
🦋 Delivery zones — group countries under one shipping fee

### LINKED ISSUE #1857: 🚚 Allow to create shipping zones in delivery management
TBD

gntw:
Until now a shipping fee could only be set per individual country or for "Other countries". For a worldwide shop that meant creating dozens of separate options just to cover regions like the EU or Africa.

Admins can now create a named "delivery zone" — a group of countries that share a single shipping fee. Zones work both in the global delivery settings and as a per-product delivery override.

- Define a zone once (name + list of countries + fee) instead of one option per country
- A specific country option still wins over a zone; a zone wins over "Other countries"
- Country pickers hide countries already used elsewhere, so zones can't overlap
- Existing shops are unaffected — no setup required, everything works as before

Closes #1857


CypherJack:
## Review: Delivery Zones & Forbidden Countries

> Non-blocking improvement suggestions. None of these should hold up merging — each can become a follow-up issue prioritized against deadlines.


## 🗂️ Admin – Global delivery settings
### 1. Countries and zones share the same visual language
**Problem:** Individual countries and delivery zones use identical card styles. "Other countries" (a blacklist-based zone) is rendered among individual countries, and zones have no visible bold title, making the list hard to scan when both types coexist.

**Suggestions:**
- Split the list into two labeled sections: Individual countries / Delivery zones
- Display the saved zone name in bold; keep the edit form below it
- Replace the single "Add fee option" button with two distinct actions: one for individual countries (country picker), one for delivery zones (Custom zone / Other countries)


### 2. Duplicate countries are hidden, should be grayed out instead
**Problem:** Countries already used elsewhere vanish from pickers entirely. A missing entry is more confusing for admins than a visible-but-unavailable one, it's unclear whether the country was omitted on purpose or by mistake.
**Suggestion:** Keep duplicates visible in the list, grayed out and unselectable. Nice to have: a tooltip indicating which option already claims that country.

### 3. No UI feedback on adding a country/zone.
**Problem:** If admin user has already multiple countries and zones adding a new one does not provide any visual feedback and could bring confusion.
**Suggestion:** Highligthing the newly created elements and displaying a pop-in notification `{COUNTRYNAME} was added to your delivery option` or `A new custom delivery zone was created`.

### 4. Import/Export for country lists
**Suggestion:** Allow creating or populating a custom zone by importing a list of countries (JSON or similar), enabling reusable presets like EU or LATAM.

Import rules:
- Countries not used anywhere → imported with price, if none set to `0`, fully editable like manually created custom zones.
- Countries already claimed by another option → flagged in red; admin must resolve the conflict before saving (manually remove from the existing option or exclude from the import)


## 📦 Admin – Product-level delivery
### 5. Import a global zone into a product
**Problem:** At product level, only creating new custom zones from scratch is supported. Admins managing products with different delivery scopes must recreate the same zone repeatedly.
**Suggestion:** Add an "Import from global" action on product delivery. Importing creates a copy of the global zone at product level unlinked from the original allowing per-product adjustments without touching global settings. This also makes global zones act as reusable templates. Also having the option to import from a JSON file (or similar) like mentioned in section 3 would be nice.

### 5. Disabled zones on global delivery
Depends on 5. Global zones need an enable/disable toggle to make the template pattern work cleanly.
Proposed behavior:
- Enabled: enforced as a global rule; countries inside count as "used" and can't be claimed by another enabled option
- Disabled: not enforced globally, but still importable at product level
- Countries may appear in multiple disabled zones; conflict rules only apply across enabled options
- Individual countries can't be disabled. They must be deleted to be removed

> For now having a import/export might be more straightforward for us and bring not much friction to admin comparing to fully recreating config each time you need Product level zones.


## 🛒 Customer cart
### 7. No visual indicator on the affected cart line(s)
**Problem:** When a cart item can't ship to the selected country, only a global banner is shown. On a large cart, it's not obvious which product(s) are causing the block.
**Suggestion:** Add a lightweight inline warning ( 🔴 , ❗️or ⚠️) directly on the affected product line(s), alongside the existing global message, no change to th  checkout logic.

CypherJack:
## Review: Remaining items after fixes, tested on branch `feature/delivery-zones-blacklist__1857-1855`

Most things works the template, importing in products and such are working properly after some tests, there is only one bug that need to be address and a few thing discussed live that you probably forgot as some were not in the PR Review so here is everything I've found:

### 👀 Missed implementations

#### 1. Save/Update button not sticky
The back-office pattern already established for other sections uses a sticky Save/Update button fixed to the screen. This section doesn't follow it. Should behave consistently with the rest of the admin.

#### 2. "Other countries" grouped with individual countries
"Other countries" is a zone (blacklist-based) and should be rendered inside the **Delivery zones** group, not the **Individual countries** group. The grouping logic needs to account for it.

#### 3. Zone group should appear above individual countries
Admins using both zones and individual countries will most likely manage zones first as they cover more ground. Current order:
1. Individual countries (+ Other countries)  ← wrong
2. Delivery zones

Expected order:
1. Delivery zones (including Other countries)  ← correct
2. Individual countries

### 🐛 Bug
#### Unsaved delivery zone not factored into country availability

Countries already claimed by an **unsaved** delivery zone can still be added as individual countries. The guard is incomplete.

| Country already used in… | Blocked from individual country picker? |
|---|---|
| Other countries blacklist (saved or unsaved) | ✅ |
| Another individual country (saved or unsaved) | ✅ |
| A **saved** delivery zone | ✅ |
| An **unsaved** delivery zone | ❌ should be blocked |

The unsaved zone state isn't being included in the "used countries" computation. It should follow the same rule as the others saved or not, a country in-use is off-limits. However it works well for the blacklist of other countries my guess is it's because they share the same svelte dropdown menu picker while the individual countries uses the browser default one.

Tirodem:
- some form updates to bring (full listing later), including adaptive country list for custom zone, custom zone selection on dropdown (a button "Add new zone" should be better) and distinguish better zones and simple countries
image

- TBD

PR #2572, opened 2026-05-21
🦋 Forbid delivery to specific countries

### LINKED ISSUE #1855: 🚚 Allow to blacklist countries from "Other countries" option in delivery settings
# Overall

Right now we can :
- whitelist a country for delivery
- whitelist all other countries under the same option

But sometimes you can, for example :
- deliver in a specific way if 8 different countries
- deliver in a generic way for all other countries
- but you are not allowed to deliver, or you cannot possibly deliver, in a couple of countries

We need an option to keep the current system but also blacklist some countries.

# Design

On /admin/config/delivery , if the option "Other countries" is selected, allow to set a country blacklist.
Image
Ideally, countries available in this selector should not include countries that are already set as Delivery option on the same page.

gntw:
Shop owners can already set a delivery fee per country, or one fee for "all other countries". This adds the missing piece: a list of countries you simply cannot deliver to.

Under the "Other countries" delivery option there is now a "Forbid delivery in these countries" field. Orders shipping to a forbidden country are blocked at checkout. It works both in the global delivery settings and per individual product.

- Pick forbidden countries from a list (countries already used by another option are hidden)
- A specific country fee always takes priority — setting a fee for a country un-forbids it
- A customer choosing a forbidden country sees a clear "not available" message and can't proceed
- Existing shops are unaffected — no setup required, everything works as before

Closes #1855

### DIFF
--- a/src/lib/cart.ts
+++ b/src/lib/cart.ts
@@ -426,6 +426,7 @@ export function computeDeliveryFees(
 			Product,
 			| 'price'
 			| 'deliveryFees'
+			| 'defaultBlacklist'
 			| 'applyDeliveryFeesOnlyOnce'
 			| 'shipping'
 			| 'requireSpecificDeliveryFee'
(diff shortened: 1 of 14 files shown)


CypherJack:
### Tested in a common branch with other Delivery v2 Issues

It was on branch `feature/delivery_v2__discount_v2___1855_1857_1854_2502`
blacklist creation and persistence, order refusal for blacklisted countries, selector symmetry, and product-level blacklist.

**Tests run:**
- Added countries to the global blacklist: chips save correctly, confirmation message appears, values persist on page reload.
- Attempted checkout with a blacklisted country delivery address: order is correctly refused ("Some products are not available in your country"). Switching to a non-blacklisted country lets the order through normally.
- Verified selector symmetry: countries already in the blacklist are not offered in the "Add fee option" picker as valid options, and countries with a dedicated option appear grayed out in the blacklist picker. Both directions work consistently.
- In "Fee depending on product" mode: set Brazil as blacklisted at product level: checkout refuses the order. Removing the blacklist from the product re-enables delivery to Brazil.

PR #2586, opened 2026-06-05
🦋 Set subscription length per product

### LINKED ISSUE #2388: 🦋 Move Subscription duration (Timing?) to Product level
## Overview
Currently the duration of subscription products is set globally in `/admin/config` or `Settings > General > Timing` would make more sense to change it from global setting to per product configuration. That way admin cal sell various subscription duration on one single be-BOP.

## For existing subscription product
Existing subscription product should inherit from global setting at the time of the update bringing the change. Also adding a warning under `Settings > General > Timing` such as `Subscription duration moved in product settings for greater flexibility`.

gntw:
Until now the renewal period for subscriptions (weekly, monthly, yearly, etc.) was a single global setting shared by the whole shop. Admins can now choose the length on each subscription product individually, so one be-BOP can sell, for example, a monthly and a yearly subscription side by side.

- Each subscription product has its own duration; leaving it on "Default" keeps using the shop-wide value, so nothing needs touching to keep current behaviour
- All existing subscription products automatically inherit the previous global value when this update is rolled out — no manual reconfiguration.
- The old global duration setting is replaced by a note pointing admins to the product page; the reminder timing setting stays where it was.
- Customers now see the renewal period (for example "Renews every week", translated in all languages) on the product page, the subscription page, the cart and checkout.

Closes #2388

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -39,6 +39,7 @@
 	import { typedFromEntries } from '$lib/utils/typedFromEntries';
 	import { type Day, dayList, productToScheduleId, type BookingSummary } from '$lib/types/Schedule';
 	import { formatDuration } from '$lib/utils/formatDuration';
+	import { formatDistance } from 'date-fns';
 	import { computeVatRate } from '$lib/utils/vat';
 	import { SUBSCRIPTION_DURATIONS } from '$lib/types/SubscriptionDuration';
 
@@ -99,6 +100,8 @@
 	let paymentMethods = product.paymentMethods || [...availablePaymentMethods];
 	let restrictPaymentMethods = !!product.paymentMethods;
 	let vatProfileId = product.vatProfileId || '';
+	let subscriptionDuration = product.subscriptionDuration || '';
+	let subscriptionReminderSeconds: number | '' = product.subscriptionReminderSeconds || '';
 	let formElement: HTMLFormElement;
 	let variationInput: HTMLInputElement[] = [];
 	let disableDateChange = !isNew;
@@ -324,9 +327,6 @@
 
 	$: if (product.type === 'subscription') {
 		product.payWhatYouWant = false;
-		if (!product.subscriptionDuration) {
-			product.subscriptionDuration = 'month';
-		}
 	}
 
 	$: if (product.payWhatYouWant) {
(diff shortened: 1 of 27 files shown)


CypherJack:
Seems to works as expected. One pre-existing behaviour worth flagging though.

---

Customers currently can't purchase multiple durations of the same subscription product in a single cart. So buying 3 months of a monthly subscription requires 3 separate checkouts and payments. The subscription duration does increment correctly in the back-office when done this way, but there's no single-payment path for it.

This is particularly rough for short-cycle subscriptions (hourly, daily, weekly or even monthly) where the workaround is most likely to come up in practice.

Edit:
Was able to purchase a currently running subscription probably because I wasn't logged in but used the same contact email or npub.

I got this error when using email, but not the npub, however active subscriber uses the npub...:
image


Either way I see no reason for us to prevent customer to order multiple time the same subscription to get multiple hours/days/weeks/months or even years within a single purchase.

CypherJack:
The "Default option" should not appear in the subscription product duration selector.
As the global duration setting is no longer editable by admins in General config so it is not relevant to keep it as a product-level duration option.

Showing "Default" as a selectable value implies the global setting can still be configured, which is no longer true.

Tirodem:
"Either way I see no reason for us to prevent customer to order multiple time the same subscription to get multiple hours/days/weeks/months or even years within a single purchase."
=> will probably be incompatible with the free trial + discount on subscription management (instead of renewing my subscription, I use a new one with discounted price on my contact information)

Tirodem:
image

is now back on admin/config

on product :

image

"subscription reminder" should be there also, otherwise it cannot be exploited properly

slashbinslashnoname:
Approved

slashbinslashnoname (approved):

Tirodem:
Bug found : when a subscription is created, "Subscription reminder" is not saved (it's saved only after the product creation), and `subscriptionReminderSeconds` is stored with 0 value on database, making the reminder disabled.
Fix in progress.

Tirodem:
@slashbinslashnoname ok from my side, you can give the final review please

slashbinslashnoname:
> @slashbinslashnoname ok from my side, you can give the final review please

Approved, that behavior is tricky to detect but you spotted it

Issue #2592, opened 2026-06-12
🆕🧑‍🔬 Add env override mechanism for self-hosted → hosted tenant migration (`ALLOW_ENV_OVERRIDE`)


Tirodem:
## Context

We're building a stack-tenant deployment service for be-BOP, and we need to support migration of existing self-hosted be-BOP instances to the hosted service **without exposing the merchant's MongoDB credentials to the operator**.

Today, be-BOP can't start without a MongoDB connection. For a migrating merchant, that means either we provision them a fresh DB (losing their data) or they share their existing connection string with us through provisioning (loss of trust).

## Goal

Allow a fresh be-BOP instance to:

1. Boot with a bootstrap local MongoDB

2. Have the merchant configure their existing S3 via the current `/admin/s3` UI

3. Upload an env override file to their own S3 bucket via a new dedicated admin route

4. Restart, read the override at boot, connect to their existing MongoDB

5. Recover their entire historical config (S3 conf, products, orders, accounts…) from their own data

### New admin route `/admin/be-bop`

Under "Node Management" section (next to NostR, SMTP, Emails). Title: "Installation be-BOP".

Three sections on the page:

- **Restart** — buttons "Restart be-BOP" (exit 100) and "Update be-BOP" (exit 101)

- **Use previous be-BOP version** — dropdown 1-10 + "Update & restart" button (exit 110-119, N-k against the 10 latest public GitHub releases)

- **Override settings** — textual instructions, file upload (text format, `.env`-style), "Save" button, "Delete override" button

Access control: super-admin role + `ALLOW_ENV_OVERRIDE=true` set at deployment time.

### Exit code nomenclature

Source of truth in this repo, `be-bop-tooling` aligns. Documented in `docs/exit-codes.md` (to be created).

| Code | Meaning |
|---|---|
| 0 | Clean shutdown, do not restart |
| 100 | Restart |
| 101 | Update to the latest public GitHub release |
| 110-119 | Rollback to N-k (110=N-1 … 119=N-10), where N is the current public GitHub release |
| 102-109, 120-125 | Reserved for future extensions |

Strict semantics on exit 0: orchestrator runs `Restart=on-failure`, so any unintended `process.exit(0)` would take the tenant down. To be protected by a CI guard (grep / lint) on new occurrences.

### Override file

- Stored in the merchant's S3 bucket (exact path/name TBD)

- Format: standard `.env` style (one `KEY=VALUE` per line)

[...]

PR #2593, opened 2026-06-12
🆕🤖 Add env override for self-hosted → hosted migration (#2592)

Tirodem:
## Summary

Implements #2592. Allows a fresh be-BOP instance to switch its MongoDB connection at next boot by reading an override file the merchant uploads to their own S3 bucket — no operator-handled credentials. Gated by `ALLOW_ENV_OVERRIDE`, super-admin only.

- **New admin page `/admin/be-bop`** under Node Management. Three sections: Restart / Update, rollback dropdown (N-k against the 10 latest public GitHub releases), and override file upload (`MONGODB_URL`, `MONGODB_DB`, `MONGODB_DIRECT_CONNECTION`, `MONGODB_IP_FAMILY` whitelisted).
- **Boot-time discovery** in `src/lib/server/env-override.ts`, imported as the very first line of `hooks.server.ts` (top-level await) so the override is applied **before** `database.ts` captures `MONGODB_URL`. The bootstrap Mongo is used only briefly to read the merchant's S3 config; the override file lives at `_bebop/override.env`.
- **Exit code nomenclature 0 / 100 / 101 / 110-119** introduced as the contract with `be-bop-tooling`; constants in `src/lib/server/exit-codes.ts`, documented in `docs/exit-codes.md`.
- **Fallback** on override application failure: marker `_bebop/override.failed` written to S3 (prevents retry loops) and `envOverrideFailed` flag set in `runtimeConfig` (surfaces a warning banner). Save/Delete actions clear both.

Surgical scope: 5 existing files touched (1-4 lines each), 5 new files. `pnpm check` clean, `pnpm lint` clean, `pnpm format` applied.

## Out of scope (deferred)

- Email / Nostr fallback notification — marker + flag mechanism is in place, only the wiring to `email.ts` / `nostr.ts` is missing.
- CI guard against new unintended `process.exit(0)` calls (would trip the strict exit-0 semantics agreed with the orchestrator).
- Tests.
- Merchant-facing documentation (a How-to block is included in the admin page UI itself).

## Test plan

- [ ] `pnpm check` and `pnpm lint` pass
- [ ] With `ALLOW_ENV_OVERRIDE` unset, route `/admin/be-bop` returns 404
- [ ] With `ALLOW_ENV_OVERRIDE=true` but non-super-admin user, route returns 403
- [ ] With flag on and super-admin, page renders three sections
- [ ] Upload of a valid override file lands the file at `_bebop/override.env` in the bucket
- [ ] Upload of a file with a non-whitelisted key is rejected with a clear message
- [ ] Restart button triggers `process.exit(100)` (verify via supervisor logs)
- [ ] Update button triggers `process.exit(101)`
- [ ] Rollback dropdown values 1-10 trigger `process.exit(110..119)`
- [ ] At boot with a valid override pointing to a reachable Mongo: `MONGODB_URL` is overridden, be-BOP connects to the override target
- [ ] At boot with an override pointing to an unreachable Mongo: marker is written, flag is set, banner appears in admin
- [ ] At boot with the marker present: override is skipped silently
- [ ] Save action clears any stale marker + flag
- [ ] Delete action removes override + marker + flag


Tirodem:
:ice_cube: for a bit because overriding database settings goes against ISCA obligations... it can be useful but it can be hurtful in some situations for be-BOP use.

PR #2598, opened 2026-06-15
 🦋 Let shop owners add their own fields to checkout

### LINKED ISSUE #2212: 🦋 Allow owner to create custom input fields on /checkout to collect additional datas for order processing
On /admin/config , add a link to /admin/config/checkout-fields :

Image

On /admin/config/checkout-fields, use elements from Payment Settings > PoS Payments for the display and the forms :

Image

Image

Image

For now I think about 3 input types :
- options list (like previous screenshot)
- free input : you allow free text, choose the size of input, eventually put text limitation or input type (alphabet, numerical, etc)
- address : give a full address form in addition to the delivery form (but that way you can collect an address without needing to manage delivery on a product). you might be able to choose either a personal address or a company address.

Once it's done, store those inputs in orders metadatas, and allow employees to see them on /orders/{id}

gntw:
Shop owners can now create custom questions on the checkout page to collect extra information for processing an order — for example a gift-wrapping choice, a free-text note, or an extra address. Fields are managed from a new admin page (Config → "Set checkout additional fields") that works just like the PoS Payments setup.

Three input types are available:
 - Option list — the customer picks one value from a list you define
 - Free input — free text, with optional length limit and format (text / numbers / letters+numbers)
 - Address — a full address form, with a personal/company toggle for the customer

Key points:
 - Each field can be marked mandatory and shown or hidden independently
 - Whatever the customer enters is saved on the order and shown on the order page
 - Answers are stored as a snapshot, so renaming or removing a field later never changes orders that were already placed
 - A field that's already used by orders can't be deleted — disable it instead

Closes #2212

### DIFF
--- a/src/lib/components/CheckoutFieldInput.svelte
+++ b/src/lib/components/CheckoutFieldInput.svelte
@@ -1,0 +1,129 @@
+<script lang="ts">
+	import { useI18n } from '$lib/i18n';
+	import type { CheckoutFieldDisplay } from '$lib/types/CheckoutFieldConfig';
+
+	export let field: CheckoutFieldDisplay;
+
+	const { t, countryName, sortedCountryCodes } = useI18n();
+
+	let isCompany = false;
+
+	$: addressPrefix = `checkoutFieldAddress.${field.slug}`;
+	let inputMode: 'numeric' | 'text';
+	$: inputMode = field.free?.format === 'number' ? 'numeric' : 'text';
+</script>
+
+{#if field.type === 'options'}
+	<label class="form-label">
+		<span
+			>{field.isPersonalData ? '🛡️ ' : ''}{field.label}{#if field.required}<span
+					class="text-red-500">*</span
+				>{/if}</span
+		>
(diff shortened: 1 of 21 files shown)


Tirodem:
:heavy_check_mark:  mandatory "*" to display better
image

:heavy_check_mark:  custom field deletion does not work

:heavy_check_mark:  personal data cleaning should also clean customCheckoutFields ; also RGPD fields should have a flag on admin, and an icon on checkout and order


For future tickets :
- on custom address field, we should add an option to either ask for name + surname OR for a company name (only 1 field) that should be used for relais locker (for example) ; currently we can tick a checkbox to add a company name and a vat number, but it needs a bit of fine tuning
- i18n
- pre-filled droplists (ex : country)
- easy access to informations for admin through UI

Tirodem:
image
:heavy_check_mark: be-BOP can have orders with X products, and when products are deleted, we don't forbid it because they are in order. Orders survive their products, their variations, so it should survive a deleted custom checkout field.

PR #2595, opened 2026-06-15
🆕🤖 Allow cart creation through shared URL (#2594)

Tirodem:
## Summary

Closes #2594.

Implements quick-link cart creation: an admin can craft a URL like `/cart?slug=tshirt-noir&qty=2&slug=mug&qty=1` and share it. Landing on that URL drives the customer through a guided flow to either accept, reconcile with an existing cart, or roll back failed additions.

## General behaviour

1. **Landing** : `GET /cart?slug=…&qty=…` is **pure** — no side-effects, safe against SvelteKit prefetch-on-hover. The page renders the regular cart plus an overlay popin that drives the next step.
2. **Decision step** depends on the current cart state:
   - empty → **Confirm popin** (Add to my cart / Cancel)
   - non-empty → **Reconcile popin** (Replace / Merge / Keep)
3. **Execution step** : a single POST action attempts the additions sequentially through the canonical `addToCartInDb` (same path as the product page Add-to-cart, same validations, same stock reservation).
4. **Outcomes** :
   - all OK → redirect to `/cart?createdFromUrl=1` (green confirmation banner above the cart)
   - any item rejected → **Errors popin** with 3 recovery options:
     - *Keep the cart as is*
     - *Empty the whole cart*
     - *Undo additions from the link* — restores the cart to the exact state it was in **before** the operation started
5. **Kill switch** : `runtimeConfig.allowCartFromUrl` (default `false`, exposed in Admin → Settings → Cart). When OFF, URL params are silently ignored; the cart renders normally.

## Fine rules per product type

| Product type | Behaviour with `?slug=…&qty=N` |
|---|---|
| **Standard** | single `addToCartInDb(qty=N)` — increments existing line if any |
| **Standalone** | loop `addToCartInDb(qty=1)` N times — N new lines |
| **PWYW** | `customPrice = recommendedPWYWAmount ?? product.price.amount`, currency from `product.price.currency` ; PWYW forces standalone → loop applies |
| **Variations product** | rejected with `cartFromUrl.errors.reasonVariationRequired` ("Requiert un choix de variation") |
| **Booking product** | rejected with `cartFromUrl.errors.reasonBookingRequired` ("Requiert un créneau de réservation") |
| **Out of stock / maxQty / notForSale / cart line limit** | reuse existing keys (`product.outOfStock`, `pos.cart.maxQuantityReached`, `product.notForSale`, `cart.reachedMaxPerLine`) |
[...]

PR #2611, opened 2026-06-19
🆕🤖 Cart share button + promoCode in /cart?slug=… URLs (#2596)

### LINKED ISSUE #2596: 🦋 Add cart share feature (link generator) on employee cart
Following #2594 , allow employee to make his/her own cart then click on a "share" button to obtain directly the link

Tirodem:
Closes #2596.

On `/cart`, a **"Copy share link 📋"** button next to the title copies a `/cart?slug=…&qty=…` link reproducing the current cart on the receiver's device (via #2595).

- Visible to everyone (no role check). Requires `Allow cart creation through shared URL` ON and at least one shareable item.
- Items with variations or a booking slot are excluded from the link.
- If a promo code is applied, it's appended as `&promoCode=…` and validated on the receiver side via the same chain as the manual promo input (same rate limit, same error messages).
- Click: `📋` → `☑️ Link copied` for 2s.

Available in all 7 languages.

### DIFF
--- a/src/lib/translations/de.json
+++ b/src/lib/translations/de.json
@@ -69,6 +69,10 @@
 			"invalid": "Gutscheincode nicht gefunden oder nicht auf Ihre Bestellung anwendbar.",
 			"rateLimited": "Zu viele fehlgeschlagene Versuche. Bitte versuchen Sie es später erneut."
 		},
+		"share": {
+			"button": "Meinen Warenkorb teilen",
+			"copied": "Link kopiert"
+		},
 		"empty": "Der Warenkorb ist leer",
 		"fillProductAlias": "Produkt-Alias ausfüllen",
 		"items": "Produkte",
--- a/src/lib/translations/en.json
+++ b/src/lib/translations/en.json
@@ -69,6 +69,10 @@
 			"invalid": "Promo code not found or not applicable to your order.",
 			"rateLimited": "Too many failed promo code attempts. Please try again later."
 		},
+		"share": {
+			"button": "Share my cart",
+			"copied": "Link copied"
+		},
 		"empty": "Cart is empty",
 		"fillProductAlias": "Fill product alias",
 		"items": "Products",
--- a/src/lib/translations/es-sv.json
+++ b/src/lib/translations/es-sv.json
@@ -69,6 +69,10 @@
 			"invalid": "Código promocional no encontrado o no aplicable a tu pedido.",
 			"rateLimited": "Demasiados intentos fallidos. Por favor intentá más tarde."
 		},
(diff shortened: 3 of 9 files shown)


Tirodem:
@slashbinslashnoname tests are ok from my side, please review

slashbinslashnoname:
Issues detected : 

- the promo code leaks in the url, will appear in browser history, server access logs, referrer headers, and any analytics/monitoring tooling. A recipient forwarding the link (e.g. on social media) also inadvertently shares the code. This is inherent to the design, but worth calling out explicitly in the PR so shop owners understand the trade-off before enabling it. My pick is to disable sharing the promocode.

- the timer is not cleaned up on destroy :  no onDestroy(() => clearTimeout(shareCopyTimer)). If the component unmounts during the 2-second window (e.g. navigation), the callback fires against a destroyed component. 

- "none" is applied to error 400 instead of having his own behavior in applyPromoCode action.

- Promo overwite is silent : If the recipient already has a different promo code applied to their cart and accepts the shared cart link, their existing promo code is silently overwritten with the one from the URL. No warning is shown.

Good implementation overall.

Tirodem:
Hmmmm about promocode, either they're set as public and communicated, either it's falsely private, either it's private but in that case when you set a promocode you also set a restriction on email / npub.
I don't feel the promocode leaking is that much of an issue, and is more about the owner setting his/her promocodes properly.
I'm open to (non-lenghty) discussion. @CypherJack also ?

I'm going for the other 3 points.

slashbinslashnoname:
If the perimeter of the issue is only sharing a cart, then sharing promocode alongside should be dismissed, it is out of the perimeter. Only cart items should be shared.

CypherJack:
Sharing a promocode pre-applied is nice but sharing a cart pre-filled and a promo code is just two lines instead of one and a very simple (and somewhat satisfying) action from users.

incrediblelink.com/?mycartissharedwithalink
INCR20


To me we can drop it even if it could have been nice.

Tirodem:
It would have been an opportunity for have shareable link to constitue cart + promocode, promocode being related to a marketing operation (for example DINOVOX2026LAUNCH). That would have acted as a referal + quickbuy feature.
But indeed that could lead to some errors with promocode use (for example if owner create a 100% discount promocode to a partner).
It still can be manually shared, but it reduces the functional risk a bit.
I can go reasonnable in here and remove promocode sharing / application from the feature, and we'll see in the future.

Tirodem:
@slashbinslashnoname for re-review

slashbinslashnoname (approved):
Heavily simplified. All is green

PR #2612, opened 2026-06-19
🐛🤖 Add show/hide toggle to nsec input in /admin/nostr (#2527)

### LINKED ISSUE #2527: 🦋 Display nsec back in /admin/nostr
When using this button :

Image

Nsec is generated.
But with the change of input type :

Image

User cannot see and save his/her nsec.

We should change the display type back to when npub was visible.

Tirodem:
Closes #2527.

## What

The "Generate New Key" button on /admin/nostr writes a freshly generated nsec into the input, then `type="password"` (added by #2263) immediately masks it — the operator can't read or copy the key before saving, and loses it.

## Fix

Keep the masking by default (#2263's intent) and add a 👁️/🙈 toggle button next to the input so the operator can reveal the value when generating a new key, copying an existing one, or auditing what is saved.


Tirodem:
@slashbinslashnoname for review

slashbinslashnoname:
All is green for me.

slashbinslashnoname (approved):

Issue #2432, opened 2026-02-27
🆕 Admin interface reforge (navigation)


Tirodem:
Instead of this :

<img width="3065" height="1645" alt="Image" src="https://github.com/user-attachments/assets/439d4770-990c-4886-95d7-a9e643b1df52" />

Let (maybe) switch to something like this :

<img width="3072" height="1706" alt="Image" src="https://github.com/user-attachments/assets/73bd041d-bd15-4604-8f5c-a9a02c8f7c25" />

?

In the future we'll have more and more and more payment methods / providers / processors, so we'll soon be limited by header width.

CypherJack:
I like vertical navigation. But there is a clear hierarchy problem with the screenshot in your proposition. It is really hard to make the difference between collapsable element (beside the small triangle) and the actual pages links. We would need to put proper UI/UX work on this to improve. Could be the use of icons for categories. Offset alignment for pages links. Stronger typography difference. Or a mix of multiple option.


A proposition to be more efficient for the be-BOP Cloud launch, would be a more narrowed reforge. Use the second nav menu as columns similar to this:
It might be less work but only @gntw can judge this, not me.

<img width="1536" height="1024" alt="Image" src="https://github.com/user-attachments/assets/6d012bab-2cd6-4287-beba-d3f565e5089d" />

.
.
.
> This would also reduce the height of the navigation and make second line only appear when admin wants to navigate between back-office pages

PR #2618, opened 2026-06-20
🆕🤖 Admin left-sidebar nav with search & per-user bookmarks (#2432)

Tirodem:
## Summary
Refactor the back-office navigation per #2432: replace the two-tier header with a left sidebar, add a search box and per-user bookmarks, plus a `runtimeConfig`-driven hook to disable admin entries at the deployment level.

- Left sidebar nav with ant-design icons (Merch, Settings, Payment Settings, Transaction, Node Management, Widgets) — level-2 leaves on click. Single source of truth stays `adminLinks.ts` (now with an `icon` field).
- Search box filters leaves by label (case-insensitive substring); auto-expands matching sections; ignores the bookmark filter.
- Empty/filled star next to each leaf → toggles bookmark via `POST /admin/back-office-bookmark`. Persisted at `users.userSettings.backOfficeBookmarks` (sub-object kept open for future employee-selfcare prefs).
- "Show all" toggle (only shown when the user has ≥ 1 bookmark): off by default → only bookmarked leaves are visible; on → everything. The leaf for the current page is always visible.
- `runtimeConfig.disabledAdminEntries`: list of canonical hrefs (e.g. `['/admin/sumup']`) controlled by the hoster via mongosh today (no in-app UI). The sidebar hides them and `hooks.server.ts` returns 404 for matching routes via the new `isAdminPathDisabled` helper (handles `/admin-<hash>` prefix normalization and sub-paths).
- Bitcoin core / Lightning LND entries continue to be hidden via their existing `isBitcoinConfigured` / `isLndConfigured` checks (unchanged).
- Mobile: burger toggle + responsive viewport meta. Full mobile optimization of the admin pages is deferred to a follow-up story.

## Data model
- `users.userSettings.backOfficeBookmarks?: string[]` — array of canonical admin hrefs starred by the user. `userSettings` is a new sub-object intended to host other employee-selfcare prefs later.
- `runtimeConfig.disabledAdminEntries: string[]` — defaults to `[]`. Edited by the hoster via `mongosh` (e.g. `db.runtimeConfig.updateOne({_id:'disabledAdminEntries'}, {$set:{data:['/admin/sumup']}}, {upsert:true})`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ vite.config.ts.timestamp-*
 static/tinymce
 lib/utils/timeZoneOffset.ts
 docker-compose.override.yml
+
+# Claude scratch / notes (not for commit)
+.bebop-notes.md
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,3 @@ vite.config.ts.timestamp-*
 static/tinymce
 lib/utils/timeZoneOffset.ts
 docker-compose.override.yml
-
-# Claude scratch / notes (not for commit)
-.bebop-notes.md
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -6,6 +6,7 @@ import { addYears } from 'date-fns';
 import { SvelteKitAuth } from '@auth/sveltekit';
 import { flatten } from 'flat';
 import { adminPrefix as _adminPrefix } from '$lib/server/admin';
+import { isAdminPathDisabled } from '$lib/server/admin-disabled';
 import '$lib/server/locks';
 import '$lib/server/sdk/pp-registry';
 import { refreshPromise, runtimeConfig } from '$lib/server/runtime-config';
@@ -274,7 +275,16 @@ const handleGlobal: Handle = async ({ event, resolve }) => {
 			throw error(403, 'Your role does not exist in DB.');
 		}
 
+		if (isAdminPathDisabled(event.url.pathname, runtimeConfig.disabledAdminEntries)) {
+			throw error(404, 'This admin section is disabled on this deployment.');
+		}
+
+		// User-self endpoints (per-user prefs) only need an admin login, no role permission.
+		const normalizedAdminPath = event.url.pathname.replace(/^\/admin-[a-zA-Z0-9]+/, '/admin');
+		const isUserSelfAdminEndpoint = normalizedAdminPath === '/admin/back-office-bookmark';
+
 		if (
+			!isUserSelfAdminEndpoint &&
 			!isAllowedOnPage(
 				event.locals.user.role,
 				event.url.pathname,
--- a/src/lib/server/admin-disabled.test.ts
+++ b/src/lib/server/admin-disabled.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from 'vitest';
+import { isAdminPathDisabled } from './admin-disabled';
+
+describe('isAdminPathDisabled', () => {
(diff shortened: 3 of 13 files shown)


Tirodem:
@slashbinslashnoname for review please
QA is ok from my side but I expect @CypherJack feedbacks and use on it monday

CypherJack:
Great changes overall the fixed layout, the icons and the mobile UX are a nice step up. A few things caught my eye:

## Underline behavior
Categories show underline on hover, nav links use it for active state only. Would be good to make this consistent: underline on hover for everything, and switch active to bold text instead of underlined.

## Link order
The current order looks like it follows implementation order, which doesn't always make sense from a navigation standpoint. Probably worth its own issue to sort out later. I think we should chose between either a better/improved intentional/opinionated grouping or alphabetical. A few examples:

- **Merch:** Pictures and Files should be adjacent; same for CMS, Products and Labels; and Layout next to Themes.
- **Payment Settings:** Bitcoin-related entries grouped together, fiat options together, with Taler and POS Payments at the end since they're more specific.

## Bookmarks
Really nice feature. One thing that would make it much more useful: letting admins reorder their bookmarks after the fact, could be a simple Edit link next to Bookmarks title that display classic be-BOP reorder arrows.

## POS session
Could we also group their inner link /pos/session and /pos-touch ?

Tirodem:
- Underline behavior : will do, nice & quick
- Link order : in a future ticket. I want to keep the current order so far to not confuse existing users. We'll think about this soon (and probably we'll provide more display options on employee selfcare).
- Bookmarks : kinda the same, so far it's from bookmark addition timestamp, in the future, blabla.
- POS Session : I don't want to expose 3 POS Links (4 soon with CTI) because it can confuse web-only users. /pos will be refacto to act as a PoS specific back-office homepage.

slashbinslashnoname:
3 points here for review that triggered me 

- `JSON.parse(await request.text())` :  use `request.json()` to avoid unhandled SyntaxError (back-office-bookmark/+server.ts on line 17)
- Silent bookmark failure :  non-ok fetch responses are silent, no user feedback. 
- `hasPosOptions` redirect removed without mention :  employees with POS options enabled but not the POS role no longer auto-redirect to /pos on login; behavior change not described in the PR
  
  Except point 1 that is needed, all is green

CypherJack:
Side idea: what if the sidebar defaulted to icon-only (with enough padding around each icon to breathe), then expanded on hover to show category names and links?

A manual toggle would control the persistent state: expanded stays expanded, collapsed stays collapsed and only opens on hover. Pretty standard pattern for dense back-offices and would free up a lot of horizontal space.

Not blocking anything, just worth considering for the future.

Tirodem:
@CypherJack : open new :butterfly: ticket for this please.

@slashbinslashnoname : code ready for final reviews.

slashbinslashnoname:
.gitignore : committing personal Claude-scratch ignores into the shared repo. Adding .bebop-notes.md ("Claude scratch / notes (not for commit)") to the project .gitignore pushes one contributor's tooling artifact onto everyone. This belongs in a global/~/.config/git/ignore, not the repo.
 
 I'll wait for the fix and validate the branch. 
 
 I land lately on this issue and discussions, but I would have deleted the top header and top menu for the admin.

Tirodem:
Done for .bebop-notes.md
Won't do topbar / navbar removal for now, because some pilots are used to it.
We'll wait feedbacks after this admin reforge to go further, not unless.

slashbinslashnoname (approved):

PR #2621, opened 2026-06-20
🐛🤖 Stop pagination buttons from sticking to viewport (#2576)

### LINKED ISSUE #2576: 🐛 Next button on product page is sticky
As a side effect of making back-office Save/Update buttons sticky, the **Previous** and **Next** buttons on the product page became sticky too. It shouldn't, it's part of the navigation, not the save system. It also renders with an odd white background, see screenshot below.

Image

Tirodem:
## Summary
Fixes #2576.

On the back-office product list (`/admin/product`) and orders list (`/admin/order`), the **Previous** and **Next** pagination buttons were sticking to the bottom of the viewport with a white background — the same sticky behaviour we use for real "Save" / "Update" action bars. Pagination is navigation, not a save action, so it should scroll normally with the page.

## Behaviour after fix
- `/admin/product` Previous/Next: no longer sticky, scrolls with the page, no stray white background.
- `/admin/order` Previous/Next: same fix, same behaviour.
- Real save/update action bars (e.g. the bottom of `/admin/config`, every edit form) keep their sticky behaviour — only the explicitly-opted-out pagination rows are excluded.

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -238,7 +238,7 @@
 	   e.g. /admin/config/vat) to avoid matching forms nested inside banner/widget components. */
 	:is(main.body-mainPlan:not(.no-sticky-actions), main.body-mainPlan:not(.no-sticky-actions) main)
 		> form
-		> div:has(input[type='submit'], button[type='submit'], button[formaction]) {
+		> div:not(.no-sticky):has(input[type='submit'], button[type='submit'], button[formaction]) {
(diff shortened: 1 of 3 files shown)


Tirodem:
@CypherJack QA'd OK for me.
@slashbinslashnoname for review

slashbinslashnoname (approved):
Approved

PR #2623, opened 2026-06-20
🐛🤖 Graceful payment generation error + PSP test connection (#2546)

### LINKED ISSUE #2546: 🐛 Improve payment generation error management
Currently, with a wrong Stripe API Key, here's what we have :

Image

402 error like this is trust-destroying for end-customer.

It should be covered by redirect to /error, or error display on /checkout like this :

Image

Or this if PoS payment :

Image

Tirodem:
## Summary
Fixes #2546.

When a payment processor refuses or stalls when be-BOP asks it to initiate a payment (bad credentials, transient PSP outage, slow remote endpoint…), be-BOP used to throw a raw 402 page at the customer's face — and on some providers like Swiss Bitcoin Pay, the connection would even hang for minutes before flipping to the generic CMS `/error` page. Trust-destroying and impossible to recover from inside the checkout flow.

### Customer
- A friendly inline banner replaces the error page on every flow that talks to a PSP — e-shop checkout, payment retry on an existing order, subscription renewal, POS split checkout. The customer stays on the page they were on (no redirect) and can try again or change method without losing their cart / order context.
- The banner reads: *"We couldn't initiate your payment. Please try again in a few moments, or contact the shop owner if the problem persists."* Translated in en, fr, de, es-sv, it, nl, pt.
- The **Pay** button on `/checkout` now stays disabled for the **entire** server round-trip (previously re-enabled itself the moment client-side validation finished, allowing accidental duplicate submissions during the seconds the PSP takes to answer).
- Swiss Bitcoin Pay specifically: the connection used to be allowed to hang indefinitely on a bad API key, eventually surfacing a reverse-proxy 5xx HTML page. It now fails fast (15s) and surfaces the same friendly banner as any other PSP failure.

### Staff (on /order/[id])
- The "Add payment" action triggered from the staff view of an order now also surfaces the friendly banner instead of an error page when the PSP refuses. The staff member stays on `/order/[id]` and can retry or pick another method.

### Super-admin
- A new automatic email lands in the shop owner's mailbox each time a payment fails to be initiated, with: the context (checkout / staff add-payment / renewal / POS split / retry), the payment method involved, and a link to the order when one was already created.
- The raw provider error message is deliberately **not** included in the email — several PSPs echo back submitted credentials in their error responses, and emails get forwarded, indexed, shared inboxes happen. The shop owner is told to consult the server logs if they want the gritty detail.

### Admin — new "Test connection" button on every supported PSP page
- On `/admin/stripe`, `/admin/sumup`, `/admin/swiss-bitcoin-pay`, `/admin/paypal` and `/admin/osb`, a new **Test connection** button lets the shop owner validate the saved credentials without waiting for a real customer to hit the broken setup. It triggers a single 1 EUR synthetic payment against the provider; nothing is persisted in be-BOP's own DB, the PSP-side artifact expires on its own.
- Result is displayed inline:
  - **Green banner** on success: credentials work.
  - **Red banner** on failure with a generic *"Connection failed. Please verify the saved credentials and try again."* — same security principle as the email, no raw provider response is shown.
  - **Separate banner** when the provider doesn't answer within 15s ("provider did not respond"), so a network blip is distinguishable from a credentials problem.
- Throttled by a server-side rate limit (5 tries per minute per IP per provider) and a 10s client-side debounce on the button — no spam, no abuse vector against the provider's own rate limits.

## Out of scope
- Distinguishing in the customer banner between *transient* (PSP down, retry in a minute) and *configuration* (bad credentials, contact us) failures. The current message covers both safely. A TODO comment on the new `PaymentGenerationError` class flags this for a follow-up iteration.
- Templatising the super-admin email; for now it follows the same hardcoded-HTML pattern as the existing "NEW DISCOUNT" notification. The comment in code points to PR #1979's broader email-notifications rework which should absorb both.

### DIFF
--- a/src/lib/components/Order/PaymentForm.svelte
+++ b/src/lib/components/Order/PaymentForm.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import type { SerializedOrderPayment } from '$lib/types/Order';
 	import { useI18n } from '$lib/i18n';
+	import { enhance } from '$app/forms';
 
 	const { t } = useI18n();
 
@@ -31,7 +32,7 @@
 			: t('pos.cta.resendPaymentMethod');
 </script>
 
-<form {action} method="post" class="flex flex-col gap-2">
+<form {action} method="post" class="flex flex-col gap-2" use:enhance>
 	<!-- Amount input -->
 	<label>
 		{amountLabel}
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -1789,6 +1789,77 @@ export async function createOrder(
 	return orderId;
 }
 
+/**
+ * Thrown when a payment record cannot be created/initialised against a PSP.
+ * Caller routes catch it and surface a friendly banner instead of a 402 error page.
+ *
+ * TODO(#2546): classify the cause so the UI can vary the wording.
+ * `kind` should become one of:
+ *  - 'transient' (PSP 5xx, timeout, ECONNREFUSED, 429) → "try again later"
+ *  - 'config' (401/403/400 from PSP, invalid credentials, bad amount) → "contact the shop"
+ *  - 'unknown' (anything else, fallback transient)
+ * Done generically for now since the inner errors aren't typed enough to discriminate safely.
+ */
+export class PaymentGenerationError extends Error {
+	readonly method: PaymentMethod;
+	readonly reason: string;
+	constructor(method: PaymentMethod, reason: string, cause?: unknown) {
+		super(`Payment generation failed for method ${method}: ${reason}`);
+		this.name = 'PaymentGenerationError';
+		this.method = method;
+		this.reason = reason;
+		if (cause) {
+			(this as { cause?: unknown }).cause = cause;
+		}
+	}
+}
+
+/**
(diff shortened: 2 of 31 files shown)


Tirodem:
@slashbinslashnoname this one is finally OK in term of QA and I've added some related features on it.
But as it's heavily sensitive (payment) I'd rather consider this pull request as an interactive specification that can be potentially redone from scratch (but you can check it, test it, and see how it's expected to work).
@CypherJack fyi

slashbinslashnoname:
What I would have done naïvely : I would have only validated the api keys as every PSP has a lightweight read-only API call that validates credentials without creating an order. 

For the review, createOrder(items, null, ...) passes null as the payment method, so generatePaymentInfo is never called and PaymentGenerationError can never be thrown there. The first try/catch is dead code, never reached.
```
  try {
      orderId = await createOrder(items, null, { ... });
  } catch (err) {
      if (err instanceof PaymentGenerationError) { ... }  // unreachable
      throw err;
  }
```
Remove the first try/catch. Only the second one (wrapping addOrderPayment) is needed.

slashbinslashnoname (approved):

PR #2629, opened 2026-06-23
🐛 Fix premature expiration of Bitcoin onchain payments

### LINKED ISSUE #2587: 🦋 Prevent premature expiration of Bitcoin onchain orders awaiting confirmation
## Context

When a merchant sets a confirmation threshold ≥ 1, an order transitions `"pending"` → `"paid"` only after the required confirmations are reached.
However, the global expiration timer runs independently: if the TX is in the mempool but blocks are slow (fee pressure, short expiry window), the order can transition `"pending"` → `"expired"` before the threshold is met, even though funds are effectively committed and this could result in a `"paid"` order wrongly marked as `"expired"`.

### The fix
The expiration logic must check for a TX in the mempool before firing. If one is present, the transition to expiring is suppressed waiting for the number of confirmation defined by the be-BOP's threshold. If the TX is no longer in the mempool and does not appear in any block (fees too low, RBF, etc..) and the global timeout has passed, expiration proceeds normally as the order will probably never being paid.

> Note: threshold `0` (unconfirmed accepted) is unaffected the order resolves to `"paid"` on mempool detection before expiration can fire.

## Acceptance criteria

- [ ] A Bitcoin onchain order with a TX in the mempool does not transition to `"expired"`, regardless of the global timeout and keep the pending state until the desired number of confirmation is reached.
- [ ] If the TX is no longer in the mempool, has not been mined in any block and that global timeout has already elapsed, the order transitions to `"expired"`
- [ ] Expiration behaviour for other payment methods is unaffected
- [ ] Retrocompatibility preserved

slashbinslashnoname:
Fixes #2587 

## Problem

Orders were expiring the moment the payment deadline passed — even when a full-amount transaction was already visible in the mempool, just not yet confirmed to the required depth.

## Fixes

Three related changes:

- **`getSatoshiReceivedNodeless`** now returns `pendingSatReceived` for TXs detected but not yet confirmed to the required depth, and removes the upper block-time bound that was dropping legitimately-late-mined TXs.
- **`PPBitcoinNodeless`** uses `pendingSatReceived` to keep an order `pending` (with `awaitingConfirmation: true`) while funds are in flight, and adds a 3-minute grace window when a TX disappears from the mempool post-deadline (rides out RBF replacements).
- **`OrderPayment`** gains `awaitingConfirmation` and `mempoolMissingSince` fields; the order-lock worker persists changes on each poll cycle; the buyer-facing order page shows "received, awaiting confirmation" with a per-language block-count message.

## Tests

Added coverage in `bitcoin-nodeless.test.ts` and `PPBitcoinNodeless.test.ts` for the pending/awaiting-confirmation paths and the mempool-disappearance grace window.

## i18n

Added the awaiting-confirmation strings to all 7 translation files (de, en, es-sv, fr, it, nl, pt).

### DIFF
--- a/src/lib/components/Order/PaymentItem.svelte
+++ b/src/lib/components/Order/PaymentItem.svelte
@@ -55,7 +55,9 @@
 				amount={payment.price.amount}
 				currency={payment.price.currency}
 			/>
-			- {t(`order.paymentStatus.${payment.status}`)}
+			- {payment.status === 'pending' && payment.awaitingConfirmation
+					? t('order.paymentStatus.awaitingConfirmation')
+					: t(`order.paymentStatus.${payment.status}`)}
 		</span>
 	</summary>
 
--- a/src/lib/components/Order/PaymentItem.svelte
+++ b/src/lib/components/Order/PaymentItem.svelte
@@ -56,8 +56,8 @@
 				currency={payment.price.currency}
 			/>
 			- {payment.status === 'pending' && payment.awaitingConfirmation
-					? t('order.paymentStatus.awaitingConfirmation')
-					: t(`order.paymentStatus.${payment.status}`)}
+				? t('order.paymentStatus.awaitingConfirmation')
+				: t(`order.paymentStatus.${payment.status}`)}
 		</span>
 	</summary>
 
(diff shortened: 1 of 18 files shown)


slashbinslashnoname:
QA has been done, feature tested. I'm not fan of the confirmation interface, so happy to have some proposals on this. 

Actually it's a text message confirming the reception and awaiting confirmation, but the qrcode could be replaced by a success message instead

Tirodem:
QA done.
Fakepool is so great and practical :tada: 
Unpaid order expires.
Paid on-chain but < confirmation treshold now waits.

Tirodem (approved):

PR #2630, opened 2026-06-23
🚸 Make linting actually run and enforce typed module boundaries

slashbinslashnoname:
## Summary

Linting was silently doing almost nothing — the `lint` script ran bare `eslint` (linting no files), and a root-only `/*.js` ignore plus missing dot-folder coverage hid source files. This branch makes linting real and types every module boundary so it stays that way.

## Changes

- **Fix the lint script**: `eslint` (linted nothing) → `eslint .`
- **Stop silently skipping source files**: replace the root-only `/*.js` ignore (which exposed the `[lang].js` route dir) and add `.well-known` dot-folder coverage
- **Enable `@typescript-eslint/explicit-module-boundary-types` globally** and type all module boundaries (routes via `./$types`, lib via explicit types)
- **Resolve the svelte-check fallout**:
  - `App.PageData.session` optional (`@auth/sveltekit` augmentation vs `@auth/core`)
  - private-impl + `ReturnType` for self-referential / shape-erased return types (`productBaseSchema`, `fetchOrderForUser`)
  - shared typed `fetchCatalog` for the home page; backup export returns 500 on error
- **Centralize reusable types** (`CurrencyType`, `ThemeMode`, `TagWidgetTag`) and type the `admin_hash` param matcher

## Merge with `main`

This branch was merged up to date with `main`. Because main's recent work was never linted under the now-enforced rules, the merge surfaced fallout that is fixed here:

- **37 `explicit-module-boundary-types` errors** — typed the `migrations` array (29 `run` fns → `Promise<void>`), 4 searchlist route `load` functions, and 2 functions in `checkoutFields`
- **3 svelte-check errors** — restored `PickDeep`/`Tag` imports the merge dropped in `cms.ts`, and narrowed main's new `searchProducts`/`buildSearchlistViews` `locals` params (new `SearchlistLocals` type) so they remain compatible with the narrowed `cmsFromContent` boundary while keeping `language` required

## Verification

- `eslint .` — clean
- `prettier --check` — clean
- `npm run check` (svelte-check) — 0 errors, 0 warnings
- `vitest run --no-threads` — 140/140 pass


slashbinslashnoname:
@Tirodem Merge issues fixed

PR #2243, opened 2025-12-04
🦋 Multi-day bookings for daily rentals

gntw:
Products with daily pricing (24h slots) now support date range selection.
Users pick start/end dates in calendar, and only available days in that range are booked and charged. Unavailable days are skipped automatically.

Example: Select Dec 1-4, but Dec 2 is unavailable → books Dec 1, 3, 4 and charges for 3 days.

Closes #1868

PR #2632, opened 2026-06-24
🆕🤖 Same-day booking control on bookable products (#2590)

### LINKED ISSUE #2590: 🐛 Fix phantom cart add on booking with same-day start or unavailable dates in range
## Overview
When adding a booking product to the cart, the add-to-cart action returns a success response but silently discards the item in (at least) two scenarios:

1. **Same-day start:** reservation starts today; a last-minute booking protection is in place but its rejection is not surfaced to the user.
2. **Unavailable date within range:** the product has blackout dates (ex: closed on weekends, or already booked) but the date picker allows selecting a range that spans one; the item is silently dropped.

In both cases the cart UI shows no error, added item pop-in, number of item in cart not incrementing, and no item really added in, the user has no way to know why the addition failed. Going to checkout just display the empty cart (or other items with the missing booking product if any others are in)

Image

Image

## References:
- #2131
- #2243

## Acceptance criteria

- [ ] Same-day start → explicit error message displayed, item not added
- [ ] Range spanning an unavailable date → explicit error message displayed, item not added
- [ ] In both cases the cart state is unchanged (no phantom entry, no stale state)
- [ ] No regression on valid booking ranges

## Gherkin

```gherkin
Scenario: Same-day start is blocked with user feedback
  Given a bookable product with last-minute protection enabled
  And today is 2026-06-11
  When the user selects a range starting 2026-06-11 and adds to cart
  Then an explicit error message is shown ("unavailable" or "last-minute booking not allowed")
  But no item is added to the cart

Scenario: Range spanning an unavailable date is blocked with user feedback
  Given a bookable product unavailable on weekends
  When the user selects a range from Friday 2026-06-13 to Tuesday 2026-06-17
  Cannot add to cart with an explicit error message is shown listing the unavailable dates
```

Tirodem:
## Summary
Fixes #2590.

### Bug
On all-day bookable products, trying to book today showed the green "added to cart" pop-in but nothing actually landed in the cart — confusing, no error surfaced.

### Behaviour after fix
Same-day bookings become an explicit shop-owner decision on all-day bookable products:

- New admin checkbox **"Allow booking on same day"**, default **off**. When off, today is greyed out in the date picker and can no longer be selected.
- When on, a **"Max hour for same-day booking"** field appears (default **14:00**). The customer can book today up to this hour; past it, today is greyed out again. The cutoff is evaluated in the **shop's timezone**, not the visitor's.
- Both controls only apply to all-day bookings — hourly bookings (coworking rooms, etc.) are unchanged.

If a same-day request still reaches the server (direct API call, stale tab), it is rejected with a translated inline error in the customer's language instead of being silently dropped or shown as a raw error page. New customer-facing and admin labels added in en, fr, de, es-sv, it, nl, pt.

Multi-day ranges that span unavailable days in the middle (introduced in #2243) are unchanged: those days continue to be silently skipped and the customer is only charged for the available days.

### Data model
- `bookingSpec.allowSameDayBooking?: boolean` — default `false`, honoured only when the slot is a full day.
- `bookingSpec.sameDayBookingMaxHour?: string` — `"HH:mm"` in the schedule's timezone, default `"14:00"`, honoured only when `allowSameDayBooking` is `true`.

Both fields are optional; existing bookable products without them behave as `allowSameDayBooking = false` (today not selectable).

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -297,6 +297,11 @@
 			sunday: null
 		}
 	};
+	// Default the same-day cutoff hour whenever the operator turns the option on, so the bound
+	// `<input type="time">` has a value to display from the first reactive cycle.
+	$: if (bookingSpec.allowSameDayBooking && !bookingSpec.sameDayBookingMaxHour) {
+		bookingSpec.sameDayBookingMaxHour = '14:00';
+	}
 	let days = typedFromEntries(
 		dayList.map((day) => [
 			day,
@@ -1163,14 +1168,42 @@
 						<div class="bg-indigo-50 p-4 rounded-lg space-y-4">
 							<label class="form-label">
 								Booking slot duration
-								<select name="bookingSpec.slotMinutes" class="form-input">
+								<select
+									name="bookingSpec.slotMinutes"
+									class="form-input"
+									bind:value={bookingSpec.slotMinutes}
+								>
 									{#each bookingSpecSlotOptions as { value, label }}
-										<option {value} selected={product.bookingSpec?.slotMinutes === value}>
+										<option {value}>
 											{label}
 										</option>
 									{/each}
 								</select>
 							</label>
+
+							{#if bookingSpec.slotMinutes === 60 * 24}
+								<label class="form-label flex-row gap-2 items-center">
+									<input
+										type="checkbox"
+										name="bookingSpec.allowSameDayBooking"
+										bind:checked={bookingSpec.allowSameDayBooking}
+									/>
+									<span>Allow booking on same day</span>
+								</label>
+
+								{#if bookingSpec.allowSameDayBooking}
+									<label class="form-label">
+										Max hour for same-day booking (schedule timezone)
+										<input
+											type="time"
+											name="bookingSpec.sameDayBookingMaxHour"
+											class="form-input"
+											bind:value={bookingSpec.sameDayBookingMaxHour}
+											required
+										/>
+									</label>
+								{/if}
+							{/if}
 
 							<label class="form-label">
 								Timezone
--- a/src/lib/components/ScheduleWidget/ScheduleWidgetCalendar.svelte
+++ b/src/lib/components/ScheduleWidget/ScheduleWidgetCalendar.svelte
@@ -28,7 +28,11 @@
 	>;
 	let className = '';
 	export let isDayDisabled: (date: Date) => boolean = () => false;
-	export let selectedDate = new Date();
+	// Type is nullable so consumers that need a "nothing preselected" state can pass `null`
+	// explicitly (booking product page). Default stays `new Date()` so other consumers — notably
+	// the public ScheduleWidget calendar display — keep their "today preselected" behaviour
+	// without needing to bind the prop.
+	export let selectedDate: Date | null = new Date();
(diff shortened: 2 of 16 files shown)


Tirodem:
@slashbinslashnoname QA's done, for review please

slashbinslashnoname:
Solid .

 A few things worth addressing before merge, most of them trace back to one root cause (same-day logic + timezone handling is duplicated and inconsistent).

1. Centralize the same-day rule (fixes several issues at once)

The "parse HH:mm → cutoff → compare to now" logic is reimplemented in three places with inconsistent timezone bases:
- cart.ts:191-210 (server : all in shop tz ✅)
- +page.svelte:59-79 sameDayBlockedReasonInit (client)
- +page.svelte:263 in computeFreeIntervals (client : but isSameDay(date, now) here is in the browser tz)

The selected day is also sent as browser-local startOfDay(...).toISOString() (+page.svelte:993,117) and reinterpreted server-side in the shop tz (cart.ts:196). That drift is real: a same-day-allowed product (cutoff 23:59) preselected tomorrow with time=...T22:00:00Z in local testing.

Suggestion: extract one pure helper and share it client+server, with all date math in spec.schedule.timezone:
// $lib/types/booking.ts
export function sameDayBookingStatus(
  spec: NonNullable<Product['bookingSpec']>, now: Date
): 'disabled' | 'cutoffPassed' | null
Build bookedDates/time as shop-tz calendar days too, so client and server agree on "today."

2. Shared-component regression

ScheduleWidgetCalendar.svelte:31 changed the default selectedDate from new Date() → null. ScheduleWidget.svelte (public calendar display mode) renders it without the prop, so it now loads with no day selected and an empty event panel. Please pass selectedDate explicitly from the product page rather than changing the shared default.

3. Misleading no-op reactive / cutoff freshness

+page.svelte:98 $: sameDayBlockedReason = sameDayBlockedReasonInit is a reactive alias to a const — a tab open past the cutoff still shows today as bookable until the server rejects it. Either make it reactive off currentTime, or drop the fake reactive and use the const directly.

4. Missing tests

The timezone-sensitive cart logic ships with no unit test. Once #1 extracts the helper, add: same-day off → disabled; on + before cutoff → null; on + after cutoff → cutoffPassed; and a cross-timezone case (visitor tz far from schedule.timezone).

5. Minor

ProductForm.svelte:1196 — sameDayBookingMaxHour uses value= while this PR switched slotMinutes to bind:value (:1169). Binding it is more consistent and allows pre-submit validation.

slashbinslashnoname (approved):
I approve it but would add just a small test to have `toShopTzCalendarDayInstant` and `isSameDayInShopTz` covered !

The two helpers have no test. That's the trickier, DST/day-boundary-sensitive code, and it's exactly where a regression would silently reappear even if code seems good for me.

One case would cover it : 

```
// visitor far ahead of the shop: picking a calendar day must serialize to that
// day's shop-tz midnight, and isSameDayInShopTz must agree on "today".
const tz = 'Pacific/Niue';            // UTC−11
const picked = new Date('2026-06-25T00:00:00+14:00'); // visitor in UTC+14
expect(toShopTzCalendarDayInstant(picked, tz).toISOString())
      .toBe('2026-06-25T11:00:00.000Z'); // 2026-06-25 00:00 Niue
```

PR #2635, opened 2026-06-24
🐛🤖 Ship a real dark palette for the default theme (#2421)

### LINKED ISSUE #2421: 🐛 Default Dark Theme is actually a Light Theme
By default be-BOP ships with a Dark Theme, with the exception that it is configured like a Light Theme. Default should be different like # FFFFFF for text and something dark close to black like 0d1117 without forgeting the links, CTA and everything

Tirodem:
## Summary
Fixes #2421.

### Bug
The "Dark theme" toggle on a fresh be-BOP did nothing visible — same white backgrounds and dark text as light mode.

### Behaviour after fix
Dark mode now ships with a real dark palette out of the box: near-black background (`#0d1117`), white text, slightly lighter surfaces for navbar/cards, brighter blue for links. Shops that already configured a custom theme are unaffected.

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -57,6 +57,49 @@
 	--eventCalendar-color: #000000;
 }
 
+html:root.dark {
+	--header-backgroundColor: #161b22;
+	--header-shopName-color: #ffffff;
+	--header-tab-color: #ffffff;
+	--header-activeTab-textDecoration-color: #ffffff;
+	--navbar-backgroundColor: #1c2128;
+	--navbar-color: #c9d1d9;
+	--navbar-searchInput-backgroundColor: #0d1117;
+	--footer-backgroundColor: #161b22;
+	--footer-color: #8b949e;
+	--cartPreview-backgroundColor: #0d1117;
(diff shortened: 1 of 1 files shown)


Tirodem:
@slashbinslashnoname for review :pray:

slashbinslashnoname:
Thats better like this, but theme is not consistent on forms. Forms are still white.

I'll do a full review of code if you confirm it's intended, but it's worth a fix.

Tirodem:
@slashbinslashnoname I'm merging as it is, we'll refine it later with Leo.

CypherJack:
> Thats better like this, but theme is not consistent on forms. Forms are still white.
> 
> I'll do a full review of code if you confirm it's intended, but it's worth a fix.

Probably the class `form-input` have some hardcoded values, here what I got by inspecting a form:

```
.form-input {
    background-color: rgb(255 255 255 / var(--tw-bg-opacity));
    color: rgb(0 0 0 / var(--tw-text-opacity));
}
```

Issue #2113, opened 2025-10-06
🦋 In Merch > Layout, improve social icon management


Tirodem:
Currently :

<img width="825" height="856" alt="Image" src="https://github.com/user-attachments/assets/dbf0b375-62f8-45e8-abc8-a66e009d7e6b" />

You create a footer icon with a target url and you fill the input with a svg code, like :
```
[...]

PR #2636, opened 2026-06-24
🦋🤖 Layout admin: link translation slots + social icon presets (#2475)

### LINKED ISSUE #2475: 🦋 Bulk localization
TBD
Some content like nav links could have all language showed to the admin and let them translate them all in one localization page instead of each language being manually selected one after the other. Improving productivity.

Tirodem:
## Summary
Fixes #2475 and #2113

### Behaviour after fix

**Link translation** (`/admin/layout/translations`) — for topbar, navbar and footer, the page now pre-creates one input row per link present in the main config, with the original label / href shown as placeholders. Translators see the reference next to an empty input instead of having to add and align rows by hand for each language. Strict 1:1 with the main config — links themselves are still added or removed in `/admin/layout`, this page only translates the existing entries. Partial saves are supported (untranslated rows fall back to the main config on the storefront); a graceful inline banner replaces the previous raw 422 page on validation failure.

**Social icons** (`/admin/layout`) — a new "Preset" dropdown next to each icon row offers ready-to-use SVGs for Discord, Twitch, X, Facebook, YouTube, Mastodon, SoundCloud, Nostr, Instagram, Telegram. Picking a preset fills the SVG textarea (and the name field when empty). The default "Custom" entry keeps the previous paste-your-own-SVG flow available unchanged. No DB / schema change — icons are still stored as `{ name, svg, href }`.


Tirodem:
@slashbinslashnoname QA's mostly done here, your review is welcome.

slashbinslashnoname:
> @slashbinslashnoname QA's mostly done here, your review is welcome.

Two features: (1) `/admin/layout/translations` pre-creates one translation row per main-config link so translators fill blanks; (2) `/admin/layout` adds a "Preset" dropdown that injects ready-made social SVGs. The social-preset half is solid; the translation half has a **save-blocking bug I reproduced**, plus a missing-error-handling gap it exposes.

### 1. Saving a partial translation fails with 422 on full page (annnoying blocker, when I go to previous page, form is empty and current translating work is deleted)
Translating a single link (or just the brand name) and clicking **Save** throws **HTTP 422** on a new blank page:
```
topbarLinks.0.label: Required, topbarLinks.2.label: Required,
footerLinks.0.label: Required, footerLinks.1.label: Required,
navbarLinks.0.label: Required, navbarLinks.1.label: Required, navbarLinks.2.label: Required
```
The page can only be saved if **every** topbar/navbar/footer link has a translated label — which defeats incremental translation. Two distinct defects combine here:

**(a) Root cause — validation rejects untranslated rows.** `translations/+page.svelte` now pre-fills each Url input with the main href as a real value:
```svelte
value={row.isTranslated ? row.href : data.defaultConfig.topbarLinks[i]?.href ?? ''}
```
The action (`translations/+page.server.ts:33`) only records non-empty fields (`if (value) set(...)`), and the schema (`layout-schema.ts:8`) requires both `href` and `label`. With **no post-parse filter**, an untranslated row submits `{ href }` with no `label` → `z.parse()` throws `Required` → whole save fails. This is a regression from pre-filling href (previously both fields were empty and skipped, so partial saves worked).

> **Fix:** mirror the sibling `/admin/layout` action — set unconditionally, then filter after parse:
> ```js
> for (const linkKey of ['topbarLinks','navbarLinks','footerLinks'] as const) {
>   res[linkKey] = res[linkKey]?.filter((item) => item.href && item.label) ?? [];
> }
> ```
> so untranslated rows drop instead of failing. (Or render href as a `placeholder` rather than a `value`, and fall back to the main href on the storefront.)

**(b) UX — no client-side error handling.** The form is a plain `<form method="post">` with **no `use:enhance`**, so any validation failure does a full-page POST onto a raw 422 page with Zod text, losing in-progress edits.

> **Fix:** intercept with `use:enhance`, stay on the page, show errors inline — matching the existing pattern in `product/[id]/+page.svelte`, `checkout/+page.svelte`, `order/[id]/+page.svelte` (`enhance` + inline `errorMessage`; the repo has no global toast store, so follow the inline convention):
> ```svelte
> <form method="post" class="contents" use:enhance={() => {
>   errorMessage = '';
>   return async ({ result }) => {
>     if (result.type === 'failure') { errorMessage = humanizeLinkErrors(result.data); return; }
>     await applyAction(result);
>   };
> }}>
>   {#if errorMessage}<p class="text-red-600" role="alert">{errorMessage}</p>{/if}
> ```
> Prefer human messages ("These links still need a translated label: …") over `topbarLinks.0.label: Required`.

### 2. Lost capability + description contradicts the code
`buildRows` (`translations/+page.svelte:16`) renders exactly `main.length` rows; the comment says *"strict 1:1, no add/delete, extras dropped on save."* The **PR description says the opposite**: *"A trailing empty row stays available to define a translation-only link."* The old code rendered that trailing row; this removes it, and a section with zero main links now renders nothing. Make code + description agree (I'd keep strict 1:1 — a French-only nav link is usually a mistake — but then fix the description and add a hint: "Links are managed in Layout; here you only translate them.").

### 3. Social-preset SVGs are hardcoded `fill="white"` → invisible on light backgrounds
Every preset in `social-icon-presets.ts` is white. On a light header/footer the icons disappear — and this repo just shipped a dark-theme PR (#2635), so backgrounds vary. Use `fill="currentColor"` so icons inherit the surrounding text color and work on any theme.

### 4. No icon preview in the admin row
The operator picks "Discord" and sees raw SVG text in a readonly box — they can't tell it rendered until they save and visit the storefront. A small rendered swatch next to the dropdown closes the loop.

### Minor
- `buildRows` maps `translated[i] ↔ main[i]` positionally; if a saved translation has a different count/order, labels can misalign. Worth a comment.

---

### As a PM
- **1 is the blocker** — the core promise ("translate the links you care about") is exactly what's broken. I'd gate merge on (a) and add a test: saving with one label filled persists that one and drops the rest, no 422.
- **Ship (b) regardless of (a)** — even after the validation fix, any other validation path would still throw a translator onto a raw 422 page. Graceful inline errors are table stakes for an admin form, and the pattern already exists in the repo to copy.
- **`currentColor` for presets (3)** is the highest-impact fix in the social half — a preset that's invisible on half of shops isn't really "ready-to-use."
- **Prefer placeholder-over-value for href** — it both avoids the save bug and communicates intent ("leave blank to reuse the main URL").

src/routes/(app)/+layout.server.ts
```diff
@@ -30,6 +30,25 @@ import {
...
+function mergeLocalizedLinks(
+	translated: ReadonlyArray<{ label: string; href: string }> | undefined,
+	main: ReadonlyArray<{ label: string; href: string }>
+): Array<{ label: string; href: string }> {
+	if (!translated) {
+		return [...main];
+	}
+	return main.map((m, i) => {
+		const t = translated[i];
+		return t && t.label && t.href ? t : m;
```
slashbinslashnoname: Deleting, inserting, or reordering a main link in the Layout admin shifts every index after it and silently re-maps that language's overrides onto the wrong links (or drops them). Editing main links currently do not re-sync existing translations. The robust fix is to key overrides by a stable link id rather than position. Hard without migration. Not fond of this solution but acceptable for a V1.

I would try to use the established way of handling translations . I understand there was no way except position to handle them, but it introduces a second way of handling translations instead of focusing on improving the first one.

It's not "invent a new architecture", it's more "give links an id so the existing pattern applies." Add a stable id (uuid/slug) to each link object, then store and resolve exactly like products/CMS, but *it requires a migration*

```
// translations.fr.config
topbarLinks: { lnk_a1: { label: 'À propos' } }     // href omitted → follows main

// resolve (same spirit as the brandName `||` and product `$ifNull`)
main.map(link => ({ ...link, ...overrides[link.id] }))
```

Sorry for that, I'm available if you want to discuss the functionnality

PR #2637, opened 2026-06-24
🆕🤖 GDPR cookie consent banner gating analytics (#2164)

### LINKED ISSUE #2164: 🆕 Optional GPRD feature
Analytic snippet can introduce non-essential cookies (for example, through Amplitude).
So we need to introduce a GPRD cookie acceptation if (so far) an analytic snipped is setted on be-BOP.
In case of deny, do not load analytic snippet.
This might evolve in the future.

Tirodem:
## Summary
Fixes #2164.

### Behaviour after fix
As soon as the admin saves a non-empty Analytics script snippet in `/admin/config`, visitors see a sticky bottom banner on every page asking them to Accept or Deny. Until they accept, no analytics code is fetched. The banner displays the destination hostnames extracted from the snippet ("data sent to plausible.io, …") and links to `/privacy` when that CMS page exists. Translations in en, fr, de, es-sv, it, nl, pt.

A 🍪 button in the shop footer and on `/identity` lets the visitor change their mind at any time (RGPD requires revocation to be as easy as consent). When the snippet is empty no banner and no 🍪 button are shown.

### Data model
- `cookieConsent` cookie (`accepted` / `denied`), 6 months TTL, SameSite=Lax, Secure on HTTPS. No DB change.

### DIFF
--- a/src/lib/components/CookieConsentBanner.svelte
+++ b/src/lib/components/CookieConsentBanner.svelte
@@ -1,0 +1,62 @@
+<script lang="ts">
+	import { invalidateAll } from '$app/navigation';
+	import { useI18n } from '$lib/i18n';
+	import { cookieConsentVisible } from '$lib/stores/cookieConsentVisible';
+
+	export let hostnames: string[] = [];
+	export let hasPrivacyPage = false;
+
+	const { t } = useI18n();
+	let loading = false;
+
+	async function decide(value: 'accepted' | 'denied') {
+		loading = true;
+		try {
+			await fetch('/cookie-consent', {
+				method: 'POST',
+				headers: { 'Content-Type': 'application/json' },
+				body: JSON.stringify({ value })
+			});
+			cookieConsentVisible.set(false);
+			await invalidateAll();
+		} finally {
+			loading = false;
+		}
+	}
+</script>
+
+<aside
+	class="fixed inset-x-0 bottom-0 z-50 body-secondPlan border-t border-gray-300 shadow-lg print:hidden"
+	role="dialog"
+	aria-live="polite"
+	aria-label={t('cookieConsent.banner.ariaLabel')}
+>
+	<div class="mx-auto max-w-7xl px-4 py-3 flex flex-col gap-2 sm:flex-row sm:items-center">
+		<div class="flex-1 text-sm">
+			{#if hostnames.length > 0}
+				<p>{t('cookieConsent.banner.dataSentTo', { hostnames: hostnames.join(', ') })}</p>
+			{:else}
+				<p>{t('cookieConsent.banner.dataSentToFallback')}</p>
+			{/if}
+			{#if hasPrivacyPage}
+				<a class="body-hyperlink underline text-sm" href="/privacy"
+					>{t('cookieConsent.banner.learnMore')}</a
+				>
+			{/if}
+		</div>
+		<div class="flex gap-2">
+			<button
+				type="button"
+				class="btn body-secondaryCTA"
+				disabled={loading}
+				on:click={() => decide('denied')}>{t('cookieConsent.banner.deny')}</button
+			>
+			<button
+				type="button"
+				class="btn body-mainCTA"
+				disabled={loading}
+				on:click={() => decide('accepted')}>{t('cookieConsent.banner.accept')}</button
+			>
+		</div>
+	</div>
+</aside>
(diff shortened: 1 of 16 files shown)


Tirodem:
@slashbinslashnoname for review please, QA's done

slashbinslashnoname:
A small feature can lead to a big review, that's the case today with my review

Context 

When an admin saves a non-empty analytics snippet, a sticky bottom banner asks visitors to Accept/Deny; the snippet is only injected into <head> server-side when consent is accepted (privacy-by-default, good). A 🍪 button in the footer and on /identity re-opens the banner. Consent is stored in a 6-month cookieConsent cookie. The core gating is correct and i18n coverage is complete, but a few things undercut the feature's own RGPD goal, and one means analytics never starts in the session where consent is granted.

Findings (most severe first but nothing is a crash, it's for performance and compliance)

1. +layout.svelte:48 — analytics never executes in the session where the visitor clicks Accept. After decide('accepted'), the code calls invalidateAll(), which re-renders {@html data.analyticsScriptSnippet} in <svelte:head>. Browsers do not execute <script> inserted via innerHTML/@html — only scripts present at parse time run. So the current page view (the one where consent was given) is never tracked; analytics only loads on the next full navigation/reload. Fix: do a full location.reload() (or goto(..., { invalidateAll: true }) that triggers a document load) after accepting, or inject+execute the script programmatically.

2. CookieConsentBanner.svelte:52-60 — Accept is visually dominant over Deny, a CNIL dark pattern. Deny uses body-secondaryCTA, Accept uses body-mainCTA. CNIL/RGPD requires refusing to be as easy and prominent as accepting — this asymmetry directly contradicts the PR's stated compliance purpose. Give both buttons equal visual weight.

3. +layout.server.ts:33 — per-request Mongo query for the privacy page, site-wide. Whenever analytics is configured, every single page load runs collections.cmsPages.findOne({_id:'privacy'}). The root layout runs on all routes. Cache this (a runtimeConfig flag updated when CMS pages change, like the existing config pattern) instead of a round-trip on the hot path. Since privacy is a default seeded page that effectively always exists, the honest simplest answer is to drop the existence check entirely and always link to /privacy — the route renders the CMS page (or its empty default) anyway. You trade a theoretical dead link on a misconfigured install for zero complexity.

5. +layout.svelte:54 (product/scope) — the banner renders on /admin and /pos too. The root layout wraps the back office, so the shop owner and cashiers get a cookie consent banner over their admin/POS UI. Likely should be storefront-only. I've already mentionned the fact that (in #2618), the admin should not have the header and top menu, it's worth it to think about having a separated layout.svelte for admin routes.

6. analytics-hostnames.ts:13 — regex misses protocol-relative srcs (//plausible.io/script.js). Vendor snippets commonly use //host/...; those won't match https?://, so the banner falls back to the generic "uses analytics tracking" text even though the destination is known — defeating the transparency the banner is for.

7. **cookie-consent/+server.ts:6 — the 'reset' branch is dead code.** No client ever POST value:'reset'; the 🍪 buttons only flip the client cookieConsentVisiblestore. Either wire a real "reset/forget my choice" action (arguably a GDPR nicety) or drop the enum value and clear CookieConsent`.

8. CookieConsentBanner.svelte:27 (minor) — accessibility :  role="dialog" with no focus management. We all remember a bad experience we had with cookie banners. It's worth doing this pretty well and give the best for accessibility. Keyboard/screen-reader users aren't moved into the dialog and there's no Escape handling; aria-live="polite" on a dialog is also unusual. Either add focus handling or drop the dialog role in favor of a plain landmark.

Tirodem:
@slashbinslashnoname is done

slashbinslashnoname:
A few retours (I know it's an urgent feature)

1. isBackoffice prefix check suppresses the banner on public CMS pages — src/routes/+layout.svelte:23

$page.url.pathname.startsWith('/admin') || startsWith('/pos') also matches storefront CMS slugs like /positions, /possibilities, /administration.
Failure: A shop with a CMS page slugged /positions — a first-time visitor (analyticsConsent === null) landing there never sees the banner and can't consent, on a fully public page. Use exact/segment matching (=== '/admin' || startsWith('/admin/'), same for /pos) rather than a bare prefix.

2. the  location.reload() is so 2011, can we just inject dynamically the js when accepted ?

Tirodem:
@slashbinslashnoname redone

slashbinslashnoname (approved):

Issue #2646, opened 2026-06-27
👨‍🔬 Auto API call on paid order if a particular product is on cart


Tirodem:
For testing purpose

On /admin/product/new and /admin/product/{id}, add checkbox "this will trigger an API call once order is paid".
When checked, allow to set POST parameters (api route, secret).
When order with this product is paid, make API call with :
- order number
- order notification email / npub
- checkout custom fields

PR #2647, opened 2026-06-27
🆕🤖 Per-product paid-order webhook (#2646)

Tirodem:
## Summary
PoC for #2646.

### Behaviour after fix
A new optional checkbox on `/admin/product/new` and `/admin/product/{id}` — *"This will trigger an API call once order is paid"*. When ticked, the operator fills in an API route URL and a shared secret; once an order containing this product transitions to paid, be-BOP fires a single signed POST to the configured route. Each webhook-enabled product in the order gets its own call. PoC scope: fire-and-forget, no retry.

The receiver verifies the request by recomputing `HMAC-SHA256(secret, raw body)` and comparing in constant time against `X-Webhook-Signature: sha256=<hex>`.

Payload sample:
```json
{
  "timestamp": "2026-06-27T12:34:56.789Z",
  "orderNumber": 12345,
  "contact": { "email": "buyer@example.com", "npub": null },
  "customCheckoutFields": [
    { "slug": "company-name", "label": "Company name", "value": "Acme Inc." },
    { "slug": "billing-address", "label": "Billing address", "address": { "firstName": "Jane", "lastName": "Doe", "address": "12 rue de la Paix", "zip": "75002", "city": "Paris", "country": "FR", "state": null, "companyName": null, "vatNumber": null, "phone": null } }
  ]
}
```

### Data model
- `Product.paidOrderWebhook?: { apiRoute: string; secret: string }` — optional; absent / removed when the operator unticks the checkbox.

### DIFF
--- a/.env
+++ b/.env
@@ -7,6 +7,8 @@ ALLOW_LND_RPC=
 ALLOW_DUMPING_WALLET=
 #RISKY! Set to "true" to enable JS injection in the admin in CMS pages
 ALLOW_JS_INJECTION=
+# Set to "true" to enable the per-product outbound webhook fired when an order transitions to paid
+ALLOW_PAID_ORDER_WEBHOOK=
 # Specify in .env.local to enable bitcoin on-chain payments
 # example : http://127.0.0.1:8332
 BITCOIN_RPC_URL=
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -2000,7 +2000,8 @@
 										autocomplete="off"
 										name="paidOrderWebhook.secret"
 										class="form-input"
-										placeholder="Used by the receiver to verify the HMAC-SHA256 signature"
+										minlength={16}
+										placeholder="HMAC-SHA256 key used by the receiver to verify the signature (min. 16 chars)"
 										bind:value={paidOrderWebhookSecret}
 										required
 									/>
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -1996,7 +1996,8 @@
 								<label class="form-label">
 									Shared secret
 									<input
-										type="text"
+										type="password"
+										autocomplete="off"
 										name="paidOrderWebhook.secret"
 										class="form-input"
 										placeholder="Used by the receiver to verify the HMAC-SHA256 signature"
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -64,6 +64,7 @@
 	export let productsWithStock: { _id: string; name: string }[] = [];
 	export let currentBookings: BookingSummary[] = [];
 	export let upcomingBookings: BookingSummary[] = [];
+	export let allowPaidOrderWebhook = false;
 
 	export let product: WithId<PojoObject<Product>> = {
 		_id: '',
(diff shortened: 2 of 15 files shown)


Tirodem:
PoC is working but I'm gonna hide this feature behind a feature flag as it's more about internal use (a fully public-open feature would require more refinement).

Tirodem:
@slashbinslashnoname for review, call me if infos are needed
(mandatory for be-BOP SaaS launch)

Tirodem:
Hmmm, I need to check behavior when subscription is renewed.

slashbinslashnoname:
Ok I have some reviews to make.

Context : Adds an optional {apiRoute, secret} webhook to a Product. When an order containing that product transitions to paid, be-BOP fires a fire-and-forget HMAC-SHA256-signed POST to the configured URL. Gated behind a new ALLOW_PAID_ORDER_WEBHOOK env flag, admin-only, with UI in ProductForm.svelte. Explicitly scoped as a PoC (no retry).

1. Very high: Shared secret leaks to the buyer (and to plaintext DB) — src/lib/server/orders.ts:1426

items.map(... product: item.product ...) snapshots the entire product — now including paidOrderWebhook.secret — into every order document. order/[id]/+page.server.ts:100 returns the full order to the customer's browser, and nothing strips the field anywhere.
Failure: A buyer who purchased a webhook-enabled product opens their own order page, reads the HMAC secret from the serialized page data, and can then forge valid X-Webhook-Signature requests to the receiver — defeating the entire signature scheme. The secret also sits in plaintext in Mongo. 

Fix: strip paidOrderWebhook before snapshotting into order.items[].product (the fire path re-queries the live product anyway, so nothing breaks). Also verify the behavior client side.

2. Webhook fires twice for free orders — src/lib/server/orders.ts:168 (guard) / :456 (fire)

wasPaidBefore = order.status === 'paid' never sees the transition: onOrderPayment rebinds its local order = ret.value (:403) but never mutates the caller's in-memory order.status. Free orders call onOrderPayment twice (the existing comment at :162 documents this; addOrderPayment at :2103 and createOrder at :1800).
Failure: For every free/fully-discounted order, wasPaidBefore is false on both calls and updated.status === 'paid' on both → the POST is sent twice, contradicting the "only on the first transition" comment. The adjacent alreadyPaid = !!payment.paidAt guard (:165) is the correct signal to reuse.

3. Unhandled rejection can crash the server — src/lib/server/orders.ts:456 / order-paid-webhook.ts:~34

void firePaidOrderWebhooks(updated) has no .catch(), and the collections.products.find(...).toArray() inside sits outside the per-target try/catch (only fetch is wrapped).
Failure: A transient Mongo hiccup (failover/socket reset) right after the payment commits rejects the query → unhandled promise rejection → process termination on modern Node. The payment is already committed, so the crash is pure collateral. Wrap the whole body or add .catch() at the call site.

4. Secret rendered in cleartext — src/lib/components/ProductForm.svelte:~1843

type="text" pre-filled from product.paidOrderWebhook.secret, so it's shipped to the admin browser and shown on screen (shoulder-surf/screen-share exposure). At minimum use type="password"; ideally write-only.

slashbinslashnoname (approved):

Tirodem:
Free product doesn't trigger API call :bug:

slashbinslashnoname (approved):

PR #2655, opened 2026-07-03
🐛 Store order VAT in configured currencies, not SAT (#2492)

### LINKED ISSUE #2492: 🐛 VAT is stocked in SATs in database whatever the main or billing currency is
https://dev-bootik2.pvh-labs.ch/admin-tirodem/order/ede831d6-3c84-4203-ac24-7485438957ce/json

this should not be in SATs :

Image

here are the order informations for comparison :

Image

slashbinslashnoname:
## Summary

Fixes #2492 — VAT was stored in the internal **SAT** unit on the order document, regardless of the shop's main/billing/accounting currency.

Root cause: pricing is computed in `UNDERLYING_CURRENCY = 'SAT'`, and `createOrder` stored that raw array straight into `order.vat`. The correctly-converted per-currency amounts were already sitting in `currencySnapshot.*.vat` — only the top-level `order.vat` (and the accounting log that read it) were wrong.

## Behaviour after fix

- **`order.vat[]` now holds only the `{ rate, country }` breakdown** — no monetary amount, never SAT. The amounts live where every other order amount lives: `currencySnapshot.{main,priceReference,secondary,accounting}.vat[]` (index-aligned with the rate breakdown), each in its real currency.
- **Accounting logs (`paymentDone` / `paymentCall`) now snapshot VAT — and `totalPrice` / `discountAmount` — in every configured currency** via a small pure helper (`orderCurrencyAmounts`), instead of a single main-currency (or SAT) value. `main`/`priceReference` always; `secondary`/`accounting` only when the shop configured them.
- A **migration** rewrites historical orders: it strips the legacy SAT `price`/`partialPrice` from `order.vat[]`. No exchange-rate lookup needed — each order's `currencySnapshot` already froze the correct per-currency amounts at order time. Past accounting-log rows are left immutable (audit records).

Display was already correct (`OrderSummary` / receipt read `currencySnapshot.main.vat`), so there's no UI change.

## Before / after — `/admin/order/{id}/json`

**Before** — `order.vat` amounts in SAT:

```json
"vat": [
  { "country": "CH", "rate": 8.1,
    "price":        { "amount": 853, "currency": "SAT" },
    "partialPrice": { "amount": 853, "currency": "SAT" } }
]
```

**After** — rate breakdown only; amounts per configured currency in `currencySnapshot`:

```json
"vat": [ { "rate": 8.1, "country": "CH" } ],
"currencySnapshot": {
  "main":           { "totalPrice": { "amount": 7.2415, "currency": "EUR", "precision": 4 }, "vat": [ { "amount": 0.54, "currency": "EUR" } ] },
  "priceReference": { "totalPrice": { "amount": 6.66,   "currency": "CHF", "precision": 4 }, "vat": [ { "amount": 0.5,  "currency": "CHF" } ] },
  "secondary":      { "totalPrice": { "amount": 0.00012394, "currency": "BTC", "precision": 8 }, "vat": [ { "amount": 0.0000089, "currency": "BTC" } ] },
  "accounting":     { "totalPrice": { "amount": 7.2415, "currency": "EUR", "precision": 4 }, "vat": [ { "amount": 0.54, "currency": "EUR" } ] }
}
```

## Data model

- `Order.vat?: Array<{ rate: number; country: CountryAlpha2 }>` — was `{ price: Price; rate; country }`.
- Accounting-log `after.vat` / `after.totalPrice` / `after.discountAmount` change from a single `Price` to a per-currency object `{ main, priceReference, secondary?, accounting? }` (the log's `after` is free-form; nothing parses these fields — the CSV/JSON exports just `JSON.stringify` the record).

## Tests

- Pure unit tests for the helpers (`orderVatSnapshot.test.ts`): VAT across every configured currency, `secondary`/`accounting` omission, the empty-rates edge, and `orderCurrencyAmounts` projection.
- A regression assertion in the DB-backed `order.test.ts` (order.vat never carries an amount; `currencySnapshot.main.vat` is in `mainCurrency`, never SAT).
- A **live migration test** (`migrations.test.ts`) run against MongoDB: a legacy SAT-shaped order is reduced to `{ rate, country }` with the per-currency amounts preserved, and it's idempotent.

Verified locally against a live MongoDB (`npm run check`, eslint, prettier, and the above tests all green); the before/after JSON above is the actual local output for the reported order id.

### DIFF
--- a/src/lib/server/migrations.test.ts
+++ b/src/lib/server/migrations.test.ts
@@ -1,0 +1,76 @@
+import { describe, it, expect } from 'vitest';
+import { ObjectId } from 'mongodb';
+import { cleanDb } from './test-utils';
+import { collections, withTransaction } from './database';
+import { migrations } from './migrations';
+
+function migration2492() {
+	const migration = migrations.find((m) => m._id.equals(new ObjectId('6b1f4880e92e590e85af2492')));
+	if (!migration) {
+		throw new Error('migration #2492 not found');
+	}
+	return migration;
+}
+
+describe('migration #2492 — drop single-currency (SAT) amount from order.vat', () => {
+	it('reduces legacy SAT vat entries to {rate, country} against a live database', async () => {
+		await cleanDb();
+
+		const _id = 'legacy-order-2492';
+		// Legacy order: VAT amounts wrongly stored in the internal SAT unit (the bug), while the
+		// correct per-currency amounts are already frozen in currencySnapshot.
+		await collections.orders.insertOne({
+			_id,
+			vat: [
+				{
+					price: { amount: 853, currency: 'SAT' },
+					partialPrice: { amount: 853, currency: 'SAT' },
+					rate: 8.1,
+					country: 'CH'
+				}
+			],
+			currencySnapshot: {
+				main: {
+					totalPrice: { amount: 7.24, currency: 'EUR' },
+					vat: [{ amount: 0.54, currency: 'EUR' }]
+				},
+				priceReference: {
+					totalPrice: { amount: 6.66, currency: 'CHF' },
+					vat: [{ amount: 0.5, currency: 'CHF' }]
+				}
+			}
+		} as never);
+
+		await withTransaction((session) => migration2492().run(session));
+
+		const migrated = await collections.orders.findOne({ _id });
+		// Amount/currency stripped; only the rate breakdown remains.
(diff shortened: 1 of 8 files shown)


Tirodem:
OK on first tests :heavy_check_mark: 
I'll dig more monday.

Tirodem:
Smöl feedback on Zulip because I just have this :
`    vat: [ { rate: 8.1, country: 'CH' } ]`

Tirodem (approved):

PR #2656, opened 2026-07-03
🆕 Custom manual payment methods with configurable label + instructions (#2259)

### LINKED ISSUE #2259: 🆕 Create an asynchronous payment method with manual validation
Similar to Bank Transfer 🏦, but generic with configurable text and label.

## 🧠 Small reflexion 
~~Should we then make Bank Transfer as one of these but created by default, but it would make it editable I guess ?~~

Obsolete thought because of #1147

slashbinslashnoname:
## Summary

Closes #2259.

Adds **custom manual / asynchronous payment methods** — modelled on Bank Transfer 🏦 but generic and admin-configurable. A shop can define **any number** of them, each with a **label** and free-text **instructions**. The customer picks one at checkout, sees the instructions (no QR code), and an admin **validates the payment manually** from the order — e.g. receiving a coin, an off-platform transfer, cash on pickup, etc.

## Admin configuration — `/admin/custom-payments`

A dedicated page under **Payment Settings** (next to Stripe, SumUp, …):

- Add / edit / remove / **reorder** methods, each with a label + instructions.
- **Inline delete confirmation** (no browser popup); **deletion persists immediately** (no need to click Save).
- Empty rows are never saved; a row with instructions but no label is rejected with a clear error.
- The `'custom'` payment method becomes available at checkout as soon as ≥1 method is configured.

## Checkout & orders

- The checkout selector **expands** into one option per configured method.
- Each order payment stores `method: 'custom'` plus a **snapshot** of the chosen method (`{ id, label, instructions }`), taken at order time — so later edits/removals never change what an existing order shows (same robustness as Bank Transfer's `sellerIdentity` snapshot).
- **Manual "Mark paid"** is enabled for `custom` (like Bank Transfer / POS), with an optional free-text reference.
- **Customer order page** shows the chosen method's label + instructions, no QR.
- **Pending email** uses a new editable `order.payment.pending.custom` template with the snapshotted instructions.
- **Admin orders list** and **reporting** show the method's specific label (e.g. "Monero"), not a generic "Custom payment".
- **Discounts** can be scoped to the custom method.

## Data model

- `runtimeConfig.customPaymentMethods: Array<{ id, label, instructions }>`
- `Order['payments'][number].customPaymentMethod?: { id, label, instructions }` (snapshot)

## i18n

All new strings are translated in **en, fr, de, es-sv, it, nl, pt** (customer-facing labels + the admin config UI).

## Notes / scope

- Discounts and reporting group all custom methods under the single `custom` bucket (per-method breakdown in reporting shows the label).
- Verified end-to-end against a live MongoDB (checkout availability → pending order → manual mark-paid; order page, orders list, and pending email all render the chosen method).


Tirodem:
image
Custom payment method name should be displayed here.
This label as it is should be displayed only for deleted payment method.

Also I think it would be a good idea to log the custom payment method subtype on the sales logs because `"method": "custom"` seems not enough to me :

```
  {
    "_id": "6a47d4bd960d5753496bb823",
    "eventType": "paymentCall",
    "before": null,
    "after": {
      "method": "custom",
      "paymentId": "6a47d4bd960d5753496bb822",
      "vat": [
        {
          "country": "CH",
          "rate": 0,
          "price": {
            "amount": 0,
            "currency": "CHF"
          },
          "partialPrice": {
            "amount": 0,
            "currency": "CHF"
          }
        },
        {
          "country": "CH",
          "rate": 8.1,
          "price": {
            "amount": 0,
            "currency": "CHF"
          },
          "partialPrice": {
            "amount": 0,
            "currency": "CHF"
          }
        }
      ],
      "totalPrice": {
        "amount": 10,
        "currency": "CHF"
      }
    },
    "objectId": "c2c3b61a-7d13-4284-93ca-51d9f51d8117",
    "objectType": "order",
    "createdAt": "2026-07-03T15:26:53.948Z"
  }
```

Otherwise, it works.

slashbinslashnoname:
### Follow-up: show the method name & log the subtype

Two rounds of fixes pushed on top of the initial custom-payments work:

**1. Display the actual custom method name (`327d7ddf`)**
- The **receipt** (on-screen *and* the print/PDF version) and the **order summary** now show the custom method's own label via the per-payment snapshot, instead of the generic *"custom payment method"* text.
- That generic label is now only a **fallback** — shown when no label was snapshotted (deleted / legacy method).
- Added a `customNamed` key (`"Paid with: {label}"`) to `paidWith` + `paidWithSummary` in all 7 languages.

**2. Log the custom subtype in accounting/sales logs (`327d7ddf`)**
- `"method": "custom"` alone didn't say *which* custom method was used. The `paymentCall` and `paymentDone` events now also carry `customPaymentMethod: { id, label }`:
  ```json
  "after": {
    "method": "custom",
    "customPaymentMethod": { "id": "…", "label": "…" },
    "paymentId": "…"
  }
  ```
  (Instructions are intentionally omitted — too verbose for accounting.)

**3. i18n gap fix (`6f4809ee`)**
- `paidWithSummary` was **missing the `paypal` key in every language**, so a PayPal payment rendered the raw key `order.paidWithSummary.paypal` in the summary. Added it in all 7 langs, normalized `paypal`/`Paypal` → **`PayPal`**, and fixed a few incidental string bugs (fr colon spacing, it double space, a missing closing paren in it's bitcoin string).

Verified: `svelte-check` clean, all 7 translation files parse & pass prettier, and both `paidWith`/`paidWithSummary` blocks now have complete matching key sets across all languages.

Tirodem:
image

Tirodem (approved):

Issue #2668, opened 2026-07-06
🌐 Translate email templates (Subject/HTML Body) per recipient locale


slashbinslashnoname:
## Context

While auditing admin UI translations for #878, found that `/admin/template/emails`'s chrome (labels, buttons, hints) is correctly translated, but the actual email template content (Subject + HTML Body for all ~21 templates: passwordReset, order.payment.*, subscription.*, schedule.*, etc.) is English-only with no way to translate it.

## Current state

- Templates live in `runtimeConfig.emailTemplates` (see `src/lib/server/runtime-config.ts`), not in a DB collection with a `translations` field like `CMSPage`.
- `queueEmail()` in `src/lib/server/email.ts` always sends `runtimeConfig.emailTemplates[templateKey]` (or the default) with no locale parameter — there's no per-recipient locale resolution at all today.

## Scope for a fix

- Add a `translations` field per template (subject/htmlContent per locale), mirroring the `CMSPage.translations` pattern.
- Admin UI: a per-locale editor on `/admin/template/emails` (similar to the CMS page translations sub-tab).
- Migration to seed default translations for the ~21 existing templates across the 6 non-English locales (careful, customer-facing legal/transactional copy — needs real care, not placeholder text).
- Wire recipient-locale resolution into every `queueEmail()` call site (order confirmations, password reset, subscription reminders, etc.) — requires figuring out which locale to use per recipient, which isn't consistently tracked today.

Flagged during the #878 admin i18n work; deliberately out of scope for that branch since it's a real feature, not a quick translation fix.

PR #2669, opened 2026-07-06
🌐 Migrate admin backoffice to i18n

### LINKED ISSUE #878: 🦋 Migrate admin UI on i18n

slashbinslashnoname:
## Summary

Migrates the be-BOP admin backoffice from hardcoded English to the `t()` i18n system across all 7 supported locales (en, fr, de, es-sv, it, nl, pt). Before this branch, only ~a dozen `admin.*` translation keys existed and almost none of the ~130 admin route/component files used `useI18n()`.

Fixes #878

## Protocol / convention followed

- **Namespace**: `admin.<area>.<element>`, `area` matching the route/component slug (`product`, `order`, `config`, `stripe`, `theme`, …). Existing shared keys reused where they fit (e.g. `admin.action.save/update/reset`).
- **All 7 locales, always**: every migrated key ships translated into `en/fr/de/es-sv/it/nl/pt` in the same PR — `en` is the source of truth (existing English string, unchanged for English users).
- **Brand/technical terms stay literal**: SumUp, Stripe, NostR, BTCPay, PhoenixD, Taler, OSB, S3, OAuth, VAT, POS, CMS, font names, currency/country codes, etc. are never translated.
- **Nav** uses a derived-key scheme (`admin/+layout.svelte`'s `navKey()`/`navLabel()`/`navSection()`): slugifies the English source string → `admin.nav.link.<slug>` / `admin.nav.section.<slug>`, falls back to the English source if no translation exists. Same pattern reused for the theme editor's data-driven field labels (`ThemeForm.svelte`).
- **Default CMS pages** (home, terms, privacy, etc.) are seeded by server code, not templates — added a `translations` field + a migration to backfill existing untouched default pages, guarded so shop-customized content is never overwritten.

## Process

1. Foundation slice (nav keys, coverage gate script, 2 demo pages) landed first and reviewed.
2. Bulk migration via parallel agents (one per admin area/shared component), each returning structured translation data merged centrally into the 7 locale JSONs.
3. Live QA pass in French across the admin UI, fixing gaps as found (including a few cases where agents had falsely reported success, or a `formatDistance()` call was missing its locale so durations rendered in English regardless of surrounding translated text).
4. A final systematic static audit (every admin `.svelte` file read in full, checked against the brand/technical exclusion list) caught the remaining stragglers — including an entire subscription pricing-phases feature added mid-migration that had never been wrapped in `t()`.

## Verification

- `pnpm run check:translations` — coverage gate ensures every `admin.*` key exists in all 7 locales (not wired into CI in this PR — pushing the workflow-file change requires the `workflow` OAuth scope; the npm script is available to wire up separately).
- `svelte-check`, `eslint`, `prettier --check` all clean.
- Manual QA in French confirmed working end-to-end; a couple of pre-existing bugs unrelated to i18n were found and reported separately (not fixed here): a dead `/admin/template/nostr` nav link (issue predates this branch), and email template content (Subject/HTML Body) which isn't translatable today by design — filed as #2668.

## Not in scope

- `feat/admin-redesign` touches the same shell (`+layout.svelte`) — will need a rebase against whichever lands second.
- Email template translation (#2668).


slashbinslashnoname:
@Tirodem

CypherJack:
My testing have been made on the french locale as it's my mother tongue. I have checked on german locale and it seems consistent but cannot double check translation quality. Some inconsistencies between navbar naming and page naming seems to not be present in german as english wording is used or fully translated in some cases.


## In /admin/label

Inconsistence with label/étiquette
image


## In /admind/pos

Not used is not translated

image

image

Missing translation here too

image

image

### In /admin/discount/new

PdV is used instead of PoS

image



## In Payment settings

Inconsistence entre nodeless et sans nœud entre la nav et le titre de la page.
image

image

Inconsistence PoS -> PdV + Description default cash in english only

image

Probably because i18n in backoffice is not in main but:
image

## In widgets

Translated CMS macro resulting in not-loaded widgets
image
image

Inconsistence Tags/Etiquettes

image

Inconsistence Carousel/Slider

image

Inconsistence Agenda/Horaire

image

slashbinslashnoname:
A lot of english hardcoded missing translations, i18n wasn't even imported, that's why our scans missed some even with tens of parralels agents that correctly patched a large number of hardcoded items.

Fix in progress 😊

PR #2671, opened 2026-07-08
🐛🤖 Subscription lock: skip Nostr cursors for subs with empty/null npub (#2670)

### LINKED ISSUE #2670: Stop Nostr subscription notification attempts when Nostr is not collected / no nsec is set
## Problem

The subscription lock (`src/lib/server/locks/subscription-lock.ts`) attempts to send Nostr reminders and end-of-subscription notifications **unconditionally**, even when:

1. **No Nostr private key is configured** on the shop (`runtimeConfig.nostr.privateKey` empty → `isNostrConfigured()` returns `false`). The shop owner has explicitly opted out of Nostr, yet the loop still runs the Nostr path each tick.

2. **The subscription's user has no valid npub** but the Mongo filter lets them through. The filter `'user.npub': { $exists: true }` accepts falsy values (`""`, `null`). When a user has only an email and their `user.npub` is an empty string, the subscription is picked up by the Nostr cursor, then rejected inside `notifySubscriptionReminderViaNostr` / `notifySubscriptionEndedViaNostr`.

Result: log spam like

    Cannot send subscription reminder via Nostr: subscription <id> - user has no npub
    Cannot send subscription ended notification via Nostr: subscription <id> - user has no npub

repeated at every tick of `maintainLock`, for every affected subscription, on Nostr-disabled shops.

## Where

- `src/lib/server/locks/subscription-lock.ts`
  - `maintainLock()` (~L261) — loops over `getSubscriptionsToRemindViaNostr` and `getSubscriptionsToNotifyEndViaNostr` unconditionally.
  - `notifySubscriptionReminderViaNostr()` L87-94 — logs the message.
  - `notifySubscriptionEndedViaNostr()` L179-190 — logs the sibling message.
  - `getSubscriptionsToRemindViaNostr()` L38-48 and `getSubscriptionsToNotifyEndViaNostr()` L65-74 — filters that let empty-string npubs through.
- `src/lib/server/nostr.ts` L15 — `isNostrConfigured()` already exists.

## Expected

1. When Nostr is not configured on the shop, `maintainLock()` should not enter the Nostr blocks at all. Gate `subscriptionsToRemindNostr` and `subscriptionsToNotifyEndNostr` behind `isNostrConfigured()`.

2. When a subscription's user has an empty/null npub (e.g. email-only subscriber), it should not be returned by the Nostr cursors. Harden the Mongo filter (e.g. `'user.npub': { $exists: true, $nin: ['', null] }` or `$type: 'string'` + `$ne: ''`).

Either fix alone kills the log spam in the reported case (Nostr disabled). Both together also prevent the spam on shops that do have Nostr enabled but store some email-only subscribers.

## Not in scope

- Deciding whether email-only subscribers should be reminded via email fallback when Nostr is disabled — the email path (`notifySubscriptionReminderViaEmail`) already exists and runs independently.

Tirodem:
## Summary

Fixes #2670. Two Mongo cursors in `subscription-lock.ts` treated `'user.npub': { $exists: true }` as "user has a Nostr identity", but empty-string / null values passed the filter and reached `notifySubscription{Reminder,Ended}ViaNostr`, which then log-spammed `user has no npub` at every tick of `maintainLock`. Tightened both filters with `$nin: ['', null]`.

Out of scope of this PR (kept for a follow-up if wanted): gating the whole Nostr branch of `maintainLock` behind `isNostrConfigured()` — the filter tightening alone kills the reported log spam.

### DIFF
--- a/src/lib/server/locks/subscription-lock.ts
+++ b/src/lib/server/locks/subscription-lock.ts
@@ -46,7 +46,7 @@ export function getSubscriptionsToRemindViaNostr(
 			$lt: reminderWindow
 		},
 		cancelledAt: { $exists: false },
-		'user.npub': { $exists: true },
+		'user.npub': { $exists: true, $nin: ['', null] },
 		notifications: { $not: { $elemMatch: { type: 'reminder', medium: 'nostr' } } }
 	});
 }
@@ -72,7 +72,7 @@ export function getSubscriptionsToNotifyEndViaNostr(now: Date): FindCursor<PaidS
 			$lt: now
 		},
 		cancelledAt: { $exists: false },
-		'user.npub': { $exists: true },
+		'user.npub': { $exists: true, $nin: ['', null] },
 		notifications: { $not: { $elemMatch: { type: 'expiration', medium: 'nostr' } } }
 	});
 }


slashbinslashnoname (approved):

slashbinslashnoname:
Safe, tigheting should fix the issue

PR #2675, opened 2026-07-13
Fix get() crash on missing intermediate i18n key (breaks en fallback)

slashbinslashnoname:
## Summary
- `get()` crashed instead of returning `undefined` when an intermediate path segment was missing (e.g. `de.json` has `product` but not `product.pricingSchedule`), throwing on `undefined['intro']`.
- This crash happened *before* `t()`'s `?? get(data.en, key)` fallback in `src/lib/i18n.ts:97` could run, so partially-translated locales (like `de`) crashed instead of falling back to English as intended.
- Fix: the `reduce` now short-circuits to `undefined` as soon as the accumulator is `null`/`undefined`, letting the existing `??` fallback in `t()` do its job.

## Test plan
- [x] `tsc --noEmit` passes on the touched file
- [ ] Manually verify: navigate client-side to a product page with `de` locale active where a translation key is missing under `product.pricingSchedule` — page should render with the English fallback text instead of crashing


Tirodem:
QA OK :heavy_check_mark:

Tirodem (approved):

PR #2681, opened 2026-07-15
🐛 Fix CurrencyCalculator SSR crash on CMS pages (#2680)

### LINKED ISSUE #2680: 🐛 CurrencyCalculator crashes SSR when included on a CMS page
## Symptoms
Any CMS page containing `[CurrencyCalculator=<slug>]` returns a blank page on server-side render. The server logs:

```
TypeError: Cannot read properties of undefined (reading 'find')
    at .../CmsDesign-*.js
    at .../CmsPage2-*.js
```

The widget was designed to be embeddable in CMS pages, but embedding it currently makes the whole page unrenderable.

## Expected behavior
Adding `[CurrencyCalculator=<slug>]` in a CMS page renders the currency conversion widget, both on SSR (first load / refresh) and after client-side navigation.

Tirodem:
Fixes #2680.

`sortedCurrencies` and `currencyOptions` were declared with reactive `$:` blocks but were consumed by non-reactive `let` initializers a few lines below, which read `undefined` in SSR ordering.

### DIFF
--- a/src/lib/components/CurrencyCalculator.svelte
+++ b/src/lib/components/CurrencyCalculator.svelte
@@ -6,8 +6,8 @@
 	import { getCurrencyFromCountry } from '$lib/types/Country';
 	import Select from 'svelte-select';
 
-	$: sortedCurrencies = sortCurrencies($currencies.main, $currencies.secondary);
-	$: currencyOptions = currenciesToSelectOptions(sortedCurrencies);
+	const sortedCurrencies = sortCurrencies($currencies.main, $currencies.secondary);
+	const currencyOptions = currenciesToSelectOptions(sortedCurrencies);
 
 	function getDefaultSecondCurrency(): Currency {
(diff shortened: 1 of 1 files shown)


Tirodem:
QA OK :heavy_check_mark:

slashbinslashnoname (approved):
## Approve

Looks right. `$:` `sortedCurrencies` / `currencyOptions` were consumed by `let` initializers that ran first in SSR, so `currencyOptions.find` blew up on CMS pages (`[CurrencyCalculator=]`). Switching those to synchronous `const` (`CurrencyCalculator.svelte:9-10`) fixes evaluation order instead of hiding the crash.

`sortCurrencies` doesn’t return an empty array, so the find is safe. No new hydration/locale mismatch; labels stay ISO codes.

Nit (non-blocking): `const` snapshots `$currencies` at init — options won’t live-update. Fine for a CMS widget.

Thanks.

PR #2691, opened 2026-07-22
🆕🤖 Admin session expiry warning banner (#2510)

Tirodem:
Closes #2510.

Two-phase banner injected into the ARM back-office layout : orange warning surfaces at T-5min with **Prolonger** (+30min additive on `session.expireUserAt`) and **Ignorer** (page-scoped) buttons, red expired banner at T=0 with a direct login link and a hint about the session-duration control on the login page.

Copy delivered in the 7 repo languages under `admin.session.expiry.*`. Client uses a local JS timer against `data.expireUserAt` — no server polling.

### DIFF
--- a/src/app.d.ts
+++ b/src/app.d.ts
@@ -31,6 +31,7 @@ declare global {
 			};
 			email?: string;
 			npub?: string;
+			expireUserAt?: Date;
 			sso?: Array<{
 				provider: string;
 				email?: string;
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -228,6 +228,9 @@ const handleGlobal: Handle = async ({ event, resolve }) => {
 						hasPosOptions: user.hasPosOptions,
 						recoveryEmail: user.recovery?.email
 					};
+					if (session.expireUserAt) {
+						event.locals.expireUserAt = session.expireUserAt;
+					}
 				}
 			}
 		}
@@ -281,7 +284,9 @@ const handleGlobal: Handle = async ({ event, resolve }) => {
 
 		// User-self endpoints (per-user prefs) only need an admin login, no role permission.
 		const normalizedAdminPath = event.url.pathname.replace(/^\/admin-[a-zA-Z0-9]+/, '/admin');
-		const isUserSelfAdminEndpoint = normalizedAdminPath === '/admin/back-office-bookmark';
+		const isUserSelfAdminEndpoint =
+			normalizedAdminPath === '/admin/back-office-bookmark' ||
+			normalizedAdminPath === '/admin/session/extend';
 
 		if (
 			!isUserSelfAdminEndpoint &&
--- a/src/lib/components/SessionExpiryBanner.svelte
+++ b/src/lib/components/SessionExpiryBanner.svelte
@@ -84,12 +84,7 @@
 {:else if isWarning && !dismissed}
 	<div class="bg-yellow-300 text-black p-3 flex flex-wrap items-center gap-3" role="alert">
 		<span>{t('admin.session.expiry.warning.message', { countdown })}</span>
-		<button
-			type="button"
-			on:click={extend}
-			disabled={extending}
-			class="btn body-mainCTA"
-		>
+		<button type="button" on:click={extend} disabled={extending} class="btn body-mainCTA">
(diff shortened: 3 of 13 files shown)


CypherJack:
The yellow warning banner appeared during tests without reloading the page. When no action was taken, it correctly turned into a red expired banner with different text. The Extend/Prolonger button on the yellow banner successfully extended the session. Dismissing via the Dismiss/Ignorer button works as expected.

Beside small UI roughness, the feature looks complete and usable in multiple languages. The 5-minute warning phase and extend button is a good addition beyond the original spec!

Non-blocking, good to merge from a testing standpoint.

CypherJack:
Update: Need to verify if the banner is sticky to the viewport because I got logged out of a session with no warning. My theory is that the warning banner appeared but was at the top of the page outside my viewport.

Edit: I was right, that's why I wasn't alerted by the banner.

image
image

slashbinslashnoname (changes requested):
## Request changes

Thanks — i18n (7 locales), `role="alert"`, and the T-5min **Prolonger** (+30min on `expireUserAt`) path look fine. No session-id leak (ISO timestamp, admin-only).

The original #2510 / QA issue isn’t actually fixed: the banner can sit off-screen, then the session dies with no save warning.

### 1. Banner is not viewport-fixed
`SessionExpiryBanner.svelte` uses `sticky top-0 z-40` inside `<main>` (`admin/+layout.svelte`). Sticky inside a scrolling main means a scrolled admin page never shows the warning. CypherJack’s note on #2510 was exactly that: banner out of viewport → silent logout.

**Fix:** `fixed top-0` with a high z-index, **outside** the scrolling `main` (layout chrome, not content).

### 2. Red “expired” state is mostly unreachable
At T=0 `hooks.server.ts` unsets `userId` / `expireUserAt` and redirects to login. The red banner only exists on the client timer **on the same page**. Combined with (1), “always visible / no silent save” does not hold.

**Fix:** keep (1) so the orange warning is unmissable before T=0. Don’t rely on the red banner as the safety net.

### Nits (non-blocking)
- `role="alert"` + 1s countdown will spam screen readers — `aria-live="polite"` once the banner is `fixed`.
- Client clock vs server `expireUserAt` skew.

Happy to approve once the banner is actually unmissable on a scrolled back-office page.

Tirodem:
New version pushed thanks to your comments. Just one point : I've kept the red banner. The initial idea is not to redirect user to login UI, when expired, but to say "don't move save your stuff before login again !".

slashbinslashnoname (changes requested):
All four points from the previous round are addressed: the banner is `fixed` and rendered
outside `<main>`, the countdown runs on `serverNow` rather than the workstation clock, and
`role="status"` + `aria-live="polite"` with the seconds `aria-hidden` stops the reader being
talked over. Keeping the red banner is your call and the reasoning holds.

One blocker left.

**"Prolonger" returns 403 for every role but Super Admin.** A POST under `/admin` goes
through `isAllowedOnPage(role, path, 'write')` (`hooks.server.ts:286`), which returns false
unless the role's own `write` list matches the path (`Role.ts:67`). Of the seeded roles only
Super Admin has `['/admin/*']`; Point of sale has `[]` and Ticket checker has
`['/admin/ticket/:id/burn']`, and any narrower custom role is in the same position.

The component then does:

```js
if (res.ok) { await invalidateAll(); }
```

so a 403 shows nothing at all — the button appears to do nothing, the countdown keeps
running, and the admin loses the work the banner exists to protect.

The codebase already has the shape for this: `isUserSelfAdminEndpoint` at
`hooks.server.ts:284`, "per-user prefs only need an admin login, no role permission".
Extending your own session is exactly that, so adding the path there is a one-liner.
Worth surfacing a failed extend in the banner too, rather than swallowing it.

**Extension is unbounded.** Each POST adds 30 minutes to the stored `expireUserAt`, with no
cap and no rate limit, so the value compounds indefinitely. That makes the session-duration
control — the one the red banner's own hint points at — advisory. Capping it at the configured
duration from now would keep the setting meaningful while still rescuing a long edit.

Nit: the fixed banner covers the top of the layout with no padding compensation. Fine for the
dismissible yellow one; the red one is not dismissible.

PR #2693, opened 2026-07-24
✨ Add Blink Lightning payment processor

pretyflaco:
## Summary

Adds a `blink` payment processor for the existing `lightning` payment method, letting be-BOP shops receive Lightning payments directly into a [Blink](https://blink.sv) account. It supports all three Blink receive modes, auto-detecting the right one from the merchant's configuration:

| Mode | Config | Invoice creation | Settlement detection |
|------|--------|------------------|----------------------|
| **API-key custodial** | `blink_...` API key (+ optional wallet id) | GraphQL `lnInvoiceCreateOnBehalfOfRecipient` at `api.blink.sv` | `invoiceByPaymentHash` poll |
| **LN-address custodial** | `user@blink.sv` (no key) | public no-key GraphQL, auto-detected via `accountDefaultWallet` | `lnInvoicePaymentStatusByHash` (ledger-aware — also catches intraledger Blink-app payments) |
| **LN-address non-custodial (Spark)** | `user@domain` (no key) | LNURL-pay callback | LUD-21 `verify` URL poll |

BTC-only, receive-only.

## Architecture fit

This is a new **processor** for the existing `lightning` **method**, not a new payment method — it follows the same SDK contract as the other Lightning processors (Swiss Bitcoin Pay, BTCPay Server, phoenixd, LND):

- Implements `PaymentProcessorDefinition` (`createPayment` / `checkPayment`); registered in `pp-registry.ts`.
- Settlement is driven by the **existing 2s order poller** (`maintainOrders()` → `checkPayment()`) — no websockets, no background tasks added.
- **Zero buyer-facing frontend changes** — the checkout renders the bolt11 via the existing generic Lightning QR path.
- DB-backed admin config page (`/admin/blink`) modeled on the Swiss Bitcoin Pay page, with Save / Reset / Test connection.
- Config auto-loads through the existing `runtimeConfig` change-stream; the mode is resolved at `createPayment` and re-derived deterministically at `checkPayment` from persisted payment fields (the Spark verify URL is stored in `checkoutId`; its presence discriminates the mode).

## Files

**New:** `src/lib/server/blink.ts` (client + pure helpers), `src/lib/server/blink.test.ts` (15 unit tests), `src/lib/server/sdk/contrib/PPBlink.ts` (adapter), `src/routes/(app)/admin[[hash=admin_hash]]/blink/{+page.server.ts,+page.svelte}` (admin UI).

**Edited (small):** `payment-methods.ts` (processor enum), `runtime-config.ts` (config default), `pp-registry.ts` (registration), `locks/order-lock.ts` (tap-to-pay exhaustiveness case), `config/+page.{server.ts,svelte}` (processor-preference selector + link), `adminLinks.ts` (nav entry).

## Testing

- **15 unit tests** for the pure logic (LN-address parsing, GraphQL endpoint derivation, invoice-status mapping, `SHA256(preimage) == paymentHash` validation).
- `pnpm run check` (svelte-check): **0 errors, 0 warnings**. ESLint clean on all touched files.
- **All three modes live-validated on mainnet** against a production deployment ([bebop.twentyone.ist](https://bebop.twentyone.ist)) with real Blink payments — API-key custodial, LN-address custodial, and LN-address non-custodial (Spark).

## Design note for reviewers

For the Spark path, the payment hash is derived from the LUD-21 `verify` URL rather than decoding the bolt11 — this avoids adding a bolt11-decoder dependency and mirrors the "verify key is authoritative" stance used by the Blink BTCPay Server plugin (the returned `pr` can carry a different hash).

## Out of scope (intentionally)

USD wallets (BTC-only; a USD default wallet is rejected with a clear config error), amountless/top-up invoices (orders always have a fixed total), the `myUpdates` websocket subscription (redundant with the existing poller), and send/balance/fee-probe (receive-only).

## Context

Unsolicited feature contribution. Lionel from the be-BOP team mentioned in person at Adopting Bitcoin (San Salvador) that a Blink integration would be welcome — happy to iterate on anything here.

### DIFF
--- a/src/lib/server/blink.test.ts
+++ b/src/lib/server/blink.test.ts
@@ -1,0 +1,78 @@
+import { createHash, randomBytes } from 'crypto';
+import { describe, expect, it } from 'vitest';
+import {
+	graphqlEndpointForDomain,
+	mapGraphqlInvoiceStatus,
+	parseLnAddress,
+	validatePreimage
+} from './blink';
+
+describe('blink', () => {
+	describe('parseLnAddress', () => {
+		it('splits user@domain', () => {
+			expect(parseLnAddress('alice@blink.sv')).toEqual({ username: 'alice', domain: 'blink.sv' });
+		});
+		it('defaults a bare username to blink.sv', () => {
+			expect(parseLnAddress('alice')).toEqual({ username: 'alice', domain: 'blink.sv' });
+		});
+		it('lowercases username and domain', () => {
+			expect(parseLnAddress('Alice@Blink.SV')).toEqual({ username: 'alice', domain: 'blink.sv' });
+		});
+		it('rejects empty input', () => {
+			expect(() => parseLnAddress('@blink.sv')).toThrow();
+			expect(() => parseLnAddress('alice@')).toThrow();
+		});
+	});
+
+	describe('graphqlEndpointForDomain', () => {
+		it('maps blink.sv to api.blink.sv', () => {
+			expect(graphqlEndpointForDomain('blink.sv')).toBe('https://api.blink.sv/graphql');
+		});
+		it('prefixes api. for other domains', () => {
+			expect(graphqlEndpointForDomain('example.com')).toBe('https://api.example.com/graphql');
+		});
+		it('does not double-prefix api.', () => {
+			expect(graphqlEndpointForDomain('api.blink.sv')).toBe('https://api.blink.sv/graphql');
+		});
+		it('uses http for localhost', () => {
+			expect(graphqlEndpointForDomain('localhost:4455')).toBe('http://api.localhost:4455/graphql');
+		});
+	});
+
+	describe('mapGraphqlInvoiceStatus', () => {
+		it('maps PAID to paid', () => {
+			expect(mapGraphqlInvoiceStatus('PAID')).toBe('paid');
+		});
+		it('maps EXPIRED to expired', () => {
+			expect(mapGraphqlInvoiceStatus('EXPIRED')).toBe('expired');
+		});
+		it('maps PENDING and unknown/null to pending', () => {
+			expect(mapGraphqlInvoiceStatus('PENDING')).toBe('pending');
+			expect(mapGraphqlInvoiceStatus(null)).toBe('pending');
+			expect(mapGraphqlInvoiceStatus(undefined)).toBe('pending');
+			expect(mapGraphqlInvoiceStatus('SOMETHING_ELSE')).toBe('pending');
+		});
+	});
+
+	describe('validatePreimage', () => {
+		it('accepts a preimage whose sha256 equals the payment hash', () => {
+			const preimage = randomBytes(32).toString('hex');
+			const paymentHash = createHash('sha256').update(Buffer.from(preimage, 'hex')).digest('hex');
+			expect(validatePreimage(preimage, paymentHash)).toBe(true);
+		});
+		it('is case-insensitive on the payment hash', () => {
+			const preimage = randomBytes(32).toString('hex');
+			const paymentHash = createHash('sha256').update(Buffer.from(preimage, 'hex')).digest('hex');
+			expect(validatePreimage(preimage, paymentHash.toUpperCase())).toBe(true);
+		});
+		it('rejects a mismatched preimage', () => {
+			const preimage = randomBytes(32).toString('hex');
+			const wrongHash = createHash('sha256').update(Buffer.from('deadbeef', 'hex')).digest('hex');
+			expect(validatePreimage(preimage, wrongHash)).toBe(false);
+		});
+		it('rejects a malformed preimage', () => {
+			expect(validatePreimage('not-hex', 'a'.repeat(64))).toBe(false);
+			expect(validatePreimage('abcd', 'a'.repeat(64))).toBe(false);
+		});
+	});
+});
--- a/src/lib/server/blink.test.ts
+++ b/src/lib/server/blink.test.ts
@@ -1,11 +1,14 @@
 import { createHash, randomBytes } from 'crypto';
-import { describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
+	blinkLookupInvoice,
+	extractPaymentHashFromVerifyUrl,
 	graphqlEndpointForDomain,
 	mapGraphqlInvoiceStatus,
 	parseLnAddress,
 	validatePreimage
 } from './blink';
+import { runtimeConfig } from './runtime-config';
 
 describe('blink', () => {
 	describe('parseLnAddress', () => {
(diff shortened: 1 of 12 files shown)


Tirodem:
Wow, that's a pleasant surprise, thanks for that.
I did a quick test order on bebop.twentyone.ist, I'll dig more in a couple of days to check the back-office & settings parts.
Thanks for the contribution and I'll try to check that asap (I have a few QA attention points in mind, especially since https://github.com/be-BOP-io-SA/be-BOP/releases/tag/rel%2F2026-06-23%2Fe8c307f inegration).
@slashbinslashnoname please also feel free to check this next week please.

pretyflaco:
Thanks @Tirodem, glad it's welcome! A few notes that may help your QA pass:

**Testing:** all three receive modes were live-validated on mainnet at bebop.twentyone.ist before opening this PR — API-key custodial, LN-address custodial (paid from a Blink app, i.e. the intraledger case LUD-21 verify would miss), and LN-address non-custodial (Spark).

**Re: the `rel/2026-06-23/e8c307f` attention points** — the implementation is aligned with both parts of that release:
- Blink failures during payment creation propagate through `generatePaymentInfo`'s `PaymentGenerationError` wrapper, so customers get the graceful banner + retry (and super-admin notification), never a raw error page.
- The `/admin/blink` page implements the `testConnection` action with the same pattern as the other PSP pages: `testProcessorConnection('blink')`, server-side rate limit (5/min/IP), 10s UI debounce.

**About the failing "Check branch is up to date" CI job:** this looks like a false negative for fork PRs rather than a stale branch. The branch is based on the current `main` tip (`885a5dde` is the merge-base). The job's checkout step uses `ref: ${{ github.head_ref }}` without a `repository:` input, so it looks for the branch in `be-BOP-io-SA/be-BOP` rather than the fork and fails with "branch could not be found" — it will do this for any external contribution. Adding `repository: ${{ github.event.pull_request.head.repo.full_name }}` to that checkout step would fix it; happy to leave that to you since it's a CI change.

Happy to iterate on anything you or @slashbinslashnoname find.

CypherJack:
I have tested both LN Address and the API key options. I have been able to successfuly buy a product on both methods (had to change de-prioritize phoenixd in General) and get the money on blink.

However for some limitation I wasn't able to test with the wallet ID. This has nothing to do with be-BOP or your code but me and my limited access to blink.

CypherJack:
@slashbinslashnoname you can do the PR review.

Tirodem:
@pretyflaco QA is done, we expect code review today or tomorrow :ok_hand:

slashbinslashnoname (changes requested):
## Request changes

Thanks for the contribution — architecture (processor on existing `lightning`, poller, no buyer UI) looks right, and QA on LN-address + API-key is enough to trust the happy path. Four fail-open cases still need to be closed before merge; they're all in `src/lib/server/blink.ts`.

### 1. Spark must not mark paid without a valid preimage
`blinkCheckSparkInvoice`: `settled === true` returns `'paid'` even when `preimage` is missing or `SHA256(preimage) !== paymentHash`. The mismatch is warn-only.

A LUD-21 response that sets `settled: true` without a matching preimage would complete the order.

**Fix:** mismatch or absent preimage → `'pending'` / `'failed'`, never `'paid'`.

### 2. Configured `walletId` skips the BTC check
`blinkResolveApiKeyWallet` returns `runtimeConfig.blink.walletId` as-is. The `walletCurrency !== 'BTC'` reject only runs on the default wallet. A USD wallet id would still mint invoices.

**Fix:** `walletById` lookup + reject non-BTC, same as the default-wallet path.

### 3. GraphQL probe fail-opens to Spark
`blinkResolveMode`: any error other than `/does not exist/i` (timeout, 429, network) returns `{ mode: 'spark' }` for that call. A temporary Blink GraphQL outage would send a custodial merchant down the LNURL Spark path and the wrong settlement poller.

**Fix:** ambiguous probe errors → throw / retry, do not assume Spark.

### 4. Spark `paymentHash` from `verifyUrl.split('/').pop()`
No `/^[0-9a-f]{64}$/` check. A trailing slash, query string, or unexpected LUD-21 URL shape stores garbage in `invoiceId`.

**Fix:** parse the URL path and validate a 64-char hex hash before persisting.

---

Lint/test are green. The fork “branch up to date” CI miss is not a blocker. Wallet-id path was not live-tested on QA — (2) covers that.

Happy to approve once those four are fail-closed.

pretyflaco:
Thanks @slashbinslashnoname — all four are now fail-closed in `b89e9501` (separate commit so the review delta is easy to see):

1. **Spark preimage** — `settled:true` now only returns `paid` when a preimage is present AND `SHA256(preimage) == paymentHash`. Missing or mismatched preimage → `pending` (warn-logged), never completes the order.
2. **Configured `walletId`** — now verified via `walletById` and rejected when `walletCurrency !== 'BTC'`, same as the auto-resolved default wallet (extracted a shared `assertBtcWallet`).
3. **Probe fail-open** — an ambiguous `accountDefaultWallet` error (timeout / 429 / network) now throws instead of assuming Spark, so a temporary GraphQL outage can't route a custodial merchant down the wrong settlement poller. Only an explicit "does not exist" still resolves to Spark.
4. **Spark `paymentHash`** — now parsed from the LUD-21 verify URL via `new URL().pathname` and required to match `/^[0-9a-f]{64}$/` (tolerates trailing slash / query string); throws a clear error otherwise.

Added 11 unit tests (26 total): preimage fail-closed cases, `status:ERROR` → `failed`, transient non-OK → `pending`, and `extractPaymentHashFromVerifyUrl` edge cases. `pnpm run check` 0 errors, eslint clean.

**One related finding I deliberately did *not* fold into this PR:** I checked whether the runaway LUD-21 verify polling fixed in [Kukks/BTCPayServerPlugins#149](https://github.com/Kukks/BTCPayServerPlugins/pull/149) exists here. Mostly no — that bug comes from BTCPay's per-instance `Listen()` pollers (3s fixed rate, broken backoff, fan-out). be-BOP has none of that: it reuses the existing 2s `maintainOrders()` loop with no per-invoice listeners, and a `status:ERROR` (not-found) maps to `failed`, which drops the invoice from the poll set. The only residue is that a healthy *unpaid* invoice is re-polled every 2s for its whole expiry window (F1/F3 flavour). Fixing that properly means adding per-invoice backoff/interval state to the shared `maintainOrders` loop that every processor uses — out of scope for a payment-processor PR. Happy to open a follow-up to discuss an age-stepped minimum interval if you think it's worth it.

slashbinslashnoname (approved):
Approve.

The four fail-open cases are closed on b89e9501: Spark preimage required and hashed, configured walletId goes through walletById + BTC check, ambiguous GraphQL probe errors throw instead of assuming Spark, paymentHash parsed from the verify URL and validated as 64-char hex.

QA already signed off. Residuals noted (assertBtcWallet if currency missing, Spark ERROR stops the poller as designed, LNURL verify hash trust) are not blockers.

PR #2715, opened 2026-08-12
🤖 fix(hp): HIBP fail-open + AuthLink contact prefill + safe next redirect

edoumo:
## Fixes HP (Peak Learn)

1. **HIBP fail-open contrôlé** : `checkPasswordPwnedTimes` entouré de try/catch + `AbortSignal.timeout(3000)` — l'API pwnedpasswords injoignable ne bloque plus le login admin (conteneur sans egress).
2. **AuthLink contact prefill** : le JWT AuthLink accepte `firstName`/`lastName` signés et matérialise `personalInfo` (email) → checkout prérempli nativement.
3. **Safe next redirect** : paramètre `next` restreint à une allowlist stricte (`/checkout`, `/cart`, `/orders`, `/identity`, `/login`) — pas d'open redirect.
4. **encodeURIComponent(next)** dans le formulaire de validation AuthLink (le `&qty=1` cassait la regex).


Tirodem:
Thanks @edoumo for the contribution. It'll be reviewed in the next days.

slashbinslashnoname (changes requested):
## Request changes

Thanks — the `next` allowlist + `encodeURIComponent` fix the open redirect, and AuthLink names coming from a signed JWT (not the client) is the right constraint.

Two fail-open / overwrite issues before merge.

### 1. HIBP fail-open is global
`checkPasswordPwnedTimes` catching everything and `return 0` means: if pwnedpasswords is unreachable, a known-pwned password is **allowed**.

That’s acceptable for **admin login** (container without egress shouldn’t lock you out). It is not acceptable for **setting** a password: `createSuperAdminUserInDb` and client reset would happily store a pwned password.

**Fix:** split the call sites — login may fail-open; create/reset super-admin and client password set must fail-closed (reject or hard-error if HIBP is down). Don’t `return 0` as “not pwned”.

### 2. AuthLink prefill can wipe `UserIdentifier`
`login/+page.server.ts` ~147–157: `$set: { user: { email } }` on an existing `personalInfo` overwrites the whole `user` object (sessionId / userId / npub).

**Fix:** `$set` only `firstName` / `lastName` (and email if that’s the intent), or go through `userQuery(locals)` like identity.

### Nits (non-blocking)
- Extract `safeNext()` — duplicated in load/validate.
- Trim + max length 100 on names, same as identity.
- Comments are in french, put in english

Happy to approve once (1) and (2) are fail-closed / non-destructive, and translated in english

slashbinslashnoname:
Product-wise this still needs a discussion. It’s a trivial additive security UX, not a given.

Ideally we’d have opened a dedicated issue first and aligned on the behaviour (timing, Prolonger vs login-only, when we interrupt the back-office) before shipping a banner.

The request-changes on visibility still stand; this is separate from the implementation nits.

edoumo:
Thanks for the review — both blocking points are addressed in `a5b2f09`.

**1. HIBP fail-open is now scoped (fail-closed on password set/reset)**

`checkPasswordPwnedTimes` now returns `number | null`:
- `null` = HIBP API unreachable (network error or HTTP error) — distinct from "not pwned" (`0`).
- **Login** (admin `+page.server.ts`): stays fail-open — `null` does not block, only an actual pwned count > 0 rejects.
- **`createSuperAdminUserInDb`**: fail-closed — `null` throws 503 "Password breach check unavailable, please retry later".
- **Admin password reset** (`/admin/login/reset/[token]`): fail-closed — same 503 on `null`, and it now actually checks HIBP server-side (previously the server action stored the password without any HIBP check; only the client-side `.svelte` checked). Client-side forms also block submission with a new `login.password.unavailable` message when the check is unavailable.

**2. AuthLink prefill no longer wipes UserIdentifier**

`login/+page.server.ts` validate: `$set` now only touches `firstName` / `lastName` / `updatedAt`; `user: { email }` is set **only in `$setOnInsert`** (upsert), so an existing `personalInfo` keeps its `sessionId` / `userId` / `npub`.

**Nits (also done)**
- `safeNext()` extracted — shared by `load` and `validate`.
- `firstName` / `lastName`: `.trim()` + `max(100)` (same as identity).

Validation: `svelte-check` 0 errors / 0 warnings; i18n key parity verified across all 7 languages.

edoumo:
Thanks @slashbinslashnoname — I resumed the interrupted cleanup and verified the PR state. The remaining review feedback is now addressed: newly added code comments are in English and the PR title carries the 🤖 marker for the AI-assisted contribution. The HIBP fail-open/fail-closed split and non-destructive AuthLink behavior remain unchanged. npm run check: PASS. Ready for re-review when convenient.

slashbinslashnoname (approved):
Approve.

Product decision remains with @Tirodem.

edoumo:
Product/UX discussion for the HIBP warning behaviour is now tracked separately in #2716, following the review feedback. This keeps the product decision separate from the implementation fixes in this PR.

Tirodem:
@edoumo the workflows have been approved and ran on `970cb8ce`. Lint fails on one formatting nit.

`src/routes/(app)/login/+page.svelte` — the `action` attribute of the AuthLink validation form is one line too long for prettier. Running `pnpm run format src/routes/\(app\)/login/+page.svelte` gives:

```svelte
<form
	method="post"
	action="?/validate&token={$page.url.searchParams.get('token')}&next={encodeURIComponent(
		data.next
	)}"
>
```

That is the only diff prettier produces across the eight files it checked.

Tirodem:
~~@edoumo QA feedback.~~

~~Signing in with a wrong password, while the breach check service is unreachable, tells the user the breach check is unavailable instead of telling them their password is wrong.~~

~~At sign-in no password is being set. The state of haveibeenpwned should not disturb the journey of a user who forgot their password or made a typo: it confuses them, and it creates support noise.~~

Edit : my bad, this case is already on be-BOP, it was spotted through haveibeenpwned blocked testing, I'll create a ticket, don't bother with this comment.
https://github.com/be-BOP-io-SA/be-BOP/issues/2775

Tirodem:
Otherwise, QA OK, I appreciate the contribution. Fix the lint and we'll be good ! :heavy_check_mark:

edoumo:
Thanks @Tirodem — both remaining QA points are now addressed.

- Admin sign-in now stays fail-open when HIBP is unavailable: on `null`, the form is submitted so the server can return the actual authentication result. Password creation/reset remain fail-closed server-side.
- Applied the requested Prettier formatting to the AuthLink validation form.

The external-fork “branch is up to date” workflow issue is left untouched as discussed.

Ready for final QA/review. Thanks!

PR #2720, opened 2026-08-18
🆕🔌 API v1: API keys, auth middleware, and batch order write (#2687)

### LINKED ISSUE #2687: 🆕🔌 Provide order API endpoint (write)
Can be foundation for https://github.com/be-BOP-io-SA/be-BOP/issues/2616

We need to provide an API endpoint to allow a third party to write orders in be-BOP.

be-BOP, in that case, will be used as concentrator, so each final PoS should carry its internal order ID & payment mean.

The external PoS system will send its list of orders on be-BOP API endpoint, that shall write them in be-BOP database.

Important informations :
- Catalog consistency : orders should be supposedly made only from existing product id (after #2686 ). If not, we still should save the order (be-BOP can manage old olders with deleted products) but should be registered with an integrity label / warning.
- Currency snapshot consistency : they should be done only with the be-BOP accounting (or main). So, no currency snapshot to save & send.
- Custom fields : for various purposes (event name, seller ID, PoS ID), new fields can be pushed as custom fields in be-BOP order objects. They won't be used as UI but can be in the future, and can be exploited through raw data directly.

To decide : API call answer (ok, "ok but" & shit happened cases)

Obviously : rate limit + access control / endpoint security

slashbinslashnoname:
Face A of the public HTTP API: API keys, authentication middleware, and the batch order write.

Closes #2687

## What lands here

- **API keys** — Mongo stores `keyHash = SHA-256(secret)` plus a non-secret prefix; the secret is never persisted. Admin UI under Settings → API Keys (super-admin), with a one-shot reveal flash.
- **Face A middleware** — isolated from the SSO/session handlers: maintenance passthrough for health/contract/docs, IP safety net, CORS allowlist, uniform error envelopes.
- **Batch order write** — `POST /api/v1/orders`, 1–100 commands per request, idempotent on `(apiKeyId, externalOrderId)` enforced by a sparse unique index set atomically at insert.
- **Contract** — hand-maintained `openapi.ts` served at `/api/v1/openapi.json`, Swagger UI at `/api/v1/docs`.

## Design records

- `docs/en/api/adr-api-faces.md` — why Face A (API keys) must not be merged with Face B (storefront session Bearer).
- `docs/en/api/adr-2687-batch-semantics.md` — why a batch returns 200 with per-command results rather than 207.
- `docs/en/api/v1-architecture.md` — module map, scalability notes, non-goals.

## CORS is database-only

`runtimeConfig.apiV1.corsOrigins` is persisted in the database and edited in Admin → API Keys. There is deliberately no environment variable: an env bootstrap would be a second source of truth for the same setting. Empty allowlist denies all browser cross-origin access; `*` is never accepted.

## Key lifecycle is admin-UI only

No CLI script ships with this. A second path to mint credentials would be untested, and it could not offer the one-shot reveal the UI gives. Mongo indexes for `apiKeys` are created on application boot by `createIndexes()`.

## Review notes

Known residual risk, recorded in the architecture doc: key hashes carry no pepper or per-key salt, so a database leak allows offline brute-force. Acceptable for now, to revisit before keys are treated as long-lived high-value credentials.

First of a stack of 5 — everything else in the API v1 series builds on this tree.


slashbinslashnoname:

PR #2721, opened 2026-08-18
🆕🔌 API v1: catalog read, storefront unique key, paid-orders poll

slashbinslashnoname:
Read side of Face A: catalog, storefront unique key, and the paid-orders poll.

Closes #2686
Closes #2688
Closes #2689
Closes #2616

**Depends on #2720** — merge that one first; this PR targets its branch and its diff is only the read side.

## What lands here

- **`GET /api/v1/catalog/products` and `/{id}`** (`catalog:read`) — stable DTO: ids, aliases, type, translated name and short description, price in **minor units**, PWYW flag, shipping, tags, variations, available stock. Hidden products (neither eShop nor retail visible) are omitted. (#2686)
- **Storefront unique key** — `?key=…` preselects a unique artifact secret, stored on the cart line and copied onto the order. Not an API key: a customer-facing product identifier. (#2688)
- **`GET /api/v1/orders/paid`** (`orders:read`) — poll, not a webhook. Only orders with at least one paid payment; payload carries product lines, the optional `uniqueKey`, and the amount **actually paid**. Replaces a push webhook for partners that prefer to pull. The per-product paid webhook (#2646) is unrelated and not sufficient for this flow. (#2689)
- **RFC spike** — `docs/{en,fr}/api/rfc-2616-public-api-spike.md`, the exploration these endpoints came out of. (#2616)

## Review notes

`listPaidOrders` is deliberately not scoped to the calling key: a partner poller reads the shop's paid orders, which is the point of #2689.

PR #2722, opened 2026-08-18
🦋 Admin API keys: UI pass, scoped categories, drop live/test environment

slashbinslashnoname:
Admin polish on the API Keys screens, plus removal of a cosmetic key attribute.

No source issue — follow-up on the #2687 UI.

**Depends on #2721.**

## What lands here

- Key list redesigned as cards, compact action buttons, CORS origins moved out of the key list into its own block.
- Scopes grouped by category with select all / select none.
- **`environment` (live/test) dropped.** It was cosmetic: it changed the secret's shape but nothing in authentication, authorization, or rate limiting ever read it, so it suggested an isolation the code did not provide. Secrets are now `bebop_ak_…` throughout.

## Review notes

The `environment` removal touches the stored `ApiKey` shape. No migration is included because no deployment runs this code yet — the whole `/api/v1` surface is unreleased.


slashbinslashnoname:

PR #2723, opened 2026-08-18
🆕🔌 API v1: conditional GETs — ETag / If-None-Match / 304 on reads (#2713)

### LINKED ISSUE #2713: 🆕 be-BOP endpoint API technical foundations
Requirement for #2686, #2687, #2688 and #2616 

For @slashbinslashnoname to complete

slashbinslashnoname:
Conditional GETs on the Face A reads: strong ETags, `If-None-Match`, `304`.

Closes #2713

**Depends on #2722.**

## What lands here

The RFC 9110 validators already lived in `validators.ts` but were called from nowhere. `jsonWithETag(body, request)` is the missing bridge: it serializes once, derives a strong ETag from the exact bytes, and short-circuits to a bodyless `304` repeating the same validator when `If-None-Match` matches (weak comparison, `*` and `W/` accepted, §13.1.2).

Wired to all three reads: catalog list, catalog detail, paid-orders poll. A PoS polling `/orders/paid` on a short interval now pays a `304` with no body when nothing changed.

## Deliberately not on the write path

`POST /api/v1/orders` takes no ETag precondition — the batch write has its own `(apiKeyId, externalOrderId)` idempotency key. An OpenAPI test locks that absence in so it is not added by reflex later.

## Also here

- `Cache-Control: private, no-cache` on the reads: per-key payloads stay out of shared caches, clients can still store and revalidate.
- CORS accepts `If-None-Match` and exposes `ETag` — without that plumbing a browser caller cannot read back the validator it is asked to replay.

### DIFF
--- a/docs/en/api/v1-architecture.md
+++ b/docs/en/api/v1-architecture.md
@@ -114,7 +114,7 @@ src/lib/server/api/
     cors.ts
     errors.ts
     openapi.ts
-    validators.ts              # ETag / If-Match helpers (not wired to writes yet)
+    validators.ts              # ETag / If-Match helpers (wired to GET reads, never to writes)
     schemas/orders-write.ts
     orders/
       writeBatch.ts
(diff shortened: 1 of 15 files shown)


slashbinslashnoname:

PR #2724, opened 2026-08-18
🐛 API v1: review follow-ups on the read and write paths

slashbinslashnoname:
Review follow-ups on the API v1 series: three defects, one dead configuration path, two CLI scripts retired.

No source issue — findings from a review of the whole stack.

**Depends on #2723.** Last of the 5 — three commits.

## Bug fixes

- **Variations were never priced on Face A.** The web cart applies `productPriceWithVariations` when a line is added; `resolveProducts` did not. `createOrder` therefore computed the order total from base prices and only patched `item.customPrice` afterwards, too late for the total, the VAT, or the payment amount. A product priced 0 with a paid variation looked like a zero total, `writeOne` picked `paymentMethod: 'free'`, and the order was marked paid with no money moving.
- **Cross-currency `amountPaid` was summed raw** — the `if (snap.currency === currency)` branch and its fallback were byte-identical, so the conversion had never been written.
- **Pagination derived `hasMore` from the mapped array**, after rows had been dropped. One dropped row on a full page zeroed `nextCursor` and left a poller convinced it had reached the end.
- **The IP rate-limit net pooled every caller** under a fixed `unknown-ip` key when `getClientAddress()` threw, so a misconfigured proxy capped the whole API at 120 req/min shop-wide.
- **`PaidOrderItem.unitPrice` ignored discounts** — `quantity * unitPrice` came out above `amountPaid` on any discounted line.
- **Admin key expiry**: the timezone offset was evaluated during SSR (server zone applied to the admin's wall clock on the no-JS path), and a past expiry was accepted, minting a key that `isApiKeyUsable` rejected on first use.

Each fix ships with a regression test; the pagination and variation tests both fail on the previous commit.

## Left open on purpose

- `GET /api/v1/orders` returns every order including unpaid ones, is absent from the OpenAPI document, and is the only read without an ETag. Needs a product call: document it, scope it, or drop it.
- The `ts-node` devDependency predates this work (added in #207) and may now be unused repo-wide.

### DIFF
--- a/src/lib/server/api/create-action.prg.test.ts
+++ b/src/lib/server/api/create-action.prg.test.ts
@@ -50,6 +50,15 @@ function mockCookies(initial?: Record<string, string>) {
 	return { cookies, jar };
 }
 
+/** Always-valid expiry, so the suite does not rot the day a hard-coded date goes past. */
+function futureIso(): string {
+	return new Date(Date.now() + 30 * 24 * 60 * 60 * 1000).toISOString();
+}
+
+function pastIso(): string {
+	return new Date(Date.now() - 60_000).toISOString();
+}
+
 describe('api key create PRG + reveal one-shot', () => {
 	beforeEach(() => {
 		createApiKey.mockReset();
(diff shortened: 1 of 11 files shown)


slashbinslashnoname:

PR #2725, opened 2026-08-18
🆕🔌 API v1: public HTTP API — keys, catalog reads, batch order write, conditional GETs

### LINKED ISSUE #2616: 🆕  Expose shop data via API so users can build a custom storefront, bypassing the Default Layout
Today every be-BOP storefront is rendered by be-BOP's own SvelteKit frontend and constrained to the Default Layout (the (app) layout + theme/CSS-variable system). Themes let admins restyle colors, fonts, and CTAs, but the structure, markup, and routing are fixed.

  This issue proposes a headless mode: a switch that turns a be-BOP instance into a content/commerce backend exposing its data and actions over a stable API, so a tenant can build and host their own storefront (Next.js, Astro, a static site, a native app, etc.) while still using be-BOP for catalog, cart, checkout, payments, orders, and stock.

  The goal is to bypass the Default Layout entirely for tenants who want full control of the frontend, without forking be-BOP.

  ### Motivation

  - Tenants who want a fully custom design beyond what the theme system can express (the theme model controls palette/typography/CTAs, not structure/layout).
  - Agencies/devs who want to reuse their own frontend stack while keeping be-BOP's commerce + Bitcoin/Lightning payment backend.
  - Decoupling frontend from backend so storefronts can be hosted/cached independently (CDN, edge, mobile app).

  ### Scope / what "headless" needs to expose

  - Read API for storefront data: products, categories/tags, prices/currencies, media, CMS/widget content, stock availability, products price history
  - Cart & checkout API: create/update cart, apply promo codes, custom checkout fields, create order, fetch order status.
  - Payment flows: initiate payment (incl. Bitcoin nodeless / Lightning), surface payment address/invoice + status for the custom frontend to poll/subscribe.
  - Webhooks or polling for order/payment status changes.
  - Auth model for the API (public read vs. authenticated mutations; per-tenant API token).

  ### Acceptance criteria

  - [ ] An admin can enable Headless mode from settings (per instance/tenant)
  - [ ] When enabled, the storefront-facing API exposes catalog, cart, checkout, payment, and order data needed to render a shop externally
  - [ ] Checkout + payment (including all payments systems) can be completed end-to-end through the API without loading be-BOP's Default Layout pages
  - [ ] API authentication is documented (public read scope vs. authenticated/mutating scope, token issuance)
  - [ ] Admin/back-office continues to work unchanged while headless mode is on
  - [ ] Documented behavior for the default frontend when headless is enabled (e.g. storefront routes disabled/redirected, or left intact and simply unused)
  - [ ] API surface is documented (endpoints/payloads) so a third party can build a storefront against it

  ### Open questions

  - API style: REST vs. GraphQL vs. typed RPC? Should we expose the existing SvelteKit endpoints in a stable, versioned way, or introduce a dedicated /api/v1/... surface?
  - Payment polling vs. webhooks for the external frontend to track Bitcoin/Lightning payment status.
  - CORS / token model for browser-based custom storefronts on a different origin.

  ### Technical hints

  - The Default Layout lives in the (app) route group; headless mode would gate or bypass that group rather than the back-office (admin[[hash=admin_hash]]) routes.
  - Much of the needed logic already exists server-side (cart, checkout actions, generateNodelessAddress / payment handling) — headless is largely about exposing it as a stable, documented, versioned API surface rather than new commerce logic.
  - Consider versioning from day one (/api/v1/) since external storefronts will pin to it.

### LINKED ISSUE #2686: 🆕🔌 Provide catalog API endpoint (read)
Can be foundation for #2616 

We need to provide an API endpoint to allow a third party to request a product list & details.

Product list should be filtered with type and tags, more custom fields in the future.

Obviously : rate limit + access control / endpoint security

### LINKED ISSUE #2687: 🆕🔌 Provide order API endpoint (write)

### LINKED ISSUE #2688: 🆕🔌 Add unique ID variation on product page & checkout
Let's say I have a product named "artifact".

I can access to it's product through `mybe.bop/product/artifact`

Let's say I use features like #2646 to process informations after those artifacts are ordered.

And let's say I have many artifacts but each of them are unique and identified by a secret ID.

I don't want to allow for anyone to access artifact from another, but I want each and any one to access their own artifact through a QR Code.

The QR Code URL will send to `mybe.bop/product/artifact?key=kfdjsfeaz12845ND9xezj91820`

When I land on this URL :
- I want for the key (here `kfdjsfeaz12845ND9xezj91820`) to be hidden on product page display
- I want the key value information to be kept in memory for add to cart
- I want it to be saved on cart then order
- I want it to be sent by #2646 as product information

### LINKED ISSUE #2689: 🆕🔌 Call external API after product payment processing
Can be foundation for https://github.com/be-BOP-io-SA/be-BOP/issues/2616

We need to call and API endpoint of a 3rd party after the sale of a product.

Highly similar to https://github.com/be-BOP-io-SA/be-BOP/issues/2646 but less experimental (might work but the sent information should be updated).

When a product is sold ( = present on a fully paid order), for this particular case, we need to send :
- order ID
- product secret ID (check #2688)
- paid amount (product will be PWYW, or product with variation)

Obviously : rate limit + access control / endpoint security

### LINKED ISSUE #2713: 🆕 be-BOP endpoint API technical foundations

slashbinslashnoname:
The whole of be-BOP's public HTTP API (Face A), previously split across a six-PR stack. Those PRs were never reviewed, so they have been folded into this one rather than merged one by one; the 11 commits below still read as the original steps.

Closes #2687
Closes #2686
Closes #2688
Closes #2689
Closes #2713
Closes #2616
Refs #2646

## 1 — Foundation: API keys, middleware, batch write

- **API keys** — Mongo stores `keyHash = SHA-256(secret)` plus a non-secret prefix; the secret is never persisted. Admin UI under Settings → API Keys (super-admin), with a one-shot reveal.
- **Face A middleware**, isolated from the SSO/session handlers: maintenance passthrough for health/contract/docs, IP safety net, CORS allowlist, uniform error envelopes.
- **`POST /api/v1/orders`** — 1 to 100 commands per request, idempotent on `(apiKeyId, externalOrderId)` via a sparse unique index set atomically at insert.
- **Contract** — hand-maintained `openapi.ts` at `/api/v1/openapi.json`, Swagger UI at `/api/v1/docs`.

CORS lives only in `runtimeConfig.apiV1.corsOrigins`, edited in the admin. No environment variable: a bootstrap env would be a second source of truth for the same setting. An empty allowlist denies all browser cross-origin access, and `*` is never accepted. Key lifecycle is admin-UI only — a CLI path would be untested and could not offer the one-shot reveal.

## 2 — Reads: catalog, storefront unique key, paid-orders poll

- **`GET /api/v1/catalog/products` and `/{id}`** (`catalog:read`) — stable DTO: ids, aliases, type, translated name and short description, price in **minor units**, PWYW flag, shipping, tags, variations, available stock. Products hidden from both eShop and retail are omitted.
- **Storefront unique key** — `?key=…` preselects a unique artifact secret, stored on the cart line and copied onto the order. Not an API key: a customer-facing product identifier.
- **`GET /api/v1/orders/paid`** (`orders:read`) — a poll, not a webhook, for partners that prefer to pull. Only orders with at least one paid payment; the payload carries product lines, the optional `uniqueKey`, and the amount actually paid. It is deliberately not scoped to the calling key: a partner poller reads the shop's paid orders, which is the point of #2689.
- **RFC spike** — `docs/{en,fr}/api/rfc-2616-public-api-spike.md`, the exploration these endpoints came out of.

## 3 — Admin UI pass, and `environment` dropped

Key list redesigned as cards, scopes grouped by category with select all / none, CORS origins moved out of the key list into their own block.

`environment` (live/test) is removed. It was cosmetic: it changed the secret's shape, but nothing in authentication, authorization or rate limiting ever read it, so it advertised an isolation the code did not provide. Secrets are `bebop_ak_…` throughout. No migration ships with it because no deployment runs this code yet.

## 4 — Conditional GETs

The RFC 9110 validators already lived in `validators.ts` but nothing called them. `jsonWithETag(body, request)` is the missing bridge: it serializes once, derives a strong ETag from the exact bytes, and short-circuits to a bodyless `304` repeating the same validator when `If-None-Match` matches (weak comparison, `*` and `W/` accepted, §13.1.2). Wired to all three reads, so a PoS polling `/orders/paid` pays a `304` with no body when nothing changed.

Not on the write path: the batch write has its own idempotency key, and an OpenAPI test locks that absence in so it is not added later by reflex. Reads also carry `Cache-Control: private, no-cache`, and CORS accepts `If-None-Match` and exposes `ETag` — without that, a browser caller cannot read back the validator it is asked to replay.

## 5 — Review follow-ups

Six defects found reviewing the stack, each with a regression test; the pagination and variation tests fail on the preceding commit.

- **Variations were never priced on Face A.** The web cart applies `productPriceWithVariations` when a line is added; `resolveProducts` did not. `createOrder` computed the total from base prices and patched `item.customPrice` afterwards — too late for the total, the VAT or the payment amount. A product priced 0 with a paid variation looked like a zero total, `writeOne` picked `paymentMethod: 'free'`, and the order was marked paid with no money moving.
- **Cross-currency `amountPaid` was summed raw** — the `if (snap.currency === currency)` branch and its fallback were byte-identical, so the conversion had never been written.
- **Pagination derived `hasMore` from the mapped array**, after rows had been dropped: one dropped row on a full page zeroed `nextCursor` and left a poller convinced it had reached the end.
- **The IP rate-limit net pooled every caller** under a fixed `unknown-ip` key when `getClientAddress()` threw, capping the whole API at 120 req/min shop-wide behind a misconfigured proxy.
- **`PaidOrderItem.unitPrice` ignored discounts**, so `quantity * unitPrice` exceeded `amountPaid` on any discounted line.
- **Admin key expiry** — the timezone offset was evaluated during SSR (server zone applied to the admin's wall clock on the no-JS path), and a past expiry was accepted, minting a key `isApiKeyUsable` rejects on first use.

## 6 — Paid-order webhook payload

The per-product paid-order webhook carried a timestamp, order number, contact details, billing address and custom checkout fields — nothing about the purchase. A receiver could not tell what was sold, could not read the artifact secret from the `?key=` flow, and could not reconcile a payment against an amount. It now carries the order id, product lines, `uniqueKey` and the amount charged.

## Left open on purpose

- `GET /api/v1/orders` returns every order including unpaid ones, is absent from the OpenAPI document, and is the only read without an ETag. It needs a product call: document it, scope it, or drop it.
- The `ts-node` devDependency predates this work (added in #207) and may now be unused repo-wide.

## Review order

The 11 commits are the original six steps in sequence, so the stack can still be read one step at a time.

https://claude.ai/code/session_018x8pkAdnC97J8F1n7uDqCK


Tirodem:
Seems good, thanks for the update :ok_hand: 
I'll check tuesday with Roos.
I think one of the first update will be to remove the lock (maybe on a specific parameter) on amountMinorSchema to manage refundable products, but let's only put it on the burner as a possibility for now.

Tirodem:
Feedbacks from partner :


## CORS

16. **Origin policy on [[https://pr-isca.pvh-labs.com/](https://pr-isca.pvh-labs.com/)](https://pr-isca.pvh-labs.com/%5D(https://pr-isca.pvh-labs.com/))

*Context:* browser clients authenticate solely via `X-API-Key` (no cookies, no ambient credentials). 
*Issue:* Implementation details of the client application leak into the server configuration's CORS allow list. 
*We suggest:* set `Access-Control-Allow-Origin: *`.

Header-token auth means CSRF is structurally impossible (custom header → 
preflight; no key → 401), the wildcard hard-forbids credentialed requests per 
the Fetch spec, and an allowlist only adds maintenance/outage risk with zero 
security gain. If you insist on an allowlist, we'll send our origins (but it's 
security theater at this point).


8. **Amounts as strings.**

*Issue:* minor-unit integers are encoded using number/decimals.

Floating point numbers are commonly lead to unindended rounding errors due to 
floating point numbers being decoded differently on different platforms.

*we suggest* using a string of conventional amounts (e.g. "20.05 CHF"), or a "20.05" string.


7. **`PosSaleResult` lacks `orderId`.**

*Issue:* only `orderUrl`; parsing URLs is fragile. 
*We suggest:* add `orderId`.


1. **Order attribution.**

*Context:* we consume the paid stream to credit bracelets. 
*Issue:* events carry no `customFields` nor `externalOrderId`, so our own orders 
come back anonymous and we credit twice. 
*We suggest:* include `externalOrderId` (ideally `customFields` too) in stream 
events and reads.

Tirodem:
New needs :
- (short term requirement) store endpoint requests & answers and display on logs on /admin/api-keys
- (mid term requirement) on the key, allow to set a parameter on be-BOP UI to bypass be-BOP functional rules (for example, product max qty per order : with this setting, we can create an order with a superior qty without being trashed by the endpoint

I'm adding the first, the second will be later because I feel it'll be more complex.

Tirodem:
we'll need to be able to set api rate limit & timeout options when creating / updating a key on UI

Tirodem:
to do in the really near future : when writing orders through endpoint, choose to disable email / nostr notification through an additional parameter

Tirodem:
Order created with failed payment status stays in pending, it should be failed

Tirodem:
We need a /api/v1/orders/stream announcing orders whatever their payment status: today only paid orders are streamed, so an integration has no way of hearing about an order that is still pending, or that failed. It would be the streaming counterpart of GET /api/v1/orders.

Tirodem:
Heads-up on the order stream: when a shop has data cleanup on expired/cancelled orders enabled, cancelling an order announces it twice on the stream. The cancellation writes the order once, and the anonymisation that follows writes it again; since anonymisation only touches fields the stream does not expose, both events carry exactly the same payload. Expiry very likely does the same, for the same reason.

Tirodem:
We have quite an issue with the MAX_CONCURRENT_STREAMS_PER_KEY.

On another branch, I've pushed it to 64, and every day or two days, with only a dozen of test devices, we go over the limit and are stuck in 429 "At most XX concurrent streams per API key".

From what I understand :
- streams are never closed
- as we're working with hardware going offline then online quite often

So we provide new streams, no cleaning, we're pilling up then boom 429.
And if we remove the limit, at one moment, having so many non-cleaned streams can be bad for performances.

Unless there's something I'm missing or I don't understand, we have to do something in here.

I'm trying something on s37-demo-2-fix to avoid that for this week-end demo then we'll need to talk about it and the various topics / feedbacks following transversal testing phase.

PR #2732, opened 2026-08-23
✨ Add CLINK Lightning payment processor (Nostr kind 21001)

WoompaLoompa:
## Summary

Adds CLINK (Common Lightning Interface for Nostr Keys) as a Lightning payment transport in be-BOP. CLINK uses Nostr kind 21001 encrypted events with NIP-44 to allow CLINK-compatible wallets (ShockWallet, ZEUS) to pay via the merchant's nOffer, while any Lightning wallet can pay the bolt11 QR code directly.

## What is CLINK?

CLINK is a **transport layer**, not a Lightning backend. It enables:

1. **Bolt11 QR** — any Lightning wallet can scan and pay
2. **nOffer scan** — CLINK wallets receive the same bolt11 via Nostr relay
3. **Receipt confirmation** — Lightning.Pub sends a payment receipt (second kind 21001 event)

## Implementation

- `src/lib/server/clink.ts` — CLINK protocol client (NIP-44, nOffer, Nostr subscription)
- `src/lib/server/sdk/contrib/PPClink.ts` — PaymentProcessorDefinition adapter
- `src/lib/types/ClinkSession.ts` — MongoDB session document type
- Admin UI at `/admin/clink` for configuration
- Documentation in 7 languages (en/fr/de/es-sv/it/nl/pt)

## Security (addressing BTCPay plugin review findings)

- **Relay SSRF protection** — validates relay URLs against private/internal IPs
- **BOLT11 validation** — network + amount checks on invoices from Lightning.Pub
- **No underlying processor fallback** — CLINK bolt11s can't be looked up by Blink/LND
- **nDebit is NOT mandated** — CLINK is transport only, settlement via existing Lightning processor

## Key Design Decisions

- Eager bolt11 creation via `ClinkSDK.Noffer()` → Lightning.Pub
- Payment detection via Nostr receipt callback exclusively
- Session persistence in MongoDB with TTL index (survives restarts)
- Persistent receipt listener with dual decryption (customer requests + Lightning.Pub receipts)

## Testing

- 35 unit tests passing (vitest)
- 0 type errors (svelte-check)
- E2E tested with mainnet CLINK wallet (ShockWallet)

## Files Changed

28 files, +6608 / -3105 lines


Tirodem:
Hello,
Thanks for the contribution.
But what's the point of updating assets/bebop-b.svg for payment integration ?

Tirodem:
And sorry for the delay, I'm still on days but, but I was able to take a review of your contribution (thanks for it).

Blocker as it stands: two customers checking out at the same time can't be told apart. `decoded.offer` comes from the merchant's configured nOffer — one value for the whole shop — yet it keys both the session store and the receipt callbacks, so the second checkout replaces the first. The receipt carries no payment hash, so whoever pays, it settles whichever session is currently registered. One customer pays, another gets credited.

Fixing that likely means a per-order key (payment hash or order id) and a receipt that can be tied back to it. Would this fit the SDK registry pattern from 96b3e59f, the way Taler, OSB and Blink did? Keeping per-invoice state in the order's payment document rather than a process-wide map would avoid the extra collection and listener lifecycle here.

Three things also look like leftovers from local work: `assets/bebop-b.svg` is now a small placeholder instead of the logo, there's a 1×1 pixel PNG under `static/pictures/`, and `pnpm-lock.yaml` moves from lockfileVersion 6.0 to 9.0 while `packageManager` still says pnpm 8.3.1. For info, `@shocknet/clink-sdk` receives the shop's nsec — it stays in-process, but worth pinning rather than `^1.6.1`.

Feel free to give us more informations, context and state of the art about Clink nature, origine and use, we'll be happy to dig more into it !

WoompaLoompa:
Hi Tirodem, thanks for the thorough review. Here's some context on CLINK and responses to each finding:

CLINK (Common Lightning Interface for Nostr Keys - clinkme.dev) was built by [ShockNet](https://github.com/ShockNet/clink) as a Lightning payment transport over Nostr (instead of traditional http/https). The merchant publishes an `noffer`, a bech32-encoded pointer containing `{pubkey, relay, serviceKey}`, which CLINK-compatible wallets (ShockWallet, ZEUS) use to request invoices via encrypted kind 21000/21001 events (NIP-44). The BOLT11 is then displayed as a standard QR that any Lightning wallet can pay.

The nOffer is designed as a **long-lived payment pointer** (like a Lightning address or on-chain address). Once a customer saves it, they can pay via it indefinitely. If the merchant changes their CLINK relay in the admin panel, old nOffer strings still point to the previous relay, there's no redirect mechanism, so the merchant would need to redistribute new QR codes. This is an inherent property of the nOffer encoding, not something be-BOP controls. I also announced this integration [here](https://stacker.news/items/1553246/r/06bc1a977d). 

## Findings addressed

1. **bebop-b.svg placeholder**:  fixed in `39da2820`
The original full logo (370×561) was accidentally replaced with a 32×32 placeholder in the initial commit. Restored.

2. **1×1 pixel PNG**: fixed in `39da2820`
Removed `static/pictures/6a875fcecee470464680e3bd.png`. No code references it.

3. **Concurrent checkout collision (blocker)**: fixed in `eb751d85`
Session key changed from `decoded.offer` (merchant-level, one per shop) to `SHA-256(bolt11)` (per-invoice). Each checkout now gets its own session, `pendingReceiptCallbacks` entry, and MongoDB document. Receipts are correlated to specific invoices via `#e` tags: LP receipt events carry `#e: [requestEventId]`, which we map to payment hashes through `eventIdToPaymentHash`. This avoids storing payment hashes in Nostr events while still providing deterministic matching.

4. **pnpm-lock.yaml v6→v9**: fixed in `13f42d2e`
Reverted to `lockfileVersion: '6.0'` to match `packageManager: pnpm@8.3.1`. Regenerated with pnpm 8 to add only the `@shocknet/clink-sdk` dependency.

5. **Pin @shocknet/clink-sdk**: fixed in `39da2820`
Pinned to exact `1.6.1` (no caret). As for the nsec: `@shocknet/clink-sdk` receives the merchant's Nostr private key to decrypt NIP-44 encrypted receipt events from LP. It stays in-process and never leaves the server — same trust boundary as the existing Blink integration.

slashbinslashnoname:
Thanks — lockfile, logo and pinned SDK are settled. But `eb751d85` moves the blocker rather than closing it, and the main flow now looks like it never detects a payment at all.

`createPayment()` calls `clinkRequestInvoice()` and only *then* `setClinkSession()`. The `#e → paymentHash` mapping the fix rests on is written in one place (`clink.ts:867`), guarded by `activeSessions.get(sha256(bolt11))` inside the listener — which gets the same LP event and reaches that lookup before the session exists. The map stays empty, the receipt falls through to the `decoded.offer` branch (`clink.ts:849`) where nothing is registered any more, and `checkPayment()` stays `pending` until expiry. Settlement rides on a race between two subscriptions on the same relay.

Four more I can't take on a payment path:

1. `PPClink.ts:86` — `paymentHash: sessionKey` = `sha256(bolt11)`. No real payment hash anywhere, so nothing reconciles against the node that got the sats, and we build Z-tickets from this.
2. `checkPayment()` echoes the *expected* amount on a `res: 'ok'` event carrying neither amount nor proof. Every other lightning processor asks the node.
3. `PPClink.ts:75` swallows errors and returns the raw nOffer with no session: relay down → checkout succeeds, customer pays, order expires. It must throw.
4. The public `checkPayment` action on `/order/[id]`: no user check, no rate limit, not CLINK-only, `onOrderPayment` outside the order lock.

Smaller: `clink.enabled` is write-only, the offer check at `clink.ts:665` compares the request against itself, `validateBolt11` accepts 1% under the ordered amount, the global maps break multi-process.

The three fix commits shipped with no test — the 35 existing ones cover pure helpers only.

To look again I'd want two concurrent checkouts with receipts in reverse order, a real payment hash confirmed against the backend, 3 and 4 fixed, rebase on `main`. And I'd reconsider the shape first: per-invoice state in the payment document, like Taler/OSB/Blink via the SDK registry (96b3e59f), removes the collection, both maps, the boot loader, the replay pass and the listener lifecycle.

Nothing against CLINK, it's an interesting fit. But this is the path where a shop gets paid or doesn't.

WoompaLoompa:
Thanks for the second round — you've got the shape right. Rather than patch the race again, I rebuilt it the way you suggested: per-invoice state in the payment document, settlement by the node that actually minted the invoice, and Nostr as transport only.

**Architecture**

- Invoice creation is delegated to be-BOP's own lightning processor: `PPClink.createPayment()` resolves through `getProcessorsForMethod('lightning')` (the same SDK registry as Blink/OSB, per 96b3e59f) and mints a real bolt11 via LND/Blink/Phoenixd. The invoice hash and bolt11 are stored on the payment entry (`paymentHash`, `invoiceId`), plus `meta.backend` recording which processor minted it.
- `checkPayment()` re-dispatches to that processor and asks the node for the invoice by payment hash — the node that actually received the sats. This replaces the receipt path entirely; there is no settlement dependency on Nostr.
- The `clinkSessions` collection, both in-memory maps, the boot loader, the replay pass, and Lightning.Pub metadata are gone. The listener is transport-only: it serves bolt11s to CLINK wallets over kind 21001 and validates exact-match BOLT11.
- Each payment owns its hash in its own document, so two concurrent checkouts are fully independent — there is no shared key for a second checkout to clobber, and a receipt for another invoice is simply not part of the flow anymore.

**Your four findings**

1. Real payment hash — yes. The invoice comes from the backend with its actual `payment_hash`; Z-tickets are built from that. `sha256(bolt11)` never appears.
2. `checkPayment()` now queries the node exactly like every other lightning processor (settled / still pending) instead of echoing the expected amount.
3. `createPayment()` now throws when invoice creation fails — no silent nOffer fallback. Relay/node down → checkout fails loudly; it never returns a raw nOffer with no backend behind it.
4. The public `checkPayment` action is secured: rate-limited, user-owner-scoped, CLINK-only, and it only re-runs the node check and reports the result — settlement stays in the poller under the order lock, with no `onOrderPayment` side effects.

**Smaller items**

- `clink.enabled` (write-only) deleted; enablement is `nOffer` + `relayUrl`, same as Blink.
- The offer check canonicalizes the merchant's own nOffer chain and compares it against the request decode — not "the request against itself".
- `validateBolt11` is exact-match on the amount — no 1% tolerance.
- Global maps removed, so no multi-process hazard remains.

**Tests** — `npx vitest run src/lib/server/clink.test.ts src/lib/server/sdk/contrib/PPClink.test.ts` → 39 passing:

- `PPClink.test.ts` (7): delegation to the configured processor, real payment hash + `meta.backend` on the payment, two concurrent creates with distinct hashes, `createPayment` throws on backend failure, `checkPayment` re-dispatches with the real hash, meta-less fallback (lnd → phoenixd → blink), and throws when no lightning backend is available.
- `clink.test.ts` (32): transport helpers, label/offer paths, exact-match BOLT11 rejection.

`svelte-check --tsconfig ./tsconfig.json` → 0 errors, 0 warnings. Branch is rebased on `main` and pushed; ready for another pass.

WoompaLoompa:
@be-BOP admins — this updates the CLINK PR to address the earlier feedback.

## Admin-selectable Lightning backend

CLINK remains a **transport layer only**. The backend that mints and settles the bolt11 is now chosen in **Admin > CLINK**:

- **be-BOP Lightning processor** (default) — invoice minting + settlement delegated to be-BOP's own configured LN processor, as before.
- **Lightning.Pub node** — invoices are minted by the merchant's Lightning.Pub node directly via its HTTP API (`POST /api/user/invoice/new`, Bearer token), and settlement is queried from that same node (`POST /api/user/payment/state`).

admin

## Reviewer feedback addressed

- **Real payment hash**: The payment hash is extracted from the *signed bolt11* itself via a dependency-free BOLT11/bech32 decoder (`clink.ts` `decodeBolt11PaymentHash`), verified against the Lightning BOLT#11 test suite (`lnbc2500u1pvjluez…` → `000102…090102`). The `sha256(bolt11)` placeholder is gone — the stored hash is the actual hash in the invoice.
- **Node-backed `checkPayment`, never an echo estimate**: `checkPayment` re-dispatches per payment to the backend recorded at creation time (`meta.backend`). For Lightning.Pub, `clinkCheckInvoiceViaLightningPub` calls `payment/state` with the stored bolt11 and uses the node-reported amount/settle timestamp; a paid-but-mismatched amount throws instead of trusting the local expected value.
- **No silent fallback**: a Lightning.Pub payment missing its bolt11 throws (`Missing Lightning.Pub invoice`); LP payments never fall back to the be-BOP processors.
- **Exact-amount validation**: minted LP invoices are validated with `validateBolt11(…,{ expectedAmountSat })` (no tolerance).
- **SSRF-safe**: the LP endpoint is validated against private/internal IP ranges before every mint and settle call, mirroring the existing relay handling.
- **Secured**: `checkPayment` remains a rate-limited action; pairing is untrusted for issued invoices.

## Scope / tests

- `clink.ts` / `clink.test.ts` (51 tests): BOLT11 decoder, LP mint, `clinkCheckInvoiceViaLightningPub`, config helpers.
- `PPClink.ts` / `PPClink.test.ts` (12 tests): `meta.backend` dispatch, LP create/check branches, no-fallback guarantee.
- Admin `clink` page (server + svelte): backend radio, conditional endpoint/token fields, `superRefine` validation, SSRF check.
- Docs updated in all 7 languages (`docs/{en,de,es-sv,fr,it,nl,pt}/clink-configuration.md`).

Full suite: **366 tests / 35 files pass**, `svelte-check` 0 errors / 0 warnings. Commit `0480fb65`.

Tirodem:
I'm kinda drowning in QA so far so I have to delay testing campaign for now.
Stay tuned, I'll give news soon.

PR #2734, opened 2026-08-25
🐛 Render the home page the same at / and at /home (#2161)

### LINKED ISSUE #2161: 🐛 Advance HTML editor doesn't have the same rendering with / or with /home
Image

Tirodem:
Closes #2161.

The site root and `/home` serve the same CMS page, but not through the same code path. The root route decides whether to skip sanitization by reading a field it never asked the database for, so that decision is always made on an empty value.

What a shop owner can lose on `/`, while `/home` keeps it:

- the styling written in the Advanced HTML editor, when the page uses raw content;
- the SEO metas set on the home page;
- the page's own description for search engines, which falls back to the shop-wide one.

The back-office gives no hint, since the preview follows `/home`.

This aligns the root on what `[slug]` already does, so both addresses render the same page the same way.

Redirecting `/` to `/home` would have hidden the symptom too, but it moves the canonical address of every shop to `/home`. The root stays the root.

### DIFF
--- a/src/routes/(app)/+page.server.ts
+++ b/src/routes/(app)/+page.server.ts
@@ -19,8 +19,15 @@ export const load = async ({ locals, url }) => {
 				},
 				fullScreen: 1,
 				hasEmployeeContent: 1,
+				// Without this, `cmsPage.displayRawContent` below is always
+				// undefined and the home page is sanitized at `/` but not at
+				// `/home` — the <style> block of an advanced-HTML page is
+				// dropped on the site root only.
+				displayRawContent: 1,
 				hasMobileContent: 1,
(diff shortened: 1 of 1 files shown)


Tirodem:
QA ok :heavy_check_mark:

slashbinslashnoname (approved):
Out of scope, but the same projection is incomplete twice more for the `error` page — `[slug]/+page.server.ts:42-56` and `[slug]/[sub]/[...catchall]/+page.server.ts:12-25`, both missing `displayRawContent` and `metas`. An error page written in advanced HTML is still sanitized. Opened #2770 for it.

Accepted.

PR #2736, opened 2026-08-25
🐛 Drop the POS tab buttons from the non-POS order page (#2703)

### LINKED ISSUE #2703: 🐛 `@@Back to order tab` button shown on order page regardless of entry point
## Overview
The `@@Back to order tab?` button appears on the order page for any logged employee accessing an order, even when they didn't come from the `/pos/touch` workflow. It should only appear when the page was reached from `/pos/touch`. Separately, the raw tradkey (`@@Back to order tab`) is displayed instead of a translated string, unlike other buttons on the page.

Image

Tirodem:
Closes #2703.

Opening an order that is attached to a tab — from the admin order list, a direct link, a bookmark — showed an employee two buttons leading into the point-of-sale touch flow they had not come from: **Back to order tab**, and **Continue** when the order was being split.

Clicking either dropped the employee into `/pos/touch`, away from the page they were working on. The first also displayed a raw placeholder instead of a label, in every language.

Both buttons are removed from that view. Inside the point-of-sale flow nothing changes: the return to the tab is already handled there, split included.

### DIFF
--- a/src/routes/(app)/order/[id]/+page.svelte
+++ b/src/routes/(app)/order/[id]/+page.svelte
@@ -563,22 +563,6 @@
 								>
 									{t('order.note.seeText')}
 								</a>
-								{#if data.order.orderTabSlug}
-									{#if data.splitMode && showContinue}
-										<a
-											href="/pos/touch/tab/{data.order.orderTabSlug}/split?mode={data.splitMode}"
-											class="btn lg:w-auto w-full btn-black self-end"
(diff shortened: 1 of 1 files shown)


slashbinslashnoname (approved):
Side effect worth noting: `@@Back to order tab` was the last `@@` marker left in `src/**/*.svelte` — this PR takes the count to zero.

Accepted.

PR #2737, opened 2026-08-25
🐛 Count only paid payments in the reporting dashboard (#2730)

### LINKED ISSUE #2730: 🐛 Fix reporting dashboard exporting pending payments alongside paid ones for the same order
## Overview

When an order has multiple payment attempts (e.g. a first "appel à paiement" left pending/abandoned, followed by a second one that succeeds), the Reporting dashboard's Payment Detail and Payment Synthesis tables list every payment on the order, not just the successful one. Once an order's status is `"paid"`, the filter used to select which payments get shown/exported only checks `payment.method` — it seems that it never checks `payment.status`. So the pending payment sits right next to the paid one in the export, both rows showing `Order Status: paid`, which reads as if the order got paid twice but with `Payment Status: pending`.

Image

This is what causes the disparity with the Z-ticket: the POS session close logic sums only payments with `status === 'paid'`, while the reporting dashboard counts/sums every payment on a paid order regardless of status.

It only happen when the waiter click on "Skip for now". Cancel button has the expected behavior and as the payment is canceled it does not appear in the reporting.

Image

Note that:
Z-ticket produce correct totals, only the reporting silently add pending payment on paid orders.

### Reproduce
Create an order, trigger two payment calls for it (first left pending, second paid) → open Reporting → Payment Detail / Payment Synthesis for a period covering that order → both payments appear, inflating the count and total versus the Z-ticket for the same period.

## Expected behavior
- Reporting should not include pending payments on orders that also have a paid payment.
- By default, only validated (status === 'paid') payments on paid orders should be counted/exported, matching what feeds the Z-ticket.
- Existing admin filters (includePartiallyPaid, includeExpired, etc.) can still surface non-paid payments explicitly, but the default view/export should not silently mix them in.

Tirodem:
Closes #2730.

An order can go through several payment attempts — one left hanging, then a second one that succeeds. Reporting listed them all. Payment Detail and Payment Synthesis showed two lines for that order, both marked as paid, and both counted in the totals. It read as if the customer had paid twice.

This is what made the reporting figures disagree with the Z-ticket for the same period: the Z-ticket has always summed only the payments that went through, while reporting summed every attempt.

Counts, totals and CSV exports now cover only the payments that actually went through.

Filtering also used to mix two separate questions under one set of boxes: which orders you want to see, and which payments you want counted under them. They are now asked separately — **Order status** and **Carrying a payment** decide which orders appear, **Payments shown** decides what is listed and totalled underneath.

### DIFF
--- a/src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.svelte
+++ b/src/routes/(app)/admin[[hash=admin_hash]]/reporting/+page.svelte
@@ -21,10 +21,34 @@
 	let tableDeliveryFeesSynthesis: HTMLTableElement;
 	let tableVATSynthesis: HTMLTableElement;
 
-	let includePending = false;
-	let includeExpired = false;
-	let includeCanceled = false;
-	let includePartiallyPaid = false;
+	// Two questions the four checkboxes used to answer at once, now asked apart: which orders
+	// the reporting is about, and which of their payments it shows. The totals follow the
+	// second, so they always count exactly what is on screen.
+	const ORDER_STATUSES = ['paid', 'pending', 'expired', 'canceled', 'failed'] as const;
+
+	// Search — which orders enter every table.
+	let orderStatuses = new Set<string>(['paid']);
+	// Search — narrows to orders carrying such a payment. All ticked means no narrowing, which
+	// is why "partially paid" is a pending order carrying a paid payment rather than a case of
+	// its own.
+	let carryingPaymentStatuses = new Set<string>(ORDER_STATUSES);
+	// Display — which payment lines the payment table draws, and therefore what it totals.
+	let shownPaymentStatuses = new Set<string>(['paid']);
+
+	// Every table starts open: a reporting that hides itself on arrival would puzzle the people
+	// who use it today. Nothing is remembered between visits.
+	let collapsed = new Set<string>();
+
+	function toggleStatus(set: Set<string>, status: string) {
+		const next = new Set(set);
+		if (next.has(status)) {
+			next.delete(status);
+		} else {
+			next.add(status);
+		}
+		loadedHtml = false;
+		return next;
+	}
 	let filterByTag = !!data.tagId;
 	let selectedPaymentMethod = data.paymentMethod ?? '';
 	let html = '';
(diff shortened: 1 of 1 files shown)


Tirodem:
QA OK :heavy_check_mark:

slashbinslashnoname (changes requested):
The body says an all-expired order "still appears in Payment Detail, but with no payment line under it". It doesn't — the `<tr>` is inside the inner `{#each}`, so no payment means no row and the order drops out of the table. "Include expired orders" now does nothing there.

Tirodem:
@slashbinslashnoname thanks to your warning in your comment, I came with another idea, an evolution of filters of both orders & payments to provide a better reporting experience and to provide an initial view that won't be confusing for the final user. New code on this same PR incoming.

slashbinslashnoname (changes requested):
Telling the two questions apart is the right move — this reads much better than four
checkboxes that each meant something different.

Two things to fix.

**An order with no payment is now unreachable.** `orderFiltered` requires
`order.payments.some(...)`, and `.some()` on an empty array is `false`. `createOrder`
inserts the order first and only adds a payment `if (paymentMethod)` (`orders.ts:1819`) —
`null` being the documented point-of-sale multi-method case. An order cancelled or expired
before its first payment was created therefore drops out of every table whatever is ticked.
The help text under the boxes says "Leave every payment status ticked to keep all orders",
which is exactly the case that does not hold.

**Collapsing a table throws away the prepared PDF.** `toggleStatus` sets
`loadedHtml = false`, and all eight collapse buttons go through it. Folding "Order detail"
to reach the syntheses below means preparing the PDF again. The reset belongs to the filter
toggles, not to the display ones.

Two smaller ones.

`paidOrders = orderFiltered` keeps a name that is no longer true, and it is read fourteen
times — `orderSynthesis.count`, the CSV exports, `quantityOfProduct(paidOrders, …)` at :978.

`ORDER_STATUSES` re-lists what `ORDER_PAYMENT_STATUSES` already holds (`types/Order.ts:20`);
orders and payments share that enum. Importing it means a new status appears in the filters
instead of quietly missing from them.

Last: the description still describes the previous approach, including the sentence about an
all-expired order showing with no payment line. Worth rewriting to match what the PR now does.

PR #2739, opened 2026-08-25
🐛🤖 Cancel pending payments left on fully paid orders (#2738)

### LINKED ISSUE #2738: 🐛 Pending payments survive forever on fully paid orders
## Overview

Once an order reaches `paid`, any payment still `pending` on it stays pending for good. Nothing expires it, nothing cancels it.

Two things combine:

**The maintenance loop never looks at it again.** `maintainOrders` in `src/lib/server/locks/order-lock.ts` selects orders with:

```js
{ 'payments.status': { $in: ['pending', 'failed'] }, status: 'pending' }
```

It requires the **order** to be pending. The moment the order becomes `paid` it leaves that scope, so its stale attempts are never checked and never expired.

**Nothing cleans them up at the transition.** In `src/lib/server/orders.ts`, `onOrderPayment` marks the successful payment as paid, logs `paymentDone` and then `orderPaid`, and leaves the other payments of the order untouched.

## How an order ends up in that state

The order page offers **Skip for now** when the last payment is pending (`skipMode` in `src/routes/(app)/order/[id]/+page.svelte`). The button navigates back to the tab and does not touch the payment — which is what its name says. A second payment then settles the order, and the first attempt is stranded.

**Cancel** takes the other path: it calls `onOrderPaymentFailed(..., 'canceled')`, which is why cancelled attempts behave correctly and never showed this problem.

## Why it matters

The data in the database is wrong: an order is settled, yet it carries a payment that claims to be awaiting money nobody owes. #2730 stops the reporting dashboard from counting those attempts, but it only hides them — the documents stay.

Worth noting for anyone weighing the risk: because the maintenance loop already ignores paid orders, such a payment is **not** being watched today either. Money arriving on it would not be captured as things stand. Marking it cancelled therefore changes nothing operationally — it makes the record say what is already true.

## Expected behaviour

When an order becomes fully paid, every payment still `pending` on it is marked `canceled`, inside the same transaction as the payment that settled it.

Nothing is owed anymore, so no attempt on that order should be allowed to go through: a card transaction still live on the terminal must not be captured, and a skipped payment call must not linger.

The order status cascade in `onOrderPaymentFailed` is already guarded — it only changes the order status when every payment is cancelled or expired **and** the order is still `pending` — so a paid order keeps its status.

## Known limits, to decide separately

- **No processor-side cancellation.** `onOrderPaymentFailed` only writes to MongoDB. A Lightning invoice stays payable and a terminal stays armed. Cancelling at the processor would need SDK support that does not exist today — there is no `cancelInvoice` anywhere in the tree.
- **Existing data.** The change would only cover orders settled from then on. Orders already in the database keep their stranded pending payments; cleaning those needs a migration.
- Overpayment **within a single payment** is a different subject, tracked in #174.

Tirodem:
When an order is settled, any payment attempt still waiting on it is now closed.

Until now those attempts stayed open for ever. A waiter skipping a payment call instead of cancelling it, a card transaction left hanging on the terminal — the order was then settled another way, and the abandoned attempt kept sitting there, waiting for money nobody owed any more. It stayed visible next to the real payment in the order, and it was the source of the reporting discrepancy handled in #2730.

Deploying this version also closes the attempts already stranded on orders settled in the past, in a single pass at startup.

Fixes #2738

### DIFF
--- a/src/lib/server/migrations.ts
+++ b/src/lib/server/migrations.ts
@@ -977,6 +977,26 @@ export const migrations = [
 				);
 			}
 		}
+	},
+	{
+		_id: new ObjectId('000000000000000000002738'),
+		name: 'Cancel pending payments left on fully paid orders (issue #2738)',
+		run: async (session: ClientSession) => {
+			// Imported here and not at the top: runtime-config imports this module, so a
+			// static import of orders.ts would close the cycle at load time.
+			const { onOrderPaymentFailed } = await import('./orders');
+			for await (let order of collections.orders.find(
+				{ status: 'paid', 'payments.status': 'pending' },
+				{ session }
+			)) {
+				for (const { _id } of order.payments.filter((p) => p.status === 'pending')) {
+					const stale = order.payments.find((p) => p._id.equals(_id));
+					if (stale) {
+						order = await onOrderPaymentFailed(order, stale, 'canceled', { session });
+					}
+				}
+			}
+		}
 	}
 ];
 
--- a/src/lib/server/orders.ts
+++ b/src/lib/server/orders.ts
@@ -407,6 +407,15 @@ export async function onOrderPayment(
 
 		order = ret.value;
 		if (order.status === 'paid') {
+			// Nothing is owed anymore: no other attempt on this order may go through.
+			// `order` is replaced on each pass, so the payment is looked up again —
+			// `onOrderPaymentFailed` rejects a payment absent from the order it is given.
+			for (const { _id } of order.payments.filter((p) => p.status === 'pending')) {
+				const stale = order.payments.find((p) => p._id.equals(_id));
+				if (stale) {
+					order = await onOrderPaymentFailed(order, stale, 'canceled', { session });
+				}
+			}
 			await updateAfterOrderPaid(order, session);
 		}
 


Tirodem:
QA done :heavy_check_mark:

slashbinslashnoname (changes requested):
The runtime half is right. I checked the thing that worried me most.
`onOrderPaymentFailed` also refunds free-product units and cancels `scheduleEvents`, but
neither can fire on a paid order: both cascades need the order status to change, and it does
not. So replaying this over history only flips payment statuses. The settling payment is
already `paid` in the re-read document, so it cannot cancel itself.

Things :

**The migration writes without bound inside one transaction.** `runMigrations` wraps
`migration.run(session)` in `withTransaction`, and this iterates every paid order carrying a
pending payment, writing each one. Past MongoDB's 16MB oplog-entry limit or the 60s
`transactionLifetimeLimitSeconds`, the transaction aborts, `withTransaction` retries to its
own cap, `runMigrations` rejects, `refreshPromise` rejects, and the app does not boot. A retry
restarts from zero. The read is fine (`{ status: 1, 'payments.status': 1 }` covers it); the
writes are the problem. Batching outside the transaction, or committing per order, removes
the cliff.

Quality: the snapshot-then-relookup dance is correct and earns its comment, but it is copied
verbatim in both places, comment included. `cancelStalePendingPayments(order, session)` would
give the invariant one name and one home. Minor: the migration `_id` is hand-written
(`000000000000000000002738`) where the others are random.

PR #2745, opened 2026-09-04
🆕🤖 Set a product variation from the URL, and hide it from the page (#2688)

### LINKED ISSUE #2688: 🆕🔌 Add unique ID variation on product page & checkout

Tirodem:
A product sold in several variations could only be bought by picking each one from a dropdown. Handing someone a link that already carries the choice was not possible, and neither was selling something whose variation nobody should be able to pick from a list — a numbered seat, a bracelet, an artifact with its own identifier.

A variation can now be set from the address: /product/artifact?color=red, or several at once with ?color=red&size=xxl. A variation settled that way loses its dropdown — the choice is already made — and follows the purchase through to the cart and the order exactly as if it had been picked by hand.

Two settings sit under the variations block on the product page, one line per family:

- Hide from the product page. The family never appears as a dropdown, so the link is the only way to set it. Someone landing on the bare product address cannot pick a value, and cannot guess one from a list.
- Hide from cart and order. The value is recorded but never shown back to the customer, on any page.

A link that asks for something the product does not have — an unknown family, an unknown value, or nothing at all for a family that has no dropdown — refuses the page and says why, rather than quietly falling back to some other value. A product can be set to ignore the bad parameter and show its dropdowns instead. Parameters that name no variation at all, such as campaign tags, are left alone.

Each variation row carries a link button that copies the address landing on the product with that variation already chosen. The value a link has to carry is not always the label that was typed — a purely numeric value is prefixed with its family name, so typing 10 on a "load" family stores load10 — and nothing in the form showed it, which left the link to be guessed.

Fixes #2688

### DIFF
--- a/src/lib/components/OrderSummary.svelte
+++ b/src/lib/components/OrderSummary.svelte
@@ -9,7 +9,7 @@
 		orderItemPrice
 	} from '$lib/types/Order';
 	import type { Picture } from '$lib/types/Picture';
-	import type { Product } from '$lib/types/Product';
+	import { productLabelWithVariations, type Product } from '$lib/types/Product';
 	import PictureComponent from './Picture.svelte';
 	import PriceTag from './PriceTag.svelte';
 	import ProductType from './ProductType.svelte';
@@ -94,13 +94,7 @@
 	{#each order.items as item}
 		<a href="/product/{item.product._id}">
 			<h3 class="text-base">
-				{item.chosenVariations
-					? item.product.name +
-					  ' - ' +
-					  Object.entries(item.chosenVariations)
-							.map(([key, value]) => item.product.variationLabels?.values[key][value])
-							.join(' - ')
-					: item.product.name}
+				{productLabelWithVariations(item.product, item.chosenVariations)}
 			</h3>
 		</a>
 
--- a/src/lib/components/ProductAddedToCart.svelte
+++ b/src/lib/components/ProductAddedToCart.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import type { BasicProductFrontend } from '$lib/types/Product';
+	import { productLabelWithVariations, type BasicProductFrontend } from '$lib/types/Product';
 	import type { Picture as PictureType } from '$lib/types/Picture';
 	import Picture from './Picture.svelte';
 	import PriceTag from './PriceTag.svelte';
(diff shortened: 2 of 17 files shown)


slashbinslashnoname (changes requested):
As a feature of its own — set a variation from a link, hide a family from the page — this is well built and worth keeping. My doubt is whether it can carry #2688.

#2688 asks for a secret ID per artifact: *"I don't want to allow for anyone to access artifact from another."* This PR expresses that key as a variation value, and a variation value is a public enumeration entry. `product/[id]/+page.server.ts` returns `product: { ...product }` with `variations` projected in, so every declared key is in the page payload — open the product page, read the source, you have them all. `hiddenFromUI` removes the dropdown, not the data. On top of that, nothing can mark a key as sold (stock is per product, not per value), and one row per artifact in a single product document doesn't scale.

That key is an entitlement token, not a variation: its own collection, a sold/unsold state, resolved server-side, never sent to the client. I'd drop the `Fixes #2688` and let this PR be the generic feature it is.

Five things in the code meanwhile:

- `admin/product/new/+page.server.ts` never reads `variationFamilies` or `variationUrlPolicy`, in either path. The block does render on the new-product form, so the setting is silently dropped at creation.
- `hiddenFromUI` + `variationUrlPolicy: 'ignore'` posts a hidden input with `undefined` → empty string, which `checkProductVariationsIntegrity` accepts, and the product enters the cart with a blank variation at base price — the case the function's own comment says it prevents.
- The URL validates values, the cart POST doesn't: `checkProductVariationsIntegrity` (`Product.ts:215-226`) only checks family names. Predates this PR, but this PR makes it load-bearing.
- The 400 at `+page.server.ts:247` fires before the visibility redirect at `:257`, so a crafted `?family=bogus` confirms a hidden product's families.
- `unknownFamily` is unreachable — `cleanVariationFamilies` rules out the case it needs. An unknown family is silently ignored, so the body overstates it.

PR #2747, opened 2026-09-04
🐛🤖 Give every concurrent settlement its own invoice number (#2743)

### LINKED ISSUE #2743: 🐛 Concurrent payment settlements lose all but one invoice number
## Symptom

**Given** a shop recording sales in be-BOP  
**When** several payments are settled at the same moment  
**Then** each of them is recorded as paid, and each receives its own invoice number

**Instead**, one payment is recorded as paid and the others are lost: those orders
stay pending with an amount paid of zero, no invoice is issued for them, and
nothing in the response tells the caller.

A person at a till cannot settle two sales in the same instant, so this does not
appear in manual use. An automated flow of payments does.

## Cause

Invoice numbers are obtained by reading the highest number already stored on an
order and adding one, in `onOrderPayment` (`src/lib/server/orders.ts`). The
sequence has no home of its own: the `orders` collection is both the record of
issued invoices and the counter that issues them. Two settlements reading at the
same moment compute the same number, and the unique index on that field lets
only one of them through.

Order numbers work the other way — a dedicated `runtimeConfig` entry with an
atomic increment (`generateOrderNumber`, same file).

Not introduced by a recent change: the same code is on `main`.

## To settle when fixing

- Seeding the counter above the highest number already issued.
- Whether the sequence may skip a number, which an atomic counter allows.
- The number is held by the payment rather than by the order, so a sale settled
  in several instalments consumes several invoice numbers.

Tirodem:
Settling several payments at the same moment recorded one of them and lost the others: those orders stayed pending, showing nothing paid and carrying no invoice, while the caller was told the order had been created. A person at a till never hits this — two sales cannot be settled in the same instant — but an automated flow of payments does, and a shop taking payments that way was short by the lost sales with no sign of it anywhere.

Invoice numbers were worked out by reading back the highest one already on an order and adding one. Settlements landing together read the same highest number and computed the same next one, and only one of them could be stored. They now come from a counter of their own, which cannot hand the same number to two callers, the same way order numbers already worked.

Shops that have already issued invoices keep their numbering: the counter starts past the highest number given out, read once when the shop next starts.

Fixes #2743

### DIFF
--- a/src/lib/server/migrations.ts
+++ b/src/lib/server/migrations.ts
@@ -988,6 +988,40 @@ export const migrations = [
 				);
 			}
 		}
+	},
+	{
+		_id: new ObjectId('68b9a4c17d2f4e0a3c5b1d90'),
+		name: 'Seed the invoice number counter from the invoices already issued',
+		run: async (session: ClientSession) => {
+			// Invoice numbers used to be derived by reading the highest one already stored (#2743).
+			// The counter has to start past that, or the first invoice issued after this migration
+			// would reuse a number the shop has already given out.
+			const highest = await collections.orders
+				.aggregate<{ payments: Array<{ invoice?: { number: number } }> }>(
+					[
+						{ $match: { 'payments.invoice.number': { $exists: true } } },
+						{ $sort: { 'payments.invoice.number': -1 } },
+						{ $limit: 1 },
+						{ $project: { 'payments.invoice.number': 1 } }
+					],
+					{ session }
+				)
+				.next();
+
+			const highestNumber = Math.max(
+				0,
+				...(highest?.payments ?? []).map((payment) => payment.invoice?.number ?? 0)
+			);
+
+			await collections.runtimeConfig.updateOne(
+				{ _id: 'invoiceNumber' },
+				{
+					$set: { data: highestNumber as never, updatedAt: new Date() },
+					$setOnInsert: { createdAt: new Date() }
+				},
+				{ upsert: true, session }
+			);
+		}
 	}
 ];
 
--- a/src/lib/server/order.test.ts
+++ b/src/lib/server/order.test.ts
@@ -242,6 +242,43 @@ describe('order', () => {
 		expect(order1?.payments[1].invoice?.number).toBe(3);
 	});
 
+	it('gives every concurrent settlement its own invoice number', async () => {
+		// #2743: invoice numbers were derived by reading back the highest one already stored, so
+		// settlements landing together computed the same next number. The unique index let one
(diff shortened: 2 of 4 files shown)


slashbinslashnoname (changes requested):
The atomic counter is the right remedy: it mirrors `generateOrderNumber` exactly, the boot ordering is sound, and the new test would fail on the old code. Four things before it goes in.

**1 — The sequence can now skip numbers.** `generateInvoiceNumber` runs at the top of `onOrderPayment`, before the `Sync broken between order and payment` guard and, without `providedSession`, outside the `withTransaction(fn)` that stores it. Anything throwing in between burns a number for good — and `maintainOrders` re-polls the same pending payment every 2s while `handleTapToPayCheck` swallows its errors, so one transient failure burns one number per poll. It is inconsistent too: with `providedSession` the `$inc` joins the caller's transaction and does roll back. The old code could not leave a hole at all. Moving the allocation inside `fn`, after the guard, closes both paths. #2743 listed this under "to settle when fixing" and it is still open.

**2 — The invariant is established once and never re-checked.** `counter ≥ highest issued` holds only because the migration ran. `runMigrations` marks every migration done *without running it* when the `migrations` collection is empty (`migrations.ts:1040`), so a database that kept its orders but lost its migration records keeps its invoices and loses the seed. And a downgrade/upgrade round-trip leaves the counter behind with nothing to replay it. Both bring #2743 back with no signal. Should the seed be a startup check rather than a one-shot?

**3 — `$set: { data: highestNumber }` can rewind the counter.** If the document already holds more, this lowers it and the next settlements collide on the unique sparse index — the very failure being fixed. `$max` is equally correct and never moves backwards.

**4 — The test leans on a favourable interleaving.** It creates 8 concurrent orders of `TEST_DIGITAL_PRODUCT`, stock 5, and `amountOfStockReserved` counts pending orders; `checkCartItems` runs outside `createOrder`'s transaction. I could not trigger it in 30 runs, so fragility rather than defect — `TEST_DIGITAL_PRODUCT_UNLIMITED` is already imported here and has no stock. Separately, `runMigrations` returns early under `VITEST`, so the seeding path is exercised nowhere.

Minor: the migration re-implements `lastInvoiceNumber` instead of calling it, and that function's new comment claims the migration asks for it — it does not. And #2743's third point, instalments consuming several numbers, is left as it was; probably right, but unstated.

Checked because it reads wrong and is not: the `$sort` on `payments.invoice.number` **is** index-served. Mongo 6.0.9, 20k orders — backward `IXSCAN`, 1 document, 1 key, 1 ms, no `SORT` stage.

PR #2748, opened 2026-09-07
🐛🤖 Honour "Disable VAT for my be-BOP" everywhere a rate is shown (#2649)

### LINKED ISSUE #2649: 🐛  Fix VAT calculator ignoring "Disable VAT" global setting on product creation
## Overview
When **Disable VAT for my be-BOP** is checked in global settings, new products still default to a non-zero VAT profile and rate (ex `8,1%`, might be tied to admin IP or the hosting country) instead of `0%`. Admins have to manually create then select a `zero` VAT custom rate on each product to actually apply the VAT-free setting.

Steps to reproduce:
1. Enable **Disable VAT for my be-BOP** in global settings
2. Create a new product
3. Observe the default **VAT profile** is `No custom VAT profile` and **VAT rate** is `8,1%`, not `0%`

Image

Image

Image

## Acceptance criteria
- [ ] When VAT is disabled globally, new products default to `zero` VAT profile (`0%` rate) automatically
- [ ] Admins no longer need to manually create/select a `zero` VAT custom rate to get VAT-free pricing
- [ ] Existing products with a manually-set VAT profile are not altered by this fix
- [ ] No regression on VAT calculation when the global setting is disabled

Tirodem:
A shop that turned VAT off was still shown its country's rate. Creating a product on a VAT-free shop displayed 8.1%, and the only way to get a VAT-free product was to build a custom profile with a zero rate and pick it by hand, on every product.

Worse than the wrong figure: the product form uses that rate to work back from a VAT-included price. A shop typing what it actually charges had the amount divided by a tax it does not collect, and the price recorded — and later billed — came out lower than the one entered.

Orders were already priced VAT-free; it was every displayed rate that ignored the setting. The exemption is now settled once, at the top of the single function every rate goes through, so the product page, the catalogue listings, the touch till, the checkout and the delivery fees all agree — and a fee added later inherits it without anyone having to remember.

Nothing already recorded is altered: a product carrying a VAT profile keeps it, and a shop that has not turned VAT off is priced exactly as before.

Fixes #2649

### DIFF
--- a/src/lib/cart.ts
+++ b/src/lib/cart.ts
@@ -155,7 +155,8 @@ function computeVatForItem(
 		vatProfiles: params.vatProfiles,
 		bebopCountry: params.bebopCountry,
 		userCountry: params.userCountry,
-		vatSingleCountry: params.vatSingleCountry
+		vatSingleCountry: params.vatSingleCountry,
+		vatExempted: params.vatExempted
 	});
 	const freeUnits = params.freeUnits ?? 0;
 	const { amount: amountToBill, currency, usedFreeUnits } = priceToBillForItem(item, { freeUnits });
@@ -275,6 +276,7 @@ function computeDeliveryFeesHt(params: {
 	bebopCountry: CountryAlpha2 | undefined;
 	userCountry: CountryAlpha2 | undefined;
 	vatSingleCountry: boolean;
+	vatExempted?: boolean;
 	vatProfiles: Array<{
 		_id: string | ObjectId;
 		rates: Partial<Record<CountryAlpha2, number>>;
@@ -288,7 +290,8 @@ function computeDeliveryFeesHt(params: {
 		vatProfiles: params.vatProfiles,
 		bebopCountry: params.bebopCountry,
 		userCountry: params.userCountry,
-		vatSingleCountry: params.vatSingleCountry
+		vatSingleCountry: params.vatSingleCountry,
+		vatExempted: params.vatExempted
 	});
 	return {
 		amount: extractVat(params.deliveryFees.amount, rate),
@@ -382,6 +385,7 @@ export function computePriceInfo(
 		bebopCountry: params.bebopCountry,
 		userCountry: params.userCountry,
 		vatSingleCountry: params.vatSingleCountry,
+		vatExempted: params.vatExempted,
 		vatProfiles: params.vatProfiles
 	});
 
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -174,7 +174,8 @@
 		vatProfiles,
 		bebopCountry: $page.data.vatCountry,
 		userCountry: $page.data.vatCountry,
-		vatSingleCountry: true
+		vatSingleCountry: true,
+		vatExempted: $page.data.vatExempted
 	});
 	$: vatProfileLabel =
 		vatProfiles.find((p) => p._id === vatProfileId)?.name ?? 'No custom VAT profile';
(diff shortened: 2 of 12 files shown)


slashbinslashnoname (changes requested):
Settling the exemption once at the top of `computeVatRate` is the right place, and all ten call sites of that function were updated. But moving *one* of three exemptions into the rate, and leaving the other two where they were, splits the concept in half.

**"Exempt" now has two mechanisms.** `computePriceInfo` knows three exemptions: `vatExempted` (shop), `vatNullOutsideSellerCountry` → `isPhysicalVatExempted` (customer outside the seller country), and `reasonFreeVat` (order, folded into `vatExempted` by `orders.ts:896`). Before this PR all three were handled outside the rate, by the guard at `cart.ts:144`. After it, the shop one is carried *by the rate* and the others stay carried *by a guard*.

`computeDeliveryFeesHt` only consumes the rate. So it inherits the shop exemption and stays blind to the other two — which is the bug below, and will be the bug again for whatever fee is added next.

**Delivery fees are billed higher for VAT-exempt orders.** `cart.ts:388` forwards `vatExempted` into `computeDeliveryFeesHt`, which uses the rate to walk a VAT-included fee back to HT. A fee set to 120 TTC in a 20% shop returned `extractVat(120, 20)` = 100; at rate 0 it returns 120. A POS order ticked "VAT free" therefore bills 120, i.e. the customer pays the VAT amount as delivery — while `vatNullOutsideSellerCountry`, the sibling exemption, bypasses the rate and still yields 100 for the same customer. Separately, a shop already running VAT-off with `vatIncludedReference` goes from 100 to 120 on deploy; possibly the correct price, but "Nothing already recorded is altered" does not cover it.

**The touch till does not agree with itself.** `pos/touch/tab/[orderTabSlug]/ticket/+page.server.ts:43` has a local `getVatRate` that re-implements `computeVatRate` and never reads `runtimeConfig.vatExempted`. CH shop, VAT off, one 100 CHF item: the till shows 100, the order bills 100, the ticket prints an 8.1% breakdown and a total of 108.10.

**The optional parameter undoes the guarantee the comment makes.** `vat.ts:25` promises a new fee "inherits it without anyone remembering to ask", but `vatExempted?: boolean` lets a caller omit it and get the country rate. Same at `cart.ts:279`. Making both required is one character, and would have failed the build on the ticket page rather than leaving it to review.

**Display surfaces still contradict each other.** `ProductWidgetVariation0..4` print "VAT excluded" unconditionally, `cart/+page.svelte:396` renders "VAT 0%", and `Searchlist.svelte:30` labels the price filter "EUR TTC". Meanwhile `searchlist/[slug]/+page.server.ts:101` adds `vatExempted` that nothing reads, while `+layout.server.ts:323` already exposes it.

Smaller. One of the ten wirings is dead: `cart.ts:159` passes `vatExempted` to `computeVatRate` inside `computeVatForItem`, but the guard at `:144` already returned when the flag is true, so it can only ever be `false` there. `VatContext` is built identically three times (`cms.ts:857`, `searchlist/[slug]/+page.server.ts:38` and `:101`). `product/[id]/+page.svelte:645` assigns `showExclTax = !data.vatExempted && !showExclTax` on a div that keeps its click handler while losing its cursor. And the three exemptions read as three unrelated ideas — `vatExempted`, `isPhysicalVatExempted`, `vatNullOutsideSellerCountry` — a past participle, an `is`-prefixed derived state, and a rule name, which is most of why they are easy to mix up.

On tests: `vat.test.ts` covers the pure function, where the new logic is three obvious lines. The wiring, where the risk lives, is untested. A `computePriceInfo` test with an exempt shop, asserting the item rate *and* the delivery rate at zero, would have caught the first issue.

PR #2749, opened 2026-09-07
🦋🤖 Let a customer subscribe and buy the discounted product in one order (#2718)

### LINKED ISSUE #2718: 🦋 Add subscription products to discount v2 required-product combinations
## Overview
Discount v2 already lets admins require specific products in the cart to unlock a discount (ex: Product X or Y), including membership-style discounts that require an existing active subscription. Subscription products can't currently be selected as one of the required products themselves. This forces a two-step flow for new customers: subscribe, log back in, then buy the discounted product, instead of letting them subscribe and buy the discounted product in the same order.

Current flow:
1. Buy subscription
2. Log in
3. Buy Product A with discount

Desired flow:
1. Buy subscription and Product A together, discount applied in the same order

Tirodem:
A discount reserved for members only recognised a subscription that had already been paid for. A new customer had to order the membership, wait, come back logged in, and place a second order to get the discount — three steps to buy one thing.

Two changes make the single order possible.

A discount that requires certain products in the cart can now list a subscription among them. The picker excluded subscriptions because it was the same list used to choose what gets discounted, where a subscription has no place; requiring one as a condition is a different question, and it now has its own list.

And a members-only discount can accept the membership sitting in the cart rather than an already active one. That is a checkbox on the discount, off unless the shop turns it on: an existing members-only discount means paid-up members, and turning that on for every shop at once would start granting discounts on a membership nobody has paid for yet.

Nothing changes for discounts already configured.

Fixes #2718

### DIFF
--- a/src/lib/server/discount.test.ts
+++ b/src/lib/server/discount.test.ts
@@ -1,5 +1,10 @@
 import { describe, expect, it } from 'vitest';
-import { isPublicZeroCriteriaDiscount, publicDiscountPriceSnapshot } from './discount';
+import {
+	evaluateDiscountConditions,
+	isPublicZeroCriteriaDiscount,
+	publicDiscountPriceSnapshot
+} from './discount';
+import type { DiscountContext } from './discount';
 import type { Discount } from '$lib/types/Discount';
 
 /** Minimal valid public percentage discount; override any field per test. */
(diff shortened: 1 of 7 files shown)


slashbinslashnoname (changes requested):
The checkbox sits outside `{#if mode === 'percentage'}` in both forms, so it also shows on a `freeProducts` discount — and `baseData` is spread into that branch's `$set` too, so the value is stored. But `freeProducts` never reaches `evaluateDiscountConditions`: `freeProductsForUser` reads `paidSubscriptions` directly. The control does nothing there and nothing says so. It belongs inside the block, next to `promoCode`.

One question: on the product page the discount won't show. `product/[id]/+page.server.ts:65-72` evaluates against a fake one-line cart, so `isSubscribingNow` is only true when the viewed product *is* the subscription — and the path above it (`:43-53`) queries `subscriptionIds` in Mongo directly, bypassing `evaluateDiscountConditions` altogether. A customer with the membership in their cart sees full price on the product sheet, and the discount only appears in the cart. Intended?

Minor: `usedSubIds` (`orders.ts:923`) comes from `paidSubs`, so orders discounted through the cart subscription lose the "because of subscription …" justification in their note.

PR #2757, opened 2026-09-07
🆕🤖 Restrict a product to a whitelist of customers (#2143)

### LINKED ISSUE #2143: 🆕 Whitelist-based products
Product that only selected e-mail addresses and npubs can added them to cart.

Manual list + active subscribers from one or multiple subscription products

TBD

Tirodem:
A product can now be reserved for a chosen audience. Its admin page carries a Whitelist section, off by default, where the shop owner lists who may order it.

Four sources, and matching a single one is enough: a free text area of e-mail addresses, one per line; a free text area of npubs, one per line; a pick of subscription products, whose active subscribers are all allowed without being listed by hand; and a checkbox for every employee account, whatever its role.

A visitor who matches none of them still reaches the product page and reads it in full, but the order and add-to-cart buttons are replaced by a message saying the product is reserved for selected customers. In product widgets, tag widgets and search lists, the button takes them to the product page rather than adding the product to their cart.

The restriction is enforced on every way into the cart, not only on the buttons: a pre-configured cart link, the point-of-sale alias field and NostR all go through it. A last option lets a point-of-sale employee ring the product up for a customer who is not on the list, for shops where the counter decides.

Products with no whitelist are untouched and stay open to everyone.

Fixes #2143

### DIFF
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -63,6 +63,7 @@
 	export let defaultActionSettings: ProductActionSettings;
 	export let availablePaymentMethods: PaymentMethod[];
 	export let productsWithStock: { _id: string; name: string }[] = [];
+	export let subscriptionProducts: { _id: string; name: string }[] = [];
 	export let currentBookings: BookingSummary[] = [];
 	export let upcomingBookings: BookingSummary[] = [];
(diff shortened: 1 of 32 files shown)


Tirodem:

Tirodem:
Replaced by #2763.

PR #2758, opened 2026-09-07
🆕🤖 Add new product property "Maximum quantity for an unique user" (#2…

### LINKED ISSUE #2756: 🆕 Add new property to product "Maximum quantity for an unique user"
## The need

A product can already cap how many units go into a **single order** (`Maximum quantity per order`). Nothing caps how many units the **same person** can buy across several orders: they order 3, come back, order 3 more.

We want a new product property, "Maximum quantity for an unique user": once a customer has reached that many units of the product — over their history, not per order — "add to cart" and "order" are refused for them.

Value on its own: limited editions, fair distribution of a scarce product, one batch of tickets per household, a member benefit capped per person.

## What already exists, and why it is not enough

be-BOP already restricts a purchase per person in exactly one place: **subscriptions**. A customer cannot buy a second subscription of the same product while the first one is still running, and cannot open a second order for it while a first one is still awaiting payment.

That restriction gets two things right:

- **Who "an unique user" is** is already defined, and it is not only a logged-in account: it matches on user id, contact e-mail (including secondary ones), npub, SSO identity, or browser session.
- **Orders awaiting payment count too.** Otherwise a customer opens two orders side by side, pays both, and goes over the cap.

And one thing wrong — which is the opportunity below.

## UX opportunity

The subscription restriction refuses **at checkout, not at add to cart**. The customer fills their cart, enters their contact details, reaches the payment step, and only there gets a bare error page carrying a hardcoded English sentence that is neither translated nor the shop's own wording. Nothing on the product page had told them.

Whatever we build here should instead be checked **at add to cart** and surfaced as a cart error, the way `Maximum quantity per order` already is — and on the product page, the CTA should be disabled with a message rather than fail after the click.

The same treatment applied to the subscription restriction would fix it at the same time: "you already have an active subscription for this product" and "you already have a pending order for this product" would become translated cart errors instead of error pages.

Paths that must go through the same check, or the cap is bypassable: the product widget CTA, the pre-configured cart / share link, and the PoS "add by alias" — with, as in #2143, the question of whether a PoS employee may override, which should be a setting rather than a hard rule.

Tirodem:
…756)

A product could already cap how many units go into a single order. Nothing capped how many units the same person could buy across several orders: they ordered three, came back, ordered three more.

Products now carry a "Maximum quantity for an unique user" setting, next to the per-order cap in the admin form. Left empty, nothing changes. Filled in, a customer who has already reached that many units is refused when they add the product to their cart, and the product page shows them why with the buy button greyed out instead of letting them click and fail later.

Who counts as the same person is what the shop already knows about them: their account, their contact e-mail and its secondary addresses, their npub, their SSO identity, or their browser session. No account is needed for the cap to hold. Orders still awaiting payment count towards it, so opening two orders side by side and paying both does not get anyone over the limit.

Subscriptions were the one place where be-BOP already limited a purchase to one per person, and they keep that limit unchanged. What they gain is the same front behaviour as everything else: a customer who already holds an active subscription, or who has an unpaid order for one, now sees it on the product page and is turned away at add to cart, rather than reaching the payment step and hitting a bare error page in English. The message is translated in all seven languages.

Every way into the cart goes through the same refusal: the product page, the product widgets, a pre-configured cart or share link, and the point-of-sale "add by alias".

Fixes #2756

### DIFF
--- a/docs/fr/WIP-product-management.md
+++ b/docs/fr/WIP-product-management.md
@@ -58,6 +58,10 @@ Accessible via **Admin** > **Merch** > **Product**, cette section vous permet d'
   - [ ] **Enable preorders before available date** : Autorisez les précommandes.
   - [ ] **Display custom text instead of date for preorder** : Afficher un autre text à la place de la date de disponibilité.
 - **Max quantity per order** : Limite le nombre d’unités par commande.
+- **Maximum quantity for an unique user** : Limite le nombre d’unités qu’une même personne peut
+  acheter, toutes commandes confondues. Laissez vide pour ne pas plafonner. Les commandes en
+  attente de paiement sont comptées. Un abonnement est toujours limité à un par personne et par
+  produit.
 
   ![image](https://github.com/user-attachments/assets/146c241f-292f-4ef2-87a4-0a852cffbb58)
 
--- a/src/lib/components/ProductForm.svelte
+++ b/src/lib/components/ProductForm.svelte
@@ -8,6 +8,7 @@
 	import {
 		DEFAULT_MAX_QUANTITY_PER_ORDER,
 		MAX_NAME_LIMIT,
+		MAX_QUANTITY_PER_USER_FOR_SUBSCRIPTION,
 		MAX_SHORT_DESCRIPTION_LIMIT,
 		type Product
 	} from '$lib/types/Product';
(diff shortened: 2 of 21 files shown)


Tirodem:
Problems met along the way and what they changed in the design, written up on the issue: https://github.com/be-BOP-io-SA/be-BOP/issues/2756#issuecomment-5570143785

Tirodem:
QA ok :heavy_check_mark:

Tirodem:

Tirodem:
Replaced by #2763.

PR #2762, opened 2026-09-07
🆕🤖 Add an option to block product order if customer is not authentica…

### LINKED ISSUE #2759: 🆕 Add an option to block product order / add to cart if customer is not authentified
A per-product option to require the customer to be authenticated before they can order the product or add it to their cart. On the product page, the buy CTA is replaced by an invitation to log in.

Watch out for cart regeneration from the cookie: items added before the option was set, or before the customer logged out, must be handled.

Tirodem:
…ted (#2759)

Some products should only reach a customer the shop can recognise: a members-only item, a benefit reserved to a known buyer, anything capped or restricted per person. Until now nothing could ask for that, and an address typed into the checkout form is no answer — it is never verified, and anyone can type someone else's.

Products now carry a "Customer must be authenticated to order this product" option in the admin form. Left unticked, nothing changes.

Ticked, the customer can still fill their cart freely, but cannot order until they are logged in. The cart shows a banner naming the products concerned, with a button to the login page, and the checkout button stays out of reach. Reaching the checkout page by other means sends them back to the cart with the same banner. Nothing is created and their cart is never altered behind their back — they log in and carry on where they left off.

Being logged in means what the login page already accepts: an e-mail address, nostr, or an SSO account. A nostr session counts.

Point-of-sale is exempt. At the counter the session belongs to the seller, not to the buyer, so asking for a login there would only ever identify the wrong person.

A cart filled before the option was set, or before the customer logged out, lands in the same state as any other — the requirement is read from the product every time, never remembered from when the item was added.

Fixes #2759


Tirodem:
QA OK :heavy_check_mark: 
Following some tests, this one little rascal is required before https://github.com/be-BOP-io-SA/be-BOP/pull/2758

Tirodem:

Tirodem:
Replaced by #2763.

PR #2763, opened 2026-09-07
🆕🤖 Sales locks: whitelist, per-person cap and authentication (#2143, #2756, #2759)

### LINKED ISSUE #2143: 🆕 Whitelist-based products

### LINKED ISSUE #2756: 🆕 Add new property to product "Maximum quantity for an unique user"

### LINKED ISSUE #2759: 🆕 Add an option to block product order / add to cart if customer is not authentified

Tirodem:
Three ways to reserve a product to certain customers, built as one mechanism.

A product's Display settings gain a "Who can order" block: require the customer to be logged in (e-mail, nostr or SSO), and/or restrict ordering to a whitelist built from e-mail addresses, npubs, the active subscribers of chosen subscription products, and employees — matching one source is enough, leaving them all empty closes the product to everyone. Beside the per-order cap, "Maximum quantity for an unique user" limits how many units the same person may take across all their orders; it requires authentication by itself.

A product with no lock is untouched.

Customers are turned away before they click, not after: the product page lists every reason that applies and greys its buy button out, widgets link to the product page instead of adding to the cart, and the cart spells the reasons out and disables checkout. Refusals also happen at the till, so a cart filled before a list was drawn up cannot be paid for. Reaching the checkout with an unorderable cart sends the customer back to it with an explanation instead of an error page.

At the point of sale the session is the seller's, so buyer-specific rules do not apply there; the whitelist keeps its own per-product override.

Subscriptions stay one per person, counted as order creation already did, so renewals still go through.

Fixes #2143
Fixes #2756
Fixes #2759


Tirodem:
QA (not full, critical path) OK :heavy_check_mark:

slashbinslashnoname (changes requested):
The mechanism is the right one, and I checked the claim it rests on: every path that puts a product into a cart really does go through `addToCartInDb`, including the product page, the cart, the shared pre-configured cart and the NostR bot at `locks/handle-messages.ts:435`. The requirements from #2756 are met too: pending orders count, the identity is the subscription one, and the front greys its CTA from the same `evaluateSaleLocks` the back refuses with, so the two cannot drift.

Five things, and the first one before anything else.

The whitelist leaves the server. `product/[id]/+page.server.ts:183` adds `whitelist: 1` to the projection, and line 282 spreads the whole product into the page data. `whitelist` holds `emails` and `npubs`, so the list of authorised customers ends up in the serialised page data of a page anyone can open. Same shape at `cms.ts:644` for CMS listings and `+layout.server.ts:165` for any whitelisted product sitting in a cart. None of the three issues asked for this, and the front only needs to know that a product is locked, not who is on the list.

Stock refusals disappeared from the cart. `MAX_PER_ORDER` and `OUT_OF_STOCK` are now sale-lock codes, so `isSaleLockError` catches them, and `cart/+page.server.ts:285` discards them. The banner is built from a separate `evaluateSaleLocks` call at line 247 that omits `availableByProductId`, so it can never produce `OUT_OF_STOCK` and computes `MAX_PER_ORDER` without the stock clamp. Cart holds 3, someone else takes the stock, 1 left: no banner, `data.saleLocks` empty, checkout button enabled. The customer clicks, `checkout/+page.server.ts:55` throws, the existing catch redirects back to a cart that still says nothing. Before this PR they read "Not enough stock for product X, only 1 left".

The cart's "+" button counts the cart twice. `addToCartInDb` passes both `extraQuantityByProductId` and `quantityInCartByProductId`, so `locksFor` computes `wanted = inCart + quantity`. That is right for the additive callers, but `cart/[id]` `increase` and `setQuantity` pass `totalQuantity: true`, where `quantity` is the new total rather than a delta. Cap 10, cart at 6, click "+": the action asks for 7, `locksFor` reads 13, refuses. And `cart/[id]/+page.server.ts:81` swallows sale-lock errors and redirects, so the button does nothing at all, with no message. The legacy check at `cart.ts:154` gets this right by reading the total after mutation.

The identity match on `userId` never fires. `saleLock.ts:163` compares `stored.userId === user.userId`, but those are `ObjectId` objects, so this is reference equality and always false. The codebase uses `.equals()` everywhere else, for instance `pos/order/[id]/+page.server.ts:12`. The whitelist therefore never recognises someone by account, only by e-mail, npub or SSO, and a subscriber whose account carries neither e-mail nor npub is refused although they are entitled.

The per-person cap is advisory, and can count the wrong person. `evaluateSaleLocks` is called without a `session`, and `checkCartItems` runs at `orders.ts:859` while the order transaction only opens at `:1349`, so two checkouts side by side both read zero taken and both go through. #2756 anticipated the sequential version of this and the PR answers it by counting pending orders; the simultaneous one is still open, and "one batch of tickets per household" is exactly that scenario. Separately the count uses `userQuery(user)`, whose `$or` includes `sessionId`, and `renewSessionId` re-attributes a browser's past orders on login, so on a shared or kiosk browser customer B is refused because customer A already took the allowance. `identifiedUserQuery` is the one that fits, and authentication is forced whenever a cap is set.

Smaller. `annotateProductsWithSaleLocks` evaluates with no context, so `wanted` is zero and the test reads `taken > max` instead of `taken >= max`: a customer sitting exactly on their allowance still sees a live "Add to cart" in listings and gets a 400 on click. `cart.ts:512` derives `mode` from `user.userHasPosOptions`, an account property rather than the request channel, so an employee shopping on the ordinary e-shop is treated as a seller session while `addToCartInDb` receives the real route mode, and the two halves disagree about the same cart. Three notions of staff gate three different rules, `employee` for `LOGIN_REQUIRED`, `sellerSession` for `MAX_PER_USER` and `mode` plus `allowPosOverride` for the whitelist, which is defensible but nothing in the names says why. And `hasSaleLocks` returns true for every subscription product, so any CMS page holding one runs the full lock queries for a result that, per the point above, can never flag anything.

Tirodem:
On the identity point: a subscriber who changed their e-mail after subscribing is refused on a product reserved for active subscribers, because the subscription keeps the address it was taken with. That case has never been covered, it is wider than this PR, and it is now tracked in #2776 for the coming days.

Tirodem:
On the shared browser: signing in keeps the session of whoever used the browser before, so the previous person's orders count against the newcomer's per-person limit. The cause is wider than this PR — the same takeover carries their subscriptions and personal details too — and is now tracked in #2777.

Tirodem:
On two checkouts at the same instant: the per-person limit behaves exactly like stock and like the maximum per order — checked before the order is written, not held while it is. It is not specific to the limits added here, and is now tracked in #2778.

Tirodem:
On listings offering a locked product straight to the cart: the gap is the standalone search list in list mode on desktop, which does not carry the lock flag. On a CMS page the widgets are covered, and their CTA sends the visitor to the product page instead of adding. Fixed in this PR.

Tirodem:
New revision pushed. Six things changed.

The list of customers allowed on a restricted product no longer travels to the browser — product page, CMS pages, cart and search lists. The page learns that a product is locked, never who is on the list.

A cart holding more than the stock left says again how many remain, instead of going quiet and letting the checkout refuse a basket the cart had just declared fine.

The quantity buttons no longer count what is already in the cart twice, so raising a quantity below the limit works; and a quantity that is genuinely refused now says why instead of doing nothing.

A search list opened on its own greys out what the cart would refuse, like a CMS page does. That was the remaining gap on listings: in list mode on desktop it still offered a locked product straight to the cart.

Whether a seller may go past a per-person limit no longer depends on the screen — a shop can take payments from the ordinary cart — but on their own access to the till, and only on products where the shop ticked it. The box moves next to the limit it governs.

The counter no longer overrides the customer list: employees are let in by the list's own setting, which is what it is for.

The rest of the review is not specific to these locks and is tracked separately, to be handled in the coming days: the identity match on subscriptions in #2776, the browser session counting someone else's orders in #2777, and two orders validated at the same instant both passing a limit — stock included, long before this PR — in #2778.

slashbinslashnoname (changes requested):
The whitelist still reaches the browser — on CMS search list widgets.

`cms.ts:882` calls `searchProducts()`, whose projection includes `whitelist: 1`
(`searchlist.ts:350`), and returns those products untouched into the widget payload.
The `.map(withoutWhitelist)` at `cms.ts:824` covers the product-widget array only.

The standalone route gets it right — `searchlist/[slug]/+page.server.ts:51` annotates
*and* strips. The CMS path does neither, so those widgets also offer locked products
straight to the cart, which is the opposite of the gap described above.

Checked at type level rather than by eye:

```ts
type SL = CmsSearchlist['products'][number];
const leaks: 'whitelist' extends keyof SL ? true : never = true;      // compiles
const stripped: 'whitelist' extends keyof CmsProduct ? never : true = true; // compiles
```

Two things to fix behind it, or this comes back:

**`SearchResult['products']` keeps `'whitelist'` in its `Pick`**, so forgetting to strip
compiles fine. The guarantee currently rests on every caller remembering. Have
`searchProducts` return products already stripped and hand the evaluator what it needs
separately — then no call site can leak.

**Nothing tests the invariant.** The 25 tests cover the pure helpers well, but none asserts
the list never leaves the server. That is the test that would have caught this one.

The rest holds up. Enforcement sits in `addToCartInDb` and `checkCartItems` — the function
the cart page, checkout and order creation all call — so refusals are not UI-only. The
identity split is right and well documented: session included for refusals, identified
user only for grants, which is what stops a shared browser inheriting a subscriber's access.

Two nits: `matchesWhitelist` merges emails and npubs into one `Set`, so the two namespaces
can cross-match — harmless today, surprising later. And `saleLockedHref` is copied into
seven `ProductWidgetVariationN.svelte` files, the same drift #2771 is about.

PR #2765, opened 2026-09-07
🦋🤖 Give the struck price its own line on the product page (#2589)

### LINKED ISSUE #2589: 🦋 Improve discount display on product page
Following #2528 :

Image

Image

We should add a carriage return after striked price.

Tirodem:
On a discounted product, the crossed-out price, the discount badge and the price actually charged all sat on one line. With a long amount — four figures, or a currency written out — they ran off the side of the card and the real price was the part that got cut.

The crossed-out price now takes its own line, and the badge and the real price follow underneath. Products without a discount are unchanged, and so is the price-history button.

Fixes #2589


Tirodem:
QA OK :heavy_check_mark:

slashbinslashnoname (changes requested):
The wrapping mechanism works, but the badge ends up on the wrong line. #2589 asks for a carriage return *after* the struck price, and the body says the same — "the badge and the real price follow underneath". The `<span>` is inside the `w-full` div, so it stays up with the struck price and only the real price drops below.

While you're moving it in all three blocks: that badge is copy-pasted six times across four files — the three here, plus `cart/+page.svelte:370`, `checkout/+page.svelte:898` and `OrderSummary.svelte:151` — and the copies have already drifted. The product page uses `rounded-full whitespace-nowrap` where the other three use `rounded`, and the other three round a non-integer percentage with `toFixed(1)` while the product page prints it raw. A `<DiscountBadge percentage={…} />` here would leave one place to move instead of three, and one rendering instead of two. Aligning cart / checkout / OrderSummary on it changes other screens, so that is #2771.

Tirodem:
@slashbinslashnoname compared to the initial ticket, I made the adjustment.
If it's QA OK'd, consider the functional delta as desired evolution / improvement :ok_hand: 
(tl;dr I didn't have time to edit the original issue)

slashbinslashnoname (approved):
👍

Tirodem:
Improvement done. Please review again @slashbinslashnoname

slashbinslashnoname (approved):
The wrap works, and extracting `DiscountBadge` is the start of #2771.

It also closes a real divergence rather than just moving markup: the product page printed
`-{percentage}%` raw, while `OrderSummary`, the cart and the checkout all rounded with
`Number.isInteger ? : toFixed(1)`. A 12.345% discount read `-12.345%` here and `-12.3%`
three screens later. The component adopts the rounding, so it now reads the same everywhere.

Two nits, neither blocking.

The component keeps this page's `rounded-full whitespace-nowrap`; the other three copies
use plain `rounded`. Whoever finishes #2771 will change their pill shape — worth deciding
now rather than discovering it in the follow-up.

`percentage={data.discount?.mode === 'percentage' ? data.discount.percentage : 0}` sits
inside `{#if showStrikeThrough}`, which already implies that mode, so the `: 0` can never
render. It is there because narrowing does not survive the `@const` — hoisting the
percentage into the `@const` alongside `showStrikeThrough` would drop it.

